PR #2, opened 2019-08-25
Parser Bug: Creates headlines instead of parsing **bold** correctly

munen:
Steps to reproduce:

    - In Emacs, create an org-file with bold text at the beginning of a line
    - Open it in organ
    - Make any kind of change which triggers a sync

Here's a screenshot of Magit showing the diff after these steps:

image

For clarity, organ sees:

```
*Alain*: Anne-Catherine, Beat, steht bitte auf!

*Alain*: Lieber Beat, ich darf dich fragen: Bist du heute mit
 Anne-Catherine hierhergekommen, um vor euren Familien und Freunden
 eure Liebe zu bezeugen und den Bund der Ehe einzugehen?\\
```

and creates:

```
******** Anne-Catherine, Beat, steht bitte auf!
******** Lieber Beat, ich darf dich fragen: Bist du heute mit
Anne-Catherine hierhergekommen, um vor euren Familien und Freunden
eure Liebe zu bezeugen und den Bund der Ehe einzugehen?\\
```


munen:
This is actually a bigger bug. In `parse_org.js`:278, there's a single rather simple regex looking to find equally headers and list items. That's an oversimplification.

Headlines are actually quite complicated in Org - titles cannot even be read with context-free grammers, because they are only defined in the context of the things around.

Before we get the proper BNF parser in, this bug is quite hard to fix. We could make a workaround and detect list items and headers separately with different regex with some code around. But it could still easily be brittle in a different spot.

PR #62, opened 2019-10-02
fix: Remove unneeded code that was tagged with "ridiculous hack"

munen:
Tested on iOS 12 and 13 that the presumed iOS bug does not appear after removing the code.


munen:
Not only does this seem unnecessary, it sometimes created errors:

image

PR #78, opened 2019-10-26
Changed the icon color on swipe to contrast the background

ragone:
As a user, when I swipe a header, I want the icon to contrast the background, so that better see the action taken

chrome-capture

### DIFF
--- a/src/components/OrgFile/__snapshots__/OrgFile.integration.test.js.snap
+++ b/src/components/OrgFile/__snapshots__/OrgFile.integration.test.js.snap
@@ -65,7 +65,7 @@ exports[`Render all views Renders everything starting from an Org file renders a
           >
             <i
               class="fas fa-check swipe-action-container__icon swipe-action-container__icon--left"
-              style="display: none;"
+              style="display: none; color: rgb(0, 0, 0);"
             />
           </div>
           <div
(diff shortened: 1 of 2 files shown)


munen:
Hi @ragone 

Thank you very much for your great PR! I really appreciate the way you did it - with a user story and GIF. That makes it very easy to review.

I also like the feature very much - it makes a lot of sense! So I definitively want to merge it. It's just not working on my machine, though^^

I made a GIF on how I tested it. Do you see something amiss with my test or did you maybe not push everything?

In any case, let's get this to work and let's merge it in!

demo

ragone:
Thanks @munen! Happy to contribute :)

I think there might be a misunderstanding of what the change is, as it looks like it is working for you. The change is simply making the icon go from black to white at the same transition as the background-color. 

The background itself is unchanged. I hope that makes sense.

munen:
Hm.. you are absolutely right! I was misreading the GIF, but your user story is actually very clear on what's happening. And you're right, it works as expected and it is a great addition!

To elaborate on my mistake: I expected the background for the toggled headline to be grayed out as it shows on your GIF. However, you probably just clicked it before you started the toggle(;

In any case, your change totally works for me (as my GIF shows). The code also makes sense now^^ And improving the icon contrast definitively is a good change. Sorry for the initial mistake. It's a definitive merge!

Thanks, again! 🙏

munen:
@ragone FYI, I happily added your contributions with attribution to the changelog: https://github.com/200ok-ch/organice/commit/a4df7e4eff569b9adc44333df7dc464bdc2e874e

PR #79, opened 2019-10-26
style: change button cursors

ragone:
As a user, when I hover a clickable element, I want my cursor to change, so that I can see that it is clickable.

chrome-capture (1)

### DIFF
--- a/src/base.css
+++ b/src/base.css
@@ -35,6 +35,7 @@ h2 {
 }
 
 .btn {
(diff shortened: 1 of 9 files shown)


munen:
Hi @ragone 

This is a great addition, I really like it! Thank you for providing the code and documenting the PR well.

I tested it and it's a clean merge, of course^^

Thanks, again!

PR #81, opened 2019-10-26
fix: error on clearing date picker

ragone:
The datepicker was throwing an error when trying to parse the date on clear. If no date is picker, do nothing.


munen:
Thank you for the catch!

You are right that it's throwing an error. That's actually a missing feature that was forgotten about.

You're preventing the error, which is already good effort. But now the interface just won't budge on clicking the removal 'x'. To implement it, we'd need to write a new reducer which can actually remove planning items.

I'll gladly merge your PR, because not throwing an error in production is definitively better than throwing one. But I'll also open a bug issue to document the issue. Thanks for brining it up and coming up with a quick solution, too!(;

munen:
For future reference, here's the issue with enough information to get started on implementing the functionality: https://github.com/200ok-ch/organice/issues/83

PR #90, opened 2019-10-31
Keep all in-file settings and content lines from top of file

andersjohansson:
They will not be visible and editable for now, but are not lost on
import and export (better than the current situation in my opinion).

This can be viewed as a temporary (perhaps not so beautiful) fix for avoiding information loss until the proper parser is completed.

----

#

### DIFF
--- a/README.org
+++ b/README.org
@@ -36,19 +36,17 @@ great choice.
 "Current" means we're working hard on removing the following
 restrictions and expectations.
 
-- organice understands only few in-buffer settings (=#+TODO:= and
+- organice understands only a few in-buffer settings (=#+TODO:= and
   =#+TYP_TODO:=).
-  - If you have other in-buffer settings and save the file with
-    organice, they will be lost.
-- After potential in-buffer settings, your Org file _has to_ begin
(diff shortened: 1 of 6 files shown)


munen:
@andersjohansson Solid job, this is a great improvement until the new parser is done. The latter will take some time, as will the integration to organice. And not losing in-file settings and content prior to the first headline until then is really valuable. I actually forgot to open that issue in organice, I did open it in org-web before the fork.

In any case, I can close one of my personal TODOs now and include the in-file settings to the top of my files again. I had a workaround of a :noexport: headline in the meantime which is a little bit ugly(; Thank you for that!

After confirming it with various real world Org files with lots of settings, I wrote a [unit test](https://github.com/200ok-ch/organice/commit/7d6e5113ae089ecb1f7800b16f7e549f628b32bb) and updated the [changelog](https://github.com/200ok-ch/organice/commit/05400c533b1019fc5bdc0644ae852f6dbe636307) to accord for your contribution.

Thank you very much for this, you're making quite a few people happy :pray: :bowing_man:

Issue #105, opened 2019-11-15
Don't insert extra newlines in block without content


jamesnvc:
I'm not sure if this is by design, but I prefer that a header without a body wouldn't insert extra newlines.

e.g. I would like

```org
* Foo
** Bar
   :LOGBOOK:
   CLOCK: [2019-11-11 Mon 13:12]--[2019-11-12 Tue 13:12] => 24:00
   :END: 
** Baz
```

but organice exports like

```org
* Foo
** Bar
   :LOGBOOK:
   CLOCK: [2019-11-11 Mon 13:12]--[2019-11-12 Tue 13:12] => 24:00
   :END:
 
** Baz
```

Would a PR for this be acceptable, or is this intended behaviour?

munen:
> Would a PR for this be acceptable, or is this intended behaviour?

Very acceptable, indeed.

Thank you for coming up with a PR right away, too! :pray:  

Happy to say it's merged and I updated the changelog with your contribution: https://github.com/200ok-ch/organice/commit/4a757fa35d307a392419415d1820bed0486e442a

PR #106, opened 2019-11-15
Don't put newlines after headers with no content, add newline at EOF

jamesnvc:
Address #105, if that's desired.

This also addresses the issue in the comment in OrgFile.unit.test.js, by making sure the exported file ends with a newline


src/lib/export_org.js
```diff
@@ -301,16 +301,18 @@ export default (headers, todoKeywordSets, fileConfigLines, linesBeforeHeadings)
         contents += `\n${indentation}:END:\n`;
       }
 
-      if (header.description) {
+      if (header.description.length > 0) {
```
jamesnvc: Description is an array, so even when empty header.description is truthy

munen:
Looks good to me. Thank you for cleanin up! :pray:

Issue #111, opened 2019-11-21
Parsing of logged time in :LOGBOOK: drawers with comments fails


munen:
Clocking time within organice and saving it in a `:LOGBOOK:` drawer is a new feature (See https://github.com/200ok-ch/organice/pull/103).

It works fine, unless the clocked time is intervoven with comments in which case the parser crashes [here](https://github.com/200ok-ch/organice/blob/master/src/lib/parse_org.js#L532).

According to the [Emacs documentation](https://orgmode.org/manual/Drawers.html), "can contain anything but a headline and another drawer", though. 

A first test case (added to [logbook.org](https://github.com/200ok-ch/organice/blob/master/test_helpers/fixtures/logbook.org)) for this could be:

```
* Logbook with comments within the clocked times
  :LOGBOOK:
  CLOCK: [2019-11-21 Thu 21:12]--[2019-11-21 Thu 21:12] =>  0:00
  Here's another comment
  CLOCK: [2019-11-21 Thu 21:12]--[2019-11-21 Thu 21:12] =>  0:00
  CLOCK: [2019-11-12 Tue 14:15]--[2019-11-12 Tue 13:20] => -0:55
  There's a comment
  CLOCK: [2019-11-12 Tue 14:15]--[2019-11-12 Tue 14:25] =>  0:10
  :END:
```

munen:
The issue was first found by @Linuus in this ticket: https://github.com/200ok-ch/organice/issues/108

munen:
@jamesnvc Since you've build the initial version of clocking time and hence might be interested, I'm pinging you here. Of course, there's neither obligation to work on this or even answer to this message. If you're not interested, please just ignore the notification and be sure that I'm sorry about cluttering your inbox(;

jamesnvc:
Oh yes, I ran in to that on some of my own data; I'm happy to work on a fix

jamesnvc:
Would removing the "invalid" data be acceptable?

Linuus:
> Would removing the "invalid" data be acceptable?

There’s nothing invalid about the data so please do not remove anything? :)

munen:
@jamesnvc 

> Would removing the "invalid" data be acceptable?

Generally speaking, I'd say that parsing without crashing is good. However, for this specific case, I'd have the following two arguments:

- One goal of organice is not to alter existing Org files so that there's no diff when using Emacs and organice. That's a big differentiator to me compared with alternative tools.
[...]

Issue #110, opened 2019-11-21
Org file clock are added by accident.


tru2dagame:
![image](https://user-images.githubusercontent.com/1673006/69372167-24ebe800-0cdc-11ea-845b-c9f0092604f8.png)

I found couple times, that organice would mess up the state and clock by accident.

munen:
Yes, that's a good catch! From memory, I can see why the parser did it.

Out of curiosity, how did you manage to get those ` - State "DONE"       from "TODO"       [2019-11-21 Thu 21:46]` statements within the `:LOGBOOK:` drawer? I only know them from schedules/deadlines in which they would get a separate drawer like this:

```
* TODO Header with schedule and clocked time
  SCHEDULED: <2019-11-22 Fri +1d>
  :PROPERTIES:
  :LAST_REPEAT: [2019-11-21 Thu 21:46]
  :END:
  
  - State "DONE"       from "TODO"       [2019-11-21 Thu 21:46]
  :LOGBOOK:
  CLOCK: [2019-11-12 Tue 14:15]--[2019-11-12 Tue 13:20] => -0:55
  CLOCK: [2019-11-12 Tue 14:15]--[2019-11-12 Tue 14:25] =>  0:10
  :END:
```

I'm assuming there's a configuration flag that'll make org track the toggled time and add it to the same drawer as the clocked time.

In any case, the parser will have to be hardened, here.

tru2dagame:
Hi @Linuus
I use C-c C-t d to make it done.
The emacs configuration I use is https://github.com/purcell/emacs.d

tru2dagame:
Hi @munen,
I use C-c C-t d to make it done.
The emacs configuration I use is https://github.com/purcell/emacs.d

munen:
What function does `C-c C-t d` call? You can find out with `C-h c C-c C-t` (`C-h c` will always give you the description to key presses).

By default `C-c C-t` already calls `org-todo` which toggles the state, so the `d` wouldn't be part of the command.

tru2dagame:
![image](https://user-images.githubusercontent.com/1673006/69376393-2b329200-0ce5-11ea-988b-fe72eb869f34.png)
It triggers `org-todo` then mark it `done`.

tru2dagame:
@munen 
I think it's this one: https://emacs.stackexchange.com/questions/47951/marking-a-todo-item-done-does-not-create-a-closing-timestamp-in-org-mode

[...]

Issue #108, opened 2019-11-17
Some of the org files in dropbox can't be opened.


tru2dagame:
Some of the org files in my dropbox would show "File not found".
But the backup file named `xxx.org.organice-bak` would be always created.

![image](https://user-images.githubusercontent.com/1673006/69012776-7f233b00-09b4-11ea-978a-e150c812354c.png)

munen:
Can you try reloading the page? I suppose this is a Dropbox error. If the problem persists, it would be helpful to have the output of your browsers console.

Linuus:
Same here. Some files work if I reload. Some files never work. :(

munen:
@Linuus Do you have any errors in the console when it happens? If so, please relay them here.

Linuus:
I’ve only tried it on my iPhone and I don’t think I can get a web console here? I’ll check on my laptop tomorrow.

munen:
No, Apple doesn’t make that kind of info available, unfortunately. 

If you can get errors posted here from the laptop, it’ll be a great help! Otherwise I am stuck saying “it works for me” which I don’t want to, really(;

Linuus:
Here's the error I get on my laptop when I open the file and then reload:

![Screenshot 2019-11-19 at 07 40 47](https://user-images.githubusercontent.com/379823/69122908-4cef0780-0aa0-11ea-95ed-825c2ed6bb47.png)

munen:
Thank you for sending the screenshot with an error! From reading it, my best guess is that the Dropbox API sometimes doesn't successfully retrieve a file and sends this error. Maybe when they are under too much load. When they return "path not found" sometimes and sometimes they return a file, it seems like an error on their part.

I just spend 15min logging in and reloading organice in three different browsers on two different machines. It works for me every time, unfortunately.

There's multiple safe-guards in the code to check if a path is set. For example [here](https://github.com/200ok-ch/organice/blob/master/src/sync_backend_clients/dropbox_sync_backend_client.js#L117) for the dropbox sync backend or [here](https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/index.js#L78) where the loading of the file for the specific strategy is triggered.

It would be best if somebody can debug/confirm this who actually gets the error. The question is whether something changes in the request from organice between "it fails" and "it works". If so, there's a clean bug to fix. If not, we would need to write a work around for Dropbox that probably goes along with "keep retrying automatically until Dropbox finally finds the file"(; 

[...]

PR #112, opened 2019-11-22
Handle non-clock entries in logbook

jamesnvc:
Address #111 and (hopefully) #108 (edit: and #110)

Currently, this lacks the ability to edit comments/non-clock lines in the logbook, but will preserve them.

### DIFF
--- a/src/components/OrgFile/components/Header/index.js
+++ b/src/components/OrgFile/components/Header/index.js
@@ -307,7 +307,9 @@ class Header extends PureComponent {
       'header--removing': isPlayingRemoveAnimation,
     });
 
-    const logBookEntries = header.get('logBookEntries');
+    const logBookEntries = header
+      .get('logBookEntries')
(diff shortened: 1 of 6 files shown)


munen:
Big thanks for the quick PR! :pray: 

During testing, I found a regression. When there's comments in the `:LOGBOOK:` drawer, the only action is "clock out", never "clock in". That's because the `hasActiveClock` logic in Header/index.js only looks for missing `end` elements whereas now it should probably also look for `raw` elements.

image

munen:
@jamesnvc Super job - PR looking sharp as always! :pray: 

Gladly pulling it in without changes^^

munen:
I updated the changelog with attribution: https://github.com/200ok-ch/organice/commit/c9375f8ed34ef53e192ab75139cbea0d220a77d0

Great job on potentially closing three issues in one go! :rocket:

Issue #118, opened 2019-11-25
Feature request: honor "#+STARTUP: nologrepeat"


jamesnvc:
For entries with repeating deadlines, it would be nice if the startup option "nologrepeat" was honoured.

munen:
Perfectly implemented by jamesnvc in PR #119.

Thank you :pray:

PR #119, opened 2019-11-25
If startup option "nologrepeat" is set, don't log repeat information

jamesnvc:
Address #118

### DIFF
--- a/src/lib/org_utils.js
+++ b/src/lib/org_utils.js
@@ -212,13 +212,12 @@ const tablePartContainsCellId = (tablePart, cellId) =>
 const doesAttributedStringContainTableCellId = (parts, cellId) =>
   parts
     .filter(part => ['table', 'list'].includes(part.get('type')))
-    .some(
-      part =>
-        part.get('type') === 'table'
-          ? tablePartContainsCellId(part, cellId)
-          : part
-              .get('items')
-              .some(item => doesAttributedStringContainTableCellId(item.get('contents'), cellId))
+    .some(part =>
+      part.get('type') === 'table'
+        ? tablePartContainsCellId(part, cellId)
+        : part
+            .get('items')
+            .some(item => doesAttributedStringContainTableCellId(item.get('contents'), cellId))
     );
 
 export const headerThatContainsTableCellId = (headers, cellId) =>
--- a/src/lib/org_utils.js
+++ b/src/lib/org_utils.js
@@ -64,6 +64,21 @@ export const parentIdOfHeaderWithId = (headers, headerId) => {
   return parentHeader.get('id');
 };
 
+export const inheritedValueOfProperty = (headers, headerIndex, property) => {
+  const headerProp = headers
+    .getIn([headerIndex, 'propertyListItems'])
+    .find(item => item.get('property') === property);
+  if (headerProp) {
+    return headerProp.get('value');
+  }
+  const parentId = parentIdOfHeaderWithId(headers, headers.getIn([headerIndex, 'id']));
+  if (parentId) {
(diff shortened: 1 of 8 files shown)


src/lib/org_utils.js
```diff
@@ -212,13 +212,12 @@ const tablePartContainsCellId = (tablePart, cellId) =>
 const doesAttributedStringContainTableCellId = (parts, cellId) =>
   parts
     .filter(part => ['table', 'list'].includes(part.get('type')))
-    .some(
```
munen: This file contains only whitespace change from prettier and can be ignored during the review. Sorry for that.

munen:
Hi @jamesnvc 

As always, great work!

I did find a small issue: When there's another startup option set (for example `#+STARTUP:   showeverything nologrepeat`), the regexp didn't catch it. 

Here's my proposal to fix it: https://github.com/200ok-ch/organice/pull/119/commits/435591e92020c5a0b2030873c9eafb0b5647d246

However, it's your PR, so I'm not merging before you are OK with my proposal. Feel free to change anything, of course^^

In any case, thank you for proposing and implementing yet another great feature! :pray: :bowing_man:

src/reducers/org.js
```diff
@@ -1033,73 +1033,80 @@ function updatePlanningItemsWithRepeaters(
...
   }
   return state;
 }
 
+/**
+ * Is the `nologrepeat` feature enabled for this buffer?
+ * More info:
+ * https://www.gnu.org/software/emacs/manual/html_node/org/Repeated-tasks.html
+ */
+export function noLogRepeatEnabledP({ state, headerIndex }) {
```
jamesnvc: Should this be called `logRepeatEnabledP` instead, since it returns `true` if it should repeat and `false` if `nologrepeat` is set?

munen: Good question! I was actually thinking about this name for a moment when I had the same idea as you^^

I'm not super happy about the name, because it's quite long, but here's my thought process: The native Org feature is called `nologrepeat`. Hence it makes sense to  call the feature flag in organice by the same name and not invert the native feature flags name. The feature flag is enabled when "nologrepeat" is 'enabled' or 'set'. 

Atm it's kinda hard to read and if I would do it from scratch, I'd call the feature "logrepeatenabled", but Org mode has prior art here(;

On the other hand, having you trip up over the same naming issue that I was already thinking about is not a great argument for the name. So I'm open for a change if after reading this you have a strong fealing it woud be better the other way around!

jamesnvc: What if we keep the name, but invert what it returns, so when nologrepeat is set it returns true? That way we keep the name, but it reads a bit more naturally; e.g. the check for logging the repeat information would be `if (!noLogRepeatEnabledP(...)) { /* log repeat /* }`

munen: Oh, my, of course! It's a little late for me here, I apologize. 

I should take a break, maybe(;

jamesnvc: 😄 Can't hurt! I can make that change & push it up

munen: Sweet!

Also thanks for adding the test for detecting via properties. I was just starting to type that one in when I got the notification about your commit(;

I'll happily pull the changes in when you ping me.

jamesnvc:
Thanks for the feedback & tests @munen! I think things are looking good now.

munen:
@jamesnvc They are looking great, indeed!

I updated the changelog and added a little bit of documentation: https://github.com/200ok-ch/organice/commit/965e4b8f9fb5db4ccc5e7b68690ea54841276bcc

Issue #132, opened 2019-12-09
Docker Support


dotcs:
**Is your feature request related to a problem? Please describe.**
I as a user might have constraints that do not allow me to use the official website. I still want to use the service, but I want to run my own version of it. 

**Describe the solution you'd like**
Having a Docker container would be great, because it would be easy to use since it does not require to have any knowledge of nodejs, yarn, etc. to install and use the software. Also this allows people to use it in modern infrastructure, such as docker-compose, Kubernetes, etc.

**Describe alternatives you've considered**
By not having a dockerized solution people need to have a setup that allows them to build the software. This requires the correct versions of nodejs (and maybe yarn). It could clash with versions that people already have installed for other projects and thus require them to think about how to handle multiple versions of nodejs, which makes things quite complicated.

**Additional context**
I am working on a PR that adds a Dockerfile and support for docker-compose.

dotcs:
@munen I have created a PR that builds a working Docker image. Having `docker` and `docker-compose` installed, you can cd into the base directory of the project and run `docker-compose up` to build the image and start a container. The service is then listening on `http://localhost:5000`.

Since the docker image is working I'd like to discuss how we can automate the process of building images automatically for new versions and push them to DockerHub. This would simplify the process even more since users can then just download the image from DockerHub without the need to even create the image themselves. I guess it should be quite easy to do when we use TravisCI or similar CI services. Have you already thought about using a CI in this project?

And thanks a lot for implementing organice and

munen:
Hi @dotcs 

Thank you for building out Docker support! :pray: 

First, let me say that I'm happy to pull in this feature, of course! But I'm happy to hear you want to do it right, so let's discuss the next steps before I do so.

First a small hiccup: I did run `docker-compose up` on my machine, but it keeps hanging in step 2:

```
Step 2/10 : RUN apk add --no-cache yarn
 ---> Running in 09fd432a17b3
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
```

The file itself is accessible from my machine, though. I restartet three times. The last time, I left it running and about after 20min, it did continue building the container and start it up!  I usually build my Docker containers from Debian, so I'm happy to leave the Alpine debugging to you(; I'm not saying we should be using Debian as a source, here. It's just what I'm using, because I'm familiar with it. I'm aware that using Alpine will create a smaller image. 

> Have you already thought about using a CI in this project?

[...]

PR #133, opened 2019-12-09
[WIP] Docker Support

dotcs:
This PR adds support for creating a Docker image from the source code which has  been requested in  #132. 

The Docker image makes it easier to run this service without the need to install the correct versions for `nodejs`, `yarn`, etc.
It also allows users to be able to run this service in a Kubernetes cluster or a Docker Swarm cluster.

This PR is still in progress while a pipeline has to be built around creating the Docker image and pushing it to some registry (e.g. Docker Hub).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,14 +51,28 @@ jobs:
 
       - run: yarn test
 
+  push-image:
+    machine: true
+    steps:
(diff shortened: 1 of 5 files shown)


munen:
This looks great!

Let's merge and see what happens(;

Issue #100, opened 2019-11-10
Move organice action header into the sticky top


munen:
Right now, the action bar is attached right underneath the selected header. Therefore it is always in the way between header and description.

During a discussion at EmacsConf it turned out that this is perceived as bad UX. A solution is is quite easy: Move the action header to the top and make it sticky.

dotcs:
@munen Please help me to understand this issue better. 

With the action bar you refer to the action elements, correct? 

![image](https://user-images.githubusercontent.com/3976183/70850881-657bf300-1e8f-11ea-8999-b5c8eeaa4bd4.png)

I'm not sure if it is best to put this bar on top of the page - even it is sticky. Why? Because the action bar is currently exactly where my finger is when I tip on a headline. Moving it to the top of the page would have two negative effects in my opinion:

1. I have to move my fingers to the top of the screen (which can be complicated if the phone is very large)
2. Because the action bar is not visible all the time, the content has to shift downwards which is also kind of confusing.

I guess a better solution would be to place it above the headline, but not on the top of the page. It would look like this:
![image](https://user-images.githubusercontent.com/3976183/70850923-de7b4a80-1e8f-11ea-83ad-28ba55488c67.png)

This way the content would still move a little, but my finger could stay at the same position to access the action bar which would IMHO be very nice! What do you think?

munen:
Very good arguments. Let's do it like this!

aspiers:
This doesn't work very well for me, and apparently I'm not the only one, so I've submitted #188 as a follow-up.

PR #136, opened 2019-12-14
Feature: Move HeaderActionDrawer before headline

dotcs:
This PR covers my idea in #100. I think it's an improvement if we put the `HeaderActionDrawer` before the headline instead of placing it after the headline.


munen:
Great addition, yet again, @dotcs! :pray: 

Thanks also for really thinking about the issue. I agree with your assessment that this is likely the better solution!

I updated the changelog with attribution to your addition.

Have a great Sunday!

PR #139, opened 2019-12-17
Improve webdav url input

dotcs:
This PR is just a minimal change, but it provides a better UX for mobile users. It changes the type of the URL input box from `type="text"` to `type="url"`, which enables mobile OSes to use the correct keyboard layout. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url).

### DIFF
--- a/src/components/SyncServiceSignIn/index.js
+++ b/src/components/SyncServiceSignIn/index.js
@@ -44,7 +44,7 @@ function WebDAVForm() {
         <p>
           <label>Url:</label>
           <input
-            type="text"
+            type="url"
             value={url}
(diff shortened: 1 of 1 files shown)


munen:
Absolutely makes sense! Pulling it in, of course.

Thank you for the good UX addition :pray:

munen:
Thank you very much :pray: :bowing_man: 

I have updated the changelog with your contribution and attribution: https://github.com/200ok-ch/organice/commit/636d341bbdffe3b25f7d6a0551b4c92940829b05

Issue #143, opened 2019-12-22
Suggestion list for inputs in property editor


schoettl:
To minimize typing and typos, a smart completion for property names and values in the PropertyListEditorModal would be useful.

**Solution**
- [ ] the suggestion list for property names should contain all property names of the org file
- [ ] the suggestion list for a property value should contain all values of the same property in the org file

**Implementation**
Add a `list=` attribute in the `<input>`s and an associated `<datalist>`.

**Problem**
As far as I see, the index.js for PropertyListEditorModal has no knowledge about other properties in the file. So, an additional argument must be passed to the modal.

munen:
Implemented by @schoettl in https://github.com/200ok-ch/organice/pull/144

:pray: :bowing_man:

PR #144, opened 2019-12-22
Suggestion list for inputs in property editor

schoettl:
For issue #143

### DIFF
--- a/src/components/OrgFile/components/PropertyListEditorModal/index.js
+++ b/src/components/OrgFile/components/PropertyListEditorModal/index.js
@@ -93,8 +93,11 @@ export default class PropertyListEditorModal extends PureComponent {
         <h2 className="drawer-modal__title">Edit property list</h2>
 
         <datalist id="datalist-property-names">
-          {allPropertyNames.map(propertyName => (
-            <option value={propertyName}/>
+          {allPropertyNames.map((propertyName, idx) => (
+            <option
+              key={idx}
+              value={propertyName}
+            />
           ))}
         </datalist>
 
@@ -145,8 +148,11 @@ export default class PropertyListEditorModal extends PureComponent {
                               list={`datalist-property-${index}-values`}
                             />
                             <datalist id={`datalist-property-${index}-values`}>
-                              {allPropertyValues.map(propertyValue => (
-                                <option value={propertyValue}/>
+                              {allPropertyValues.map((propertyValue, idx) => (
+                                <option
+                                  key={idx}
+                                  value={propertyValue}
+                                />
                               ))}
                             </datalist>
                           </div>
--- a/src/components/OrgFile/components/PropertyListEditorModal/index.js
+++ b/src/components/OrgFile/components/PropertyListEditorModal/index.js
@@ -12,6 +12,7 @@ import { Droppable, Draggable } from 'react-beautiful-dnd';
 import classNames from 'classnames';
 import { fromJS } from 'immutable';
 import _ from 'lodash';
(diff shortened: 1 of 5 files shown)


src/components/OrgFile/components/PropertyListEditorModal/index.js
```diff
@@ -85,6 +85,12 @@ export default class PropertyListEditorModal extends PureComponent {
       <Drawer onClose={onClose}>
         <h2 className="drawer-modal__title">Edit property list</h2>
 
+        <datalist id="datalist-property-names">
+          <option value="blocks"/>
+          <option value="depends"/>
+          <option value="assignee"/>
```
schoettl: @munen could you give me a hint where to get a combined list of all properties of the file from?

In other words, where is the modal called from and where are the arguments passed? Sorry for that beginner question...

munen: @schoettl One of the benefits of the used stack (React, Redux) is that you've got great tooling to work on these questions! Both of them have developer tools that you can install to your browser. Using them, you can see that the `Header` component receives this info as props:

image

[Here's](https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/components/HeaderList/index.js#L127) where that rendering is happening and the props are passed.

If you go further down the rabbit hole, you'll see that the `HeaderList` component get's this information itself through props [here](https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/components/HeaderList/index.js#L43-L55) and [here](https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/components/HeaderList/index.js#L143). 

organice implements the redux connect pattern here with the function `mapStateToProps`. This pattern is documented here: https://react-redux.js.org/using-react-redux/connect-mapstate

schoettl:
Missing: Tests and testing

schoettl:
Added unit tests. Tested in Firefox and Chromium on Arch.

Ready to merge?

munen:
@schoettl I did a quick test. For the files I used (for example sample.org), there's property values missing:

image

munen:
Also, auto-complete will always show the one match "Maverick" for this file - even if I type in something completely different. 

Image from iOS

schoettl:
> @schoettl I did a quick test. For the files I used (for example sample.org), there's property values missing:

Thanks for the feedback although I don't quite understand the problem.

1. Only values of properties with the same name are in the suggestion list.
2. No suggestions are displayed if your input already filters them away (although this might depend on the browser)

I think 1. is important, to don't have unrelated suggestions for a property value.

For 2., you can clear the input field and then you shoud see all suggestions for that field.

munen:
> Thanks for the feedback although I don't quite understand the problem.

Very easy. Let me explain: After two rounds of "Scotland Yard" and "Scrabble" I had about 1kg of cheese fondue for dinner. After that, I thought it would be nice to quickly pull in some of your changes. However, after first producing very low blood sugar and then filling my stomach with cheese, I wasn't very smart about new requirements*lol*

I apologize for the false negative!

munen:
After doing another round of tests and a code review, let me rephrase my feedback: Great feature and even greater implementation. This is pretty much a best case PR - clear feature definition, good architecture and all relevant tests already present! I'll happily pull your next Christmas present in^^ 🙏

munen:
I updated the changelog with your new feature and attribution: https://github.com/200ok-ch/organice/commit/ddda0560617369bc6e771ac8178f1dd4d25355b4

:pray: :bowing_man:

schoettl:
Thank you a lot :)

Issue #142, opened 2019-12-22
Share button for headlines and support for OS share functionality


schoettl:
**Use case:**

As a supervisor in a company, I want to be able to **share header items (via e-mail, instant messenging, or other smartphone apps)** to request more information on that item or to send information to the assigned employee.

**Solution:**
- [x] Next to the edit, tags, +, ... buttons of a headline, there should be a share button.
- [ ] Clicking this button should start the OS share system (like in Android, where you can select Telegram, WhatsApp, Gmail, ... to share a photo)
- [ ] The headline with all its text body should be shared. Including subheadlines, which can still be deleted before sending.
- [ ] Also, filename and line number should be shared, in case the other party is supposed to update the item in the org file.

The share system depends on the operating system (Android, iOS, Desktop, other). I don't know yet how to use such share systems.

-> https://w3c.github.io/web-share/ not sure if Smartphone OS already support that.

As a fallback, there is already the `mailto:` action.

munen:
@schoettl FYI: There's prior work regarding native share functionality in this issue: https://github.com/200ok-ch/organice/issues/52

In that issue, the idea is to get things shared into organice, though. In any case, there might be overlap in the implementation.

@dotcs Pinging you here so that you know that there's another interested party in implementing native share functionality.

dotcs:
What I have implemented in the PoC in #52 is a receiver for the Android (and maybe iOS?) system intent when a text/link is shared. I'm not sure if this overlaps with triggering such an event on the mobile OS, but as soon as I find some time (which is likely next year) I'll try to understand the issue better. I like the idea of this PR. So in the future it might be possible to share from and to organice. That's a great idea.

schoettl:
I change this issue to aim for a future support of the OS share systems.

PR #145, opened 2019-12-22
"Share" button for headlines

schoettl:
#142 

Currently only simple e-mail sharing (not yet tested on smart phones).

### DIFF
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -253,6 +253,26 @@ Some description content
       });
     });
 
+    describe('Sharing', () => {
+      let windowSpy;
+      beforeEach(() => {
+        windowSpy = jest.spyOn(global, 'open');
+        windowSpy.mockImplementation(x => x);
+      });
+
+      afterEach(() => {
+        windowSpy.mockRestore();
+      });
+
+      test('sends the selected header and its body as an email', () => {
+        fireEvent.click(queryByText('Another top level header'));
+        fireEvent.click(getByTestId('share'));
+        expect(global.open).toBeCalledWith(
+          'mailto:?subject=Another%20top%20level%20header&body=%0A%0ASome%20description%20content%0A'
+        );
+      });
+    });
+
     describe('Agenda', () => {
       test('renders Agenda for an Org file', () => {
         // Agenda is not visible by default
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -268,7 +268,9 @@ Some description content
         fireEvent.click(queryByText('Another top level header'));
         fireEvent.click(getByTestId('share'));
         expect(global.open).toBeCalledWith(
-          'mailto:?subject=Another%20top%20level%20header&body=%0A%0ASome%20description%20content%0A'
+          `mailto:?subject=${encodeURIComponent(
+            'Another top level header'
+          )}&body=${encodeURIComponent('\n\nSome description content\n')}`
         );
       });
     });
--- a/src/components/OrgFile/components/Header/components/HeaderActionDrawer/index.js
+++ b/src/components/OrgFile/components/Header/components/HeaderActionDrawer/index.js
@@ -27,6 +27,7 @@ export default class HeaderActionDrawer extends PureComponent {
       onClockInOutClick,
       onScheduledClick,
       hasActiveClock,
+      onShareHeader,
     } = this.props;
 
     return (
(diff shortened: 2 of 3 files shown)


schoettl:
I think this branch is ready for testing.

It is plain sharing via e-mail.

However, for opening the mail client with a `mailto:` link, there is not one way to do it. And browser may behave different and change their behavior. I used the `window.location.href = 'mailto:...'` solution.[1] It works in current Firefox and Chromium on Arch.

How can I test in on the phone? The last time I tried it, I couldn't connect to Dropbox, I think because during auth, the phone provided no domain name, but only the IP of my computer to Dropbox OAuth...

If it's OK, I would like to merge this branch and implement more complex support for OS share systems later.

[1] https://stackoverflow.com/q/21461589/999007

munen:
@schoettl 

> How can I test in on the phone? 

The easy way (when you don't actually need synchronization): Just open the sample.org file. That works without logging in.

The harder ways: 

1. The Dropbox app for organice.200ok.ch is configured to accept connections from localhost:3000 - so you can use it easily from your Desktop. And you can use it from your phone as well, if you do a port forward to make the phones localhost:3000 point to your organice server on the Desktop.
2. Use ipv6 and set up a custom domain for your machine. Configure your own Dropbox app for this domain. Note: You'll have to use HTTPS for that.

I'm usually doing the easy version or the port forward. The last option would be nice to set up once and then keep it. The best option for that would be to add a Docker container for nginx with letsencrypt that acts as a proxy for organice. There's two containers already set up to do something like this:
- jwilder/nginx-proxy
- jrcs/letsencrypt-nginx-proxy-companion

munen:
NB: I did test on iOS. The current implementation works in mobile safari and as a PWA!

schoettl:
@munen Thank you for the explanation (we could add it to a wiki or doc).

> 1. The Dropbox app for organice.200ok.ch is configured to accept connections from localhost:3000 - so you can use it easily from your Desktop. And you can use it from your phone as well, if you do a port forward to make the phones localhost:3000 point to your organice server on the Desktop.

But I don't quite understand this part. Do you mean like adding a row to my (Android) phones' `/etc/hosts` file? I guess it's only possible on a rooted phone, right? Or are they dev options on the phone for this?

munen:
@schoettl I made minor changes, tested on iOS (browser and PWA) and added an integration test. If my small changes are OK for you, please remove the WIP from the title. I'll happily merge your PR in, then!

There's one slightly bigger thing I did: I changed the call from window.location.href to window.open. This leads to a slightly worse UX (popup), but I was able to test it. Mocking the `window` object turns out to be quite hard. What do you think about that? Having slightly worse UX in favor of having a test is not optimal, as well, of course.

In any case: Thank you for the great work. As always, it's a very nice improvement!

munen:
> Thanks for this fixes! I left the three lines of comments to make it easier when we want to share more information in future.

Sure thing. I was assuming so and that's why I left them in(;

schoettl:
I tested on Android (and as mentioned above in Firefox and Chromium).

I guess I found a bug in the Gmail app on Android. If there is a `%` (e.g. an org "cookie") in the header, it produces an undefined character in the Gmail mail subject.

But as it is URI-encoded correctly and works on Linux/Thunderbird, I guess it's not our fault.

munen:
> Thank you for the explanation (we could add it to a wiki or doc).

Sure, we could add it to the README in the dev section. Would you want to write a first draft for that?

> But I don't quite understand this part. Do you mean like adding a row to my (Android) phones' /etc/hosts file? I guess it's only possible on a rooted phone, right? Or are they dev options on the phone for this?

Changing the /etc/hosts file to point to a different machine for localhost would probably screw up some things. In any case it would require rooting your phone.

No, I meant port forwarding as in SSH port forwarding (documented in `man ssh` under the `-L` flag, or in articles like this one: https://help.ubuntu.com/community/SSH/OpenSSH/PortForwarding)

Since I don't have a rooted phone (iOS), I'm using a dedicated ssh application (https://www.termius.com/) for that. If you have a shell on your phone (which is more likely since you are an Android user^^, you could just use ssh like this: `$ ssh -L 3000:localhost:3000 you@your-dev-machine`

munen:
I'm fine with not supporting edge cases of GMail(;

munen:
My wife keeps calling. Will come back tomorrow to this thread.

Let me know if you're happy with the change regarding window.open and if I should merge.

schoettl:
> There's one slightly bigger thing I did: I changed the call from window.location.href to window.open. This leads to a slightly worse UX (popup), but I was able to test it. Mocking the `window` object turns out to be quite hard. What do you think about that? Having slightly worse UX in favor of having a test is not optimal, as well, of course.

That might actually be a good change in that it makes it working "safer". Because sometimes Chromium asked me, if I really want to "change window.location.href" because there might be changes that will be lost or so...
That should not occur with `window.open()`.

I will remove the [WIP] to mark it ready for merge. Before that, I quickly want to add a comment on this decision in the code.

schoettl:
Ah, you already did comment on this decision, great :)

munen:
I also have a hunch that calling to the object will be deprecated in the future. It’s a little weird that setting an attribute on an object has this kind of side effect. I guess that’s why `window.open` is there.

munen:
Very well, let's get this merged, then!^^

munen:
I've added it to the changelog and added your attribution: https://github.com/200ok-ch/organice/commit/b7c548d3a589fbb67e1ae1c403cd5086f3627c10

Thank you for this great feature addition, @schoettl! :pray:

Issue #23, opened 2019-08-28
As a user, when I see a web link (http[s]), I want to be able to click it


munen:
When rendering the org file, check for web links and make them anchor tags.

schoettl:
Does it make sense to wait for the new org-parser with this? Does the new parser also parse the body of headlines? Specifically, can it parse the body and URIs like `https://xxx` and also `email@address.es`?

Or is it better to do this while rendering (via regexes for mail address and URIs)? Then mail-addresses in property values and titles would also be clickable.

munen:
@schoettl

> Does it make sense to wait for the new org-parser with this? 

What makes sense and what doesn't would be up to you. If you want to work on the feature now, I propose that you should work with the current parser. The alternative is to get the new parser on feature parity with the existing one and make a PR on organice to replace the old with the new one. The latter option is going to require quite a lot of work, though.

> Or is it better to do this while rendering (via regexes for mail address and URIs)? Then mail-addresses in property values and titles would also be clickable.

That would be the first option. If you want to work on this feature before the new parser is ready, that's the best way to go forward.

[Here's](https://github.com/200ok-ch/organice/blob/master/src/lib/parse_org.js) the implementation of the parser and [here](https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/OrgFile.unit.test.js#L22-L145) are the parser tests to get you started.

> Does the new parser also parse the body of headlines?

Yes. [Here](https://github.com/200ok-ch/org-parser/blob/master/test/org_parser/parser_test.cljc) are the tests of the new parser where you can see what's being parsed alrady.

munen:
@schoettl Adding this in today is a wonderful christmas present!

Great job - I'm certain you're making quite a few people happy with that. For sure I can say that I am. This is one of the features that'll be using a lot(;

:pray: :bowing_man:

munen:
I also updated the changelog with this major contribution and added your attribution: https://github.com/200ok-ch/organice/commit/6eb97cee16c5e124cb049d5b97d5d855c8ed7ef5

PR #147, opened 2019-12-23
Parse and render URLs, mail-addresses and phone numbers

schoettl:
Issue #23


schoettl:
It does not work yet: it seems like the additional regex parts (URL, e-mail, phone number) are not parsed although the regex work at https://www.debuggex.com/

@munen and I don't know how to run the *.unit.test.js files.

munen:
@schoettl You can run the tests with `yarn test`.

schoettl:
Ah, thanks, that work.
`yarn -h | grep test` did not mention that the test command exists :)

munen:
Nice that you tried to find how to run the tests!(;

It's not a standard flag for `yarn` as `yarn` is a generic dependency management tool.

However, `yarn` uses the same `package.json` file that `npm` introduced. Within it, it is best practice to have a 'scripts' hash declaring how to `start` and `test` the application.

NB: For future contributors, I added how to run tests in the ["development" section of the README](https://github.com/200ok-ch/organice#development).

P.S.: I like you already wrote tests despite you didn't know yet how to run them 👍

munen:
Generally speaking, if you want to know what other scripts `yarn` or `npm` have to offer, you can either look it up in the `package.json` file, or you can run `yarn run` which gives you a list and potential description.

schoettl:
Ah, I see the "scripts" key in the package.json. Thank you for the explanation!

munen:
Quickly looking over your PR, I have a small request: I know that the `markupAndCookieRegex` variable is not your mess. However, before we merge this PR, I'd like to invoke the [boy scout rule](https://deviq.com/boy-scout-rule/) in that it would be nice if we wouldn't make this super long and unreadable regexp any longer, but instead implement the new logic as a composable regex. I'm not saying we have to also clean up the old expressions, but implementing the new logic in a composable manner would leave the code a little cleaner and easier to test. Would that be ok for you? Not that you have to do it straight away, of course. Making it first work and then making it better totally works, of course!

munen:
> Ah, I see the "scripts" key in the package.json. Thank you for the explanation!

You're welcome, any time 🙏 

The whole React front-end ecosystem is quite complex and you said that it's new for you. For that, you're groking the whole codebase super fast! 🚀 :godmode:

schoettl:
Yes, you're right. I will have to look into the parser details to find out, how we can refactor it.

munen:
Great, big thanks!^^

schoettl:
@munen I think this is pretty ready for merge. I opened issue #150 for refactoring but I rather would to do that in a separate PR.

Could you review, comment, test a bit?

It even works in properties, e.g.

```
* Test

:PROPERTIES:
:assignee: Jakob Schöttl <my@email.org>
:END:

+Don't+ call me on: +498025123456789 (should become a tel: link)

An URL https://foo.bar.baz/xyz?a=b&d#foo and /more text/ and
this is an e-mail foo.bar@baz.org in a line of text.
```

Of course there is still room for enhancements like
- shorten long URLs with ellipses ... (only the displayed URL)
- support www.example.com URLs without protocoll
- support ftp: or other protocols

munen:
@schoettl Great job! This is just about good to be merged.

I have added user acceptance tests to your branch in this commit: https://github.com/200ok-ch/organice/pull/147/commits/d2f369ff514a21a15361ff9702fdec2f025e8595

When you run them, make sure to run `yarn install` before.

The tests you wrote are good. They ensure that the parser and the exporter will not change the original Org file by one byte. To prove that the user also sees the behavior as described by you, I added the acceptance tests.

While doing so, I did catch one error. Interestingly, I caught it only because I used the same line for the URL recognition test that you used in the unit test. The test would run if there wouldn't be a `=URL=` in the line. Here's a screenshot showing the issue:

image

This is the header I used for manual testing:

```
* TODO Header with different kinds of links
  
This is a test https://200ok.ch

http://200ok.ch

200ok.ch

Call me on +41764050567.

Write me to alain@200ok.ch.

an URL https://foo.bar.baz/xyz?a=b&d#foo and /more text/ and so 


an =URL= https://foo.bar.baz/xyz?a=b&d#foo and /more text/ and so 

Just a regular =URL= in some text.

This is an e-mail foo.bar@baz.org in a line of text. 

+Don't+ call me on: +498025123456789 (should become a tel: link) 
```

It's not a major issue, but if you could fix it before we merge it in, that would be nice. If you can't find it, I'll have a look. If I won't find it, let's change the test to not use `=URL=` in the line and then it'll run.

In any case, you're about to have contributed yet another great feature to organice! :pray: :bowing_man:

Issue #148, opened 2019-12-23
Suggestion list for inputs in tag editor


schoettl:
**Problem**
When you insert a tag for a title, you have to type it out.

**Solution**
Like in #143 there should be a completion list with all existing tags.

schoettl:
I know now, that there are already clickable tags to quickly select tags for a headline. These where hidden behind my DevTools...

Despite that, I think this feature makes sense:
- for big files with a lot of tags
- when someone wants to type tag names by hand, this would help and lead him
- consistency to property editor

PR #149, opened 2019-12-23
Add suggestion list for tags

schoettl:
issue #148 

I think this is ready to merge.

It was only a small change since the global tag list was already there.


munen:
This is actually a great idea and implementation. You're right, the list is already available in the scope - and HTML5 has native support for this kind of auto-completion. Since it's a native feature, I only tested it manually on the desktop and mobile phone. There's no need for an automated test imo.

Great thinking and implementation :pray: :bowing_man:

munen:
@schoettl I have updated the changelog with the new feature and attribution: https://github.com/200ok-ch/organice/commit/e1b974e8fe0514a59cdadd73a2a217c152b18b5e

Happy X-Mas - what a nice present as the newest organice contributor! :gift:

Issue #141, opened 2019-12-21
Filterable TODO list view


schoettl:
The agenda view only lists items that have active timestamps. However, we need see a (flat) list of tasks that can be narrowed down by filters.

For my exact use case, see below.

**Two possible implementations:**

A)

- Next to (or integrated in) the "Show agenda" button there could be a "Show task list" button.
- A modal should show the task list (read-only, like the agenda).
- A click on an item should jump to the item (like in the agenda).
- At the top of the pop-up there should be a filter interface to filter for tags, regex, ... [1]

B)

- At the top of the app, next to undo/redo, there should be a filter/search icon that opens an inline filter text input.
- This filter should narrow down the main item view and could work similar to `org-sparse-tree` [2]

As far I understand the Sparse tree feature (I'm not an emacs user yet), it folds all headlines except the ones, that are selected by the filter. But for sparse trees, it seems that we cannot filter for tag AND property AND TODO-state, right? A combined AND-filter (like in GitHub issues) would be important for me.

Wich approach do you think is better?

**Filter logic and syntax:**
For me, the most important options for filtering are:

1. TODO state
2. case-insensitive match headline text
3. tags
4. property value search

Points 1. and 2. can be combined:
- "smart-case" search
- `TODO` is case-sensitive search in the headline text (because it contains upper-case letters)
- `abc` is case-insensitive search in the headline text (because it's all lower-case)
- for both forms, `|` means a logical OR, e.g. `TODO|DEFERRED`
- space inbetween words means logical AND

Syntax suggestion for points 3. and 4.:
- search for tag: `:tagx` (like orgmode tag syntax)
- search for property with value: `:property:val` where val matches even if it's only a part of value (like orgmode property syntax)
- the `:` triggers the completions/suggestions for tags, property names, and property values

**Background – our use-case:**
The workshop supervisor should be able to work with a tablet and

- [x] look for tasks that fit for a specific employee
- [ ] graph view for dependencies inbetween task (similar to org2dot [3])
- [x] share a headline item (via e-mail) e.g. to ask a question about it
- [x] see status of a task with subtasks (`[%] [/]`)
- [x] see the agenda (`DEADLINE:`/`SCHEDULED:`)
- [x] quickly add tasks (-> capture templates)
- [x] add complex tasks from a template (the "build plan" for an order with sub tasks)
- [x] manage assignees (-> properties and smart property value completion)

[1] https://orgmode.org/manual/Filtering_002flimiting-agenda-items.html
[...]

PR #154, opened 2019-12-25
[WIP] Filterable task list

schoettl:
Related issue: #141 

## Search filter parser

- [x] filter function `isMatch(filterExpr)(header)`
- [x] parser for filter expression
- [x] suggestions/completions for filter
- [x] UI integration of filter: todo list as modal (like the agenda)

## UI enhancements

- [x] Can only a part of a modal be scrollable so that the filter input sticks always visible on the bottom? Bottom is better because then does not jump around.
- [x] Render a red border on the filter input on incorrect syntax?
- [x] Fix the CSS causing the messed-up Agenda!
- [x] Put planning items (scheduled/deadline/date) below each headline and next to each other to maximize line length and be conform with other tasks without planning items.
- [x] Enlarge the filter input field
- [x] Ability to show all headers in the TaskListModal (not just the ones with todoKeywords)
- [x] Make filter input wider and bigger?

### Known bugs

- [x] Since moving the component state into the react-redux architecture, the filtered headlines will persist when changing files. When changing files, clear the filter.

### State management

- [x] Refactor the filtering into a Redux actions/reducers
- [x] Implement toggle for switching between 'headers' and 'headers with todoKeyword' as Redux filter
  - Action: `export const setVisibilityFilter = filter => ({ type: types.SET_VISIBILITY_FILTER, filter})`
    
### Other

- [ ] Refactor the used components (if possible into functional components)
- [ ] Remove duplication between the originally copied `AgendaModal` and `AgendaDay` and the new components.

### Tests

- [ ] Write integration tests


schoettl:
If someone wants to contribute to this branch, you are welcome!

I would probably use a parser generator library for the filter expression because it may grow in future (e.g. allowing quoted text and exclusion/negation of terms). A well written parser would be easier to maintain.

I'd be happy if I got help with the UI integration. See the issue for two alternatives of UIs.

munen:
Oh, wow. You've started building a parser generator for a new query language for search, that's literally amazing!

I looked through your code and large and very readable test cases, and I'm stoked so far!

I propose to keep this PR open until there's at least a rudimentary GUI implementation, so that all relevant code is in one diff and easy to find. Or do you want me to pull this code in, already, because the parser itself is in good shape and you'd prefer having the GUI in a separated branch?

NB: I've been thinking a little big about flat list of results vs. sparse tree. Emacs Org mode shows a sparse tree for results. You're going with the flat list. At first, I wasn't sure if I liked that his diverges on how Emacs does it. However, after some time of thinking on it, my conclusion is that you're absolutely nailing it with a flat list approach! On a smartphone, there's way less space to display results. And even when organice is used from a Desktop browser (and there's equal space), the navigation capabilities of Emacs will always be far supior to what we can offer in organice. Hence, when looking into a real life/non-trivial/big Org file, narrowing the results down to the _actual_ found results will likely be a far superior UX for organice. Good call on that, Sir!(;

schoettl:
Thank for the positive feedback! I had a great chrismas evening doing test-driven pair programming with my best friend after the official part :)

schoettl:
I'm working on the UI now.

src/components/OrgFile/components/TaskListModal/index.js
```diff
@@ -72,15 +72,19 @@ class TaskListModal extends PureComponent {
     const date = new Date();
 
     const filterSuggestions = ['test'];
-    const text = filterSuggestions[0];
```
munen: I was just reading through your code when you made the force push. Then I pulled your new commits, but the application didn't start anymore. The variables `text` and `i` couldn't be found.

You're likely seeing it yourself, so just a friendly heads up(;

schoettl: Ups, I forgot to remove that line. Fixed.

munen: Sweet^^

munen:
@schoettl Out of curiosity: Will there be an option to find headlines by the text of the headline? I mean without having a tag or a todoKeyword present in the headline.

schoettl:
> @schoettl Out of curiosity: Will there be an option to find headlines by the text of the headline? I mean without having a tag or a todoKeyword present in the headline.

sure ;) just type "hello world" to find all headlines containing this two words.

munen:
Sweet! *happy* :+1: :1st_place_medal:

munen:
You've used a NodeJS API (readFileSync). That's perfectly fine as long as you want that code to run in a NodeJS environment. However, the "real application" that you mentioned above is a front-end application. Hence it runs in a browser and not within NodeJS. This means you cannot use APIs that are only available within NodeJS. For this particular API, there's no equivalent in the Browser for security reasons. It wouldn't be safe to just be able to read files from somewhere, would it?(;

However, since organice is an SPA and runs webpack, we can bundle the parser with the application. That's what I did. Otherwise I cannot include it in the front-end part of the application. 

The alternative solution would be to expose the file in the `public` directory and do an Ajax request to fetch it.

munen:
> Another way would be to generate the parser with the command line program pegjs and something like a Makefile.

That would work, too. But that would complicate the build process. Right now, we only need NodeJS and `yarn`.

All three options are sub-optimal. 

Let's do the commandline genaration of the parser as you mentioned. Let's do it like this:

- For `yarn build`, add another step that runs `pegjs` on the commandline
- Document the requirement to also install `pegps` appropriately
- Let's not check in the generated file

What do you think?

schoettl:
Yes, I think commandline generation is the best solution so far. Ajax request does not turn me on either ^^

munen:
Ok, then.

I can do it, but I'm done for the day on the computer. If there's no progress on that front next time I'm on my dev machine, I'll continue there. Should be a reasonable quick refactor.

munen:
I found some more time today. I'll do the refactor now.

munen:
I think this is a quite acceptable solution: https://github.com/200ok-ch/organice/pull/154/commits/8a36a09f62d83df66a4dc6883e2ccbdde437d980

Let me know if you think there's something to improve.

src/lib/parse_org.js
```diff
@@ -6,7 +6,7 @@ import _ from 'lodash';
...
-const timestampRegex = /* scroll to the right ---> */                                                                                                                     /([<[])(\d{4})-(\d{2})-(\d{2})(?: ([^0-9]{1,9}))?(?: ([012]?\d:[0-5]\d))?(?:-([012]?\d:[0-5]\d))?(?: ((?:\+)|(?:\+\+)|(?:\.\+)|(?:-)|(?:--))(\d+)([hdwmy]))?(?: ((?:\+)|(?:\+\+)|(?:\.\+)|(?:-)|(?:--))(\d+)([hdwmy]))?[>\]]/;
+const timestampRegex = /* scroll to the right ---> */ /([<[])(\d{4})-(\d{2})-(\d{2})(?: ([^0-9]{1,9}))?(?: ([012]?\d:[0-5]\d))?(?:-([012]?\d:[0-5]\d))?(?: ((?:\+)|(?:\+\+)|(?:\.\+)|(?:-)|(?:--))(\d+)([hdwmy]))?(?: ((?:\+)|(?:\+\+)|(?:\.\+)|(?:-)|(?:--))(\d+)([hdwmy]))?[>\]]/;
```
schoettl: That is why I don't like automatic formatters... :/
After I read The Pragmatic Programmer, I accepted that aligning code in coloms can be very useful sometimes, e.g. here where I aligned the the identical timestamp regexes.

munen: Sometimes it gives, sometimes it takes away.

In a FOSS project like this, it's otherwise impossible to keep a sane style. For organice, we're using prettier as documented in step 6 in the contributors guideline: https://github.com/200ok-ch/organice/blob/master/CONTRIBUTING.org

I'm sorry it screwed up your alining in this particular case. I'm quite certain that it's a big improvement overall, still.

munen: It could also be argued (and we did) that the regexp itself is a very nasty code smell that should be dealt with(;

schoettl: Yeah, I know. I haven't got prettier correctly installed yet in vim...

munen:
> Yes, I like that! Thanks. Pretty seamless and we have a clean file for the grammar :)

Yes, I think so, too. Thanks to npx, there's also nothing additionall to install, so even the build didn't get complicated. We definitively took the best option available(;

src/lib/parse_org.js
munen: No worries. I'm not complaining and I'm not a super big stickler on formatting. 

In others (like LISP), formatting can actually be very helpful to convey meaning regularly. In other languages that are just very verbose and syntax-heavy (like modern JS), I have come to like auto-formatting for certain languages.

munen:
We now have a super quick'n'dirty integration of your super nice filter into the GUI.

It's nothing more than a POC, though. The proper integration should come through a new Redux action.

schoettl:
Cool – but does not work as I expected, after all the unit tests :D I will look into it :)

munen:
No worries - that's not your fault! I hard coded this filter expression: `const filterExpr = [{ type: 'ignore-case', words: [this.state.filterString] }];`

It's really a super dirty hack all around(;

Next steps:

- [ ] Integrate the parser correctly (not with the hack I did)
- [ ] Refactor the filtering into a Redux action
- [ ] Show all headers in the TaskListModal (not just the ones with planningItems)
- [ ] Refactor the used components (if possible into functional components)

schoettl:
Point 3 is already implemented by in `AgendaDay`.

munen:
Hm.. I'm only seeing headers with planningItems. This is what the sample.org file looks like for me: 
image

schoettl:
The first three do not have planning items, right? No active timestamps at least.

But only headers with a TODO keyword are searched, because I regard this as the *Task list*

munen:
In any case: We've got a prototype showing off the capability of your great work. That totally calls for a virtual high-five :hand: (;

munen:
> The first three do not have planning items, right? No active timestamps at least.

> But only headers with a TODO keyword are searched, because I regard this as the Task list

Ic. In that case it works as you intended it!

I apologize, the phrase 'planning item' wasn't correct here. I meant to say "header without todoKeyword".

Having a "task list" is definitively one important case. However, many Org files don't contain todoKeywords. I would argue that those might even be the bigger files in which it'll be harder to find stuff(;

That's why I [asked you](https://github.com/200ok-ch/organice/pull/154#issuecomment-569080797) about this:

> @schoettl Out of curiosity: Will there be an option to find headlines by the text of the headline? I mean without having a tag or a todoKeyword present in the headline.

I do see a use case of having the list filtered down to tasks only. However, I would probably do that with either an additional argument in the search statement or with a toggle switch.

What do you think? Of course we don't have to do it right away. I'm merely speaking about the vision of the feature^^

schoettl:
Ah, I misunderstood that question... good point, an additional flag or checkbox would be a good thing.

src/lib/headline_filter_parser.unit.test.js
```diff
@@ -17,9 +17,12 @@ describe('Headline filter parser', () => {
     });
   });
 
-  describe('Parsing intermediary states while typing', () => {
-    test('it does not crash when beginning to type a tag', () => {
-      expect(() => parser.parse(':')).not.toThrowError();
+  describe('Parser fails with invalid syntax but this is OK', () => {
+    test('Parser fails when beginning to type a tag', () => {
+      expect(() => parser.parse(':')).toThrowError();
```
munen: Just to confirm: It's going to be expected behavior for the parser to throw errors on intermediary states?

That's OK, imo. But then we should document the `parse` method accordingly and state that it is expected to throw exceptions which should be ignored.

src/components/OrgFile/components/TaskListModal/index.js
```diff
@@ -0,0 +1,162 @@
...
+    let filteredHeaders = headers;
+    if (!_.isEmpty(this.state.filterExpr)) {
+      filteredHeaders = this.props.headers.filter(header => {
+        return isMatch(this.state.filterExpr)(header);
+      });
+    }
+
+    const date = new Date();
+
+    // TODO: read from localStorage (the react way?)
```
munen: Should the value of `lastUsedFitlerStrings` be available over multiple usage sessions?

Because if not, after reafactoring this whole logic into Redux, you can read back the info from the store.

And if you want to do it across browser sessions, we still have to go through Redux, so let's first do the this refactor before introducing another side effect with localStorage(;

munen:
I've just read through all of the new code. There has been some _really great_ work done on this PR so far. I'm very impressed with the parser and that there's already a POC that works! Amazing job, @schoettl! :rocket: 

The code for the parser and the tests that you wrote, they are perfect the way they are. Regarding the new components, I would propose a moratorium with regards to new features for a moment. The components, especially the state management, are neither done the React nor Redux way. That's totally fine so far, of course! Firstly, you did mention that you're new to React and Redux. Secondly, it's always nice to have a quickly working POC before cleaning up. However, the further we go with new features (like introducing localStorage) in these components, the harder it will be to get proper architecture and state management into the components.

My proposal is that we first work on refactoring the components `TaskListModal` and `TaskListView` with proper React and Redux architecture before adding any new features to them. I can take care of that.

In the meantime, you can continue work on any part that's not directly code in the component (like the parser or CSS).

When done, I'll announce it, of course.

Is that OK for you?

In any case, I'll get started on the refactorings. First, I'll start by moving some of the local state into the Redux flow.

src/components/OrgFile/components/TaskListModal/index.js
```diff
@@ -0,0 +1,162 @@
...
+      // state change triggers rendering? -> infinite loop
+      this.setState({ filterExpr });
+    } catch (e) {
+      //console.log(e);
+      // TODO: highlight the input (syntax error)
+    }
+  }
+
+  render() {
+    // TODO: is this THE way to get this variable? it must be provided
```
munen: The architecturally correct way to get this variable is by getting it from the Redux store by setting it in the `mapStateToProps` function.

munen: This is how to do it: https://github.com/200ok-ch/organice/pull/154/commits/aa993c8827932544de12e580040e28e0257c0ff8

munen:
I have introduced the react-redux frp (functional reactive programming) architecture for the state `filterString` and `filterStringExpr`. I've also cleared at least one of your TODOs along the way.

Taking a little break now. I don't know if I'll have the power/time to continue today. I'm still a little sick on the flu and my wife also wants to spend some time with me^^

src/components/OrgFile/components/TaskListModal/index.js
schoettl: Right, this list should be available over multiple sessions on one device/browser. But each device should have their own list, in my opinion.

Let's wait for the refactoring.

schoettl:
> When done, I'll announce it, of course.
> 
> Is that OK for you?

That's perfect. I'm happy when you do the refactoring!

bin/compile_search_parser.sh
```diff
@@ -7,4 +7,4 @@ npx pegjs -o src/lib/headline_filter_parser.{js,grammar.pegjs}
 # we're telling eslint to ignore the generated file.
 PARSER_FILE=src/lib/headline_filter_parser.js
 # Prepend the generated parser code with a `eslint-disable` comment.
-echo "/* eslint-disable */"|cat - $PARSER_FILE > /tmp/out && mv /tmp/out $PARSER_FILE
+sed -i '1i \/* eslint-disable */' "$PARSER_FILE"
```
munen: Good refactoring. I didn't use awk or sed, because I wanted to keep the script as platform agnostic as possible. I used only bash features, because bash is already a dependency of running the script.

But yeah, with adding bash as a dependency it will only run on a Unix based system anyway and that is likely to have sed just as well.

Whenever the first Windows contributor comes along, she can refactor it, I guess(;

Having thought more on it, good call on sed. It's definitively cleaner code and it's likely running on the same systems as bash.

schoettl: I think, even if we would relay on a POSIX shell, not bash, a POSIX sed would be available.

munen: Works for me!

src/components/OrgFile/components/TaskListModal/index.js
munen: I'd say this is part of a bigger initiative, then. This is the current state of offline support of organice is documented here: https://github.com/200ok-ch/organice/#pwa--offline-support

Of course, it would be possible to just include one more piece of information into localStorage (we're already persisting 'settings' and login information [here](https://github.com/200ok-ch/organice/blob/master/src/util/settings_persister.js)). But then we're making the whole system more complicated.

A better approach would be to create a proper architecture for more offline support and then integrate this one feature into this architecture.

Hence, I'd vote to leave this feature out of the current PR (which is already quite big and has many moving parts) and to create a new issue for that with the information from above.

Would that be OK for you?

schoettl: OK, I create the issue.

munen:
Ok, I'm done for the day. I got quite a few things done. Most importantly all state is now handled through the react-redux pattern, so we have new actions, reducers and props bindings. Along the way I also did some related refactoring and got rid of some eslint warnings as well as at least one TODO of you. I also made some quick changes to the UI to improve UX. Finally I've updated the description of this PR with the current state of affairs.

I'll leave it at that for the moment as my wife is calling. There's one bug that the new state management introduced that I'm aware of: The filtered headlines persist between different files on the load of the TaskListView. Say you open file 'foo.org' and then filter some things, then you open 'bar.org' and visit the TaskListView - then you will see your filtered results from the last file until you start typing. Of course that's a bug. And I'll take care of it when I get back to this.

In the meantime, since all the state has now been encapsulated in the react-redux architecture, and I won't be able to work on this tomorrow, I'm hereby removing the moratorium for others to also work on components. Feel free to make _any_ changes anywhere in the code. When I get back and have time, I'll continue with refactoring from whatever the current state was.

Thanks for letting me have a crack at this for today in the afternoon. It's been fun and I'm stoked that this feature is already in such a good shape. Amazing work, @schoettl! :pray:

munen:
Having documented the last bug, it became obvious how to fix it. So I did(;

munen:
> BTW, I got the autoformatting on save with prettier working :)

Sweet, thank you for the extra effort!

Btw, I was super skeptical about auto formatting code like this just one year ago. But some young guns kept asking me about it regularly, so I gave it a couple of tries. By now, I really appreciate it for some code bases and languages. With a bit of luck, it might grow on you^^ And if it doesn’t, thanks doubly for the extra effort! 🙏🏻

schoettl:
> > BTW, I got the autoformatting on save with prettier working :)
>
> Sweet, thank you for the extra effort!

What I really don't like are commits only for auto-formatting where you can never be sure, if someone sneaked a bug in ^^ That was the best motivation for installing it :D

src/reducers/org.js
```diff
@@ -880,6 +886,52 @@ export const updateLogEntryTime = (state, action) => {
...
+    const searchFilterExpr = headline_filter_parser.parse(searchFilter);
+    state = state.setIn(['search', 'searchFilterExpr'], searchFilterExpr);
+
+    let filteredHeaders = headers;
+    if (!_.isEmpty(searchFilterExpr)) {
+      filteredHeaders = filteredHeaders.filter(header => {
+        return isMatch(searchFilterExpr)(header);
+      });
+
+      state = state.setIn(['search', 'filteredHeaders'], filteredHeaders);
```
schoettl: @munen Is there a specific reason that you put many lines in the try-catch? Because I'd keep the try-catch only around the one line that I expect to fail (the parser). OK, but doing it this way, we don't need `let` instead of `const`.

schoettl: And I'm tempted to simplify line 898 to `filteredHeaders.filter(isMatch(searchFilterExpr));`

munen: The main reason would be that I had to think fast yesterday^^ I wanted the refactoring to be finished and be architecturally sound, so that you’re not blocked to continue(;

There’s certainly room for further improvement in the actual logic implementation, though!

Please feel free to make any changes you want in here. Both your suggestions are absolutely valid! I have full confidence in your general programming and JS skills. It’s clear from your code that you know what you’re doing^^ I just want to lend a helping hand where I can - which is mainly React-Redux architecture and acceptance tests so far. Those are very (front end and framework) specific skills. In anything more general, you’re opinion is likely just as good as mine^^ 

But that you for asking, it’s much appreciated. It’s nice and really refreshing that we can work together collaboratively like this even though it’s the first time we do so. That’s not the standard norm and I’m very grateful and happy for that!

alphapapa:
It would be helpful if the filter syntax were similar to an existing implementation (I don't see an example of what it looks like now, sorry if this is moot).  For example, Orgzly has one (although I personally find it unintuitive and somewhat awkward).  Naturally, I prefer `org-ql`'s "non-Lisp" syntax, which might be a good choice here.  It's not difficult to parse.  The documentation is here: https://github.com/alphapapa/org-ql#non-sexp-query-syntax

schoettl:
I messed something up in the agenda:

I thought, the CSS files next to each component apply only to that component. Because I also saw an `@import` somewhere, once. But it seem that all CSS is globally used by every component. Is this true?

So, I added some CSS for the TaskListView component. But that overrides the CSS rules of the AgendaDay component. Therefore there is now a display bug in the Agenda.

How do we best solve this? Copy all rules that apply from AgendaDays to TaskListViews CSS and then rename the classes from `agenda-day__*` to `task-list__*`?

alphapapa:
> > @schoettl Out of curiosity: Will there be an option to find headlines by the text of the headline? I mean without having a tag or a todoKeyword present in the headline.
> 
> sure ;) just type "hello world" to find all headlines containing this two words.

Also, forgive me if this has already been discussed (there are 82 hidden comments, which is a lot to read through), but it's important to have a dedicated query operator for heading text as opposed to entry text.  I'd recommend that a bare string search the whole entry, and that a `heading` operator search only heading text (which should not include to-do keywords, priority, tags, etc--those should all be searched with separate operators).

schoettl:
@alphapapa Thanks for the input; well the parser is already ready, with my syntax proposal from #141. My goals with the syntax are:

- easy and quick to input on the phone
- somehow intuitive for org users (`:` is for for tags and for properties, because they start with a colon in orgmode, too)
- smart-case search which fits pretty well for TODO keywords *and* headline text.

Given, it was intended to fit my needs to filter a task list.

Here are some examples:
```
TODO organice
-organice :medium
-DONE doc|man :assignee:none
```

- `-` for excluding is also common in search syntax (e.g. google uses it)
- `|` instead of `,` seems also common (although I can't give an example)
- quoted strings to search in heading are not crucial because in my syntax "hello world" finds headings with "hello world". It also finds "world hello" but that is not a problem for quick task list filtering.

To be honest, as a user, I don't want to have to put the keywords "todo:" or "heading:" in front of every simple search string.

Your search syntax is definetly good and even more powerful than mine. If you can compile your parser to JS, we can have both syntaxes! Mine for a quick todo list filtering and yours for advanced search.

Other opinions?

schoettl:
Alternatively, we can also seek for a way to add advanced features like the regex and timestamp search to my search syntax. For example `something:xxx` can be implemented in my parser without conflicts with the existing syntax. And if you really want to search for `something:xxx` in the headline, you could write `headline:something:xxx` ^^

For example `desc:text` to search in description (this is also the filter syntax in hledger).

schoettl:
I added support for single- or double-quoted search terms (it applies for the headline text filter and for the property value filter). E.g.

```
'bug in'  :reported_by:"Firstname Lastname"   :critical
```

munen:
@schoettl I have just finished writing acceptance tests for this feature which should close the last blocking issue for this amazing new feature^^

I suggest to open the following issues separately and merge this PR:

- [ ] Debounce search input (on big files it can be quite slow otherwise)
- [ ] Fix bug: on iOS (scrolling a long doesn't TaskList does not work)
- [ ] Discuss strategy on Search and TaskLists
- [ ] Write documentation in `sample.org`

Please let me know what you think. If it's a "go" for you, I'd read through the whole code one more time which I haven't done since a couple days - just to make sure - and then merge to `master`.

munen:
NB: Tomorrow, well will be having Templestay guests and meditate through the night. Hence, my time on the computer will be severely limited. 

Just an advance notice^^

schoettl:
> Please let me know what you think. If it's a "go" for you, I'd read through the whole code one more time which I haven't done since a couple days - just to make sure - and then merge to `master`.

Cool, I'm looking forward for the merge! It's good to open a follow-up PR with enhancements. I'd do that (after the merge) because then I will be able to also edit the checklist of the new PR.

munen:
I just realized: The PR is 108 commits and 35 files changed. Wow! That's totally worth a high five ✋

src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
```diff
@@ -0,0 +1,126 @@
...
+                </div>
+              </div>
+            );
+          })}
+        </div>
+      </div>
+    );
+  }
+
+  getPlanningItemsAndHeaders({ headers, searchAllHeaders, todoKeyword }) {
```
schoettl: Werde ich noch umbenennen.

src/reducers/org.js
```diff
@@ -994,11 +1057,26 @@ export default (state = new Map(), action) => {
...
+      return setSearchFilterInformation(state, action);
+    case 'SET_SEARCH_ALL_HEADERS_FLAG':
+      return setSearchAllHeadersFlag(state, action);
+
     default:
       return state;
   }
 };
 
+function getTodoKeywordSetsAsFlattenedArray(state) {
```
schoettl: This function sounds more like a helper function for org_utils.js...

munen:
@schoettl I have updated the changelog with the changes and attribution from this PR: https://github.com/200ok-ch/organice/commit/5ffa30c8393291a157a1a33030336a2532c3ed06

src/base.css
```diff
@@ -82,6 +82,10 @@ h2 {
   font-size: 16px;
 }
 
+.label-for-checkbox {
```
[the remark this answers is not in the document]
munen: Totally alright! :+1:

src/components/HeaderBar/index.js
```diff
@@ -279,11 +274,9 @@ class HeaderBar extends PureComponent {
...
 
           {isAuthenticated && (
-            <i
-              className={settingsIconClassName}
-              onClick={this.handleSettingsClick}
-              title="Settings"
-            />
+            <Link to="/settings">
+              <i className={settingsIconClassName} title="Settings" />
+            </Link>
```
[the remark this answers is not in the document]
munen: Yes, react-router-dom is a pretty good router for SPAs^^

src/components/OrgFile/__snapshots__/OrgFile.integration.test.js.snap
```diff
@@ -39,10 +39,14 @@ exports[`Render all views Org Functionality Renders everything starting from an
         class="fas fa-question-circle header-bar__actions__item"
         title="Help"
       />
-      <i
-        class="fas fa-cogs header-bar__actions__item"
-        title="Settings"
-      />
+      <a
```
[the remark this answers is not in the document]
munen: > Is there a reason why we use <a> here and <Link> above (somewhere)?

Yes, there is. The file you're commenting on is rendered markup and not JSX.

> What are these snapshot integration tests anyway?

For important regression tests like this one: https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/OrgFile.integration.test.js#L195

Documentation for jest snapshots: https://jestjs.io/docs/en/snapshot-testing

src/components/OrgFile/__snapshots__/OrgFile.integration.test.js.snap
```diff
@@ -412,11 +416,16 @@ exports[`Render all views Org Functionality Renders everything starting from an
...
+          class="btn btn--circle action-drawer__btn fas fa-lg fa-tasks"
+          style="opacity: 1;"
+          title="Show task list"
+        />
         <div
           class="action-drawer__capture-buttons-container"
         >
           <button
-            class="btn btn--circle action-drawer__btn fas fa-lg fa-list-ul"
+            class="btn btn--circle action-drawer__btn fas fa-lg fa-plus"
```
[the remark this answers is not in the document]
munen: You're welcome^^

src/components/OrgFile/components/TaskListModal/index.js
```diff
@@ -0,0 +1,124 @@
...
+class TaskListModal extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    _.bindAll(this, [
+      'handleHeaderClick',
+      'handleToggleDateDisplayType',
+      'handleFilterChange',
+      'handleSearchAllCheckboxChange',
+    ]);
```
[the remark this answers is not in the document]
munen: > I don't understand this. I thought, in ES5/6 classes, methods are methods of the instantiated object automatically.

You're right about this. However, we're not dealing with a standard class, we're dealing with a React component here. And with legacy syntax. As mentioned before, we should upgrade to the new Functional Component syntax. That will also get rid of this manual call to `bindAll`.

src/components/OrgFile/components/TaskListModal/index.js
```diff
@@ -0,0 +1,124 @@
...
+            })}
+            placeholder="e.g. -DONE doc|man :simple|easy :assignee:nobody|none"
+            list="task-list__datalist-filter"
+            onChange={this.handleFilterChange}
+          />
+          <div className="agenda__tab-container">
+            <input
+              type="checkbox"
+              className="checkbox"
+              // TODO: Why does the .checkbox css rule from the Checkbox component apply for this input? If it is by accident, can we duplicate/move the css class rule to base.css?
```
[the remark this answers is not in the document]
munen: Yes, easy. All CSS files are applied globally.

organice uses the BEM (Block Element Modifier) methodology for naming CSS. Generic names like `checkbox` should only be used for generic things that should be applied everywhere.

Documentation on BEM: Block Element Modifier

src/reducers/org.js
munen: True. I'll move it there.

src/components/OrgFile/components/TaskListModal/index.js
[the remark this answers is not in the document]
schoettl: Thanks! So I think it would a good thing to move the generic checkbox class to base.css where also the new label-for-checkbox class resides. I can do that and remove the comment.

src/reducers/org.js
schoettl: Thanks!

munen:
Thanks, but I already moved it as part of https://github.com/200ok-ch/organice/pull/203 (;

PR #155, opened 2019-12-25
[WIP] Failing unit tests for inline-markup

schoettl:
Bug was discussed [here](https://github.com/200ok-ch/organice/pull/147#issuecomment-568739231)
and via Mail with @munen


munen:
Great, thanks for adding those, @schoettl!

PR #159, opened 2019-12-26
Add parsing of www.example.com URLs without https?://

schoettl:
I added one more URL type. It only appends 47 letters to the end of the monster regex. So it does not change any capture group numbers. I hope this is OK for you @munen


munen:
Looking great. Perfect PR! :pray:

munen:
I've updated the changelog and added the attribution: https://github.com/200ok-ch/organice/commit/93afacc2184589362ad29579f55433090432c154

Great job, once again^^

PR #161, opened 2019-12-27
Refactor org parser

### LINKED ISSUE #150: Refactor current parser and regex
After I added some regexes with `|` to the end of the big regex `markupAndCookieRegex` in #147, we would like to refactor some things in `parse_org.js`:

- [ ] ~split up the monster regex in smaller pieces~ not simply possible as this would create a mess of state-ful regexes and parallel calls to `exec()` on each...
- [ ] ~refactor `parseMarkupAndCookies`: maybe use [`matchAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) instead of `exec()` instead of the while loop~ no, `matchAll()` does not provide the non-matches (normal text inbetween inline-markup)
- [x] refactor `parseMarkupAndCookies`: extract function that creates `lineParts`
- [x] use switch-case instead of repeating else-if constructs

schoettl:
should close #150

### DIFF
--- a/src/components/OrgFile/OrgFile.unit.test.js
+++ b/src/components/OrgFile/OrgFile.unit.test.js
@@ -19,7 +19,21 @@ function parseAndExportOrgFile(testOrgFile) {
 }
 
 describe('Unit Tests for Org file', () => {
-  describe('Parsing', () => {
+  describe('Test the parser', () => {
+    const expectType = result => expect(result.map(x => x.type));
+    describe('Parsing inline-markup', () => {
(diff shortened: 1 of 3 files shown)


schoettl:
Ready for merge, I think.

munen:
Great job, @schoettl! This is perfect to merge as is :pray: :bowing_man:

munen:
I updated the changelog and added attribution: https://github.com/200ok-ch/organice/commit/15b1382df3f168e01f6fd959f2ed6e8e1f0ee08f

Issue #83, opened 2019-10-26
As a user, when I'm on a headline with a planning item (schedule or deadline), I want to be able to remove it


munen:
Right now, there's a safe-guard when clicking the 'x' when trying to remove a planning item: https://github.com/200ok-ch/organice/pull/81/files

For the implementation, we'd need to need to write a new reducer which can actually remove planning items.

I'll tag it as a `bug`, because the interface shows a possibility for removal, but I'll also tag it as `enhancement`, because it's a proper new feature.

Issue #36, opened 2019-09-01
Embed a proper date picker when editing deadlines and schedules


munen:
The iOS HtML5 Date picker is lacking a lot in functionality. It only shows a date, no ideas on what day it is. There’s great components that look similar to the Emacs Org calendar. Pick one if those and implement it in the modal. 

Candidates:

- http://airbnb.io/react-dates/?path=/story/singledatepicker-sdp--default

![27F9F8BE-0F46-4E85-9C3E-34ED6BA8DE9D](https://user-images.githubusercontent.com/505721/64076791-872ee100-ccc9-11e9-8946-3e61b6d7a6da.png)

munen:
The new iOS date picker is great and makes this issue obsolete:

![25333717-5902-4096-9E04-458751086186](https://user-images.githubusercontent.com/505721/140048236-75f91fe5-c541-48a4-8730-f0c82da7b7d5.png)

PR #163, opened 2019-12-27
[WIP] Feature: delete timestamps in the timestamp edit modal

### LINKED ISSUE #30: Ability to delete schedules and deadlines
- `organice` features SCHEDULED and DEADLINE timestamps and can manage them quite well.
- However, deletion of such a timestamp would have to be done by editing the description of the headline.
- Add a simple 'delete' button in the timestamp edit modal.

schoettl:
This should close #30 

@munen can you have a look at the TODO comments in my commit? I'm curious how to get this working :)

### DIFF
--- a/src/components/OrgFile/components/TimestampEditorModal/components/TimestampEditor/index.js
+++ b/src/components/OrgFile/components/TimestampEditorModal/components/TimestampEditor/index.js
@@ -9,6 +9,7 @@ import TabButtons from '../../../../../UI/TabButtons/';
 
 import * as orgActions from '../../../../../../actions/org';
 
+import { headerWithId } from '../../../../../../lib/org_utils';
 import { renderAsText } from '../../../../../../lib/timestamps';
 
 import _ from 'lodash';
@@ -73,6 +74,18 @@ class TimestampEditor extends PureComponent {
         startOrEnd === 'start' ? ['startHour', 'startMinute'] : ['endHour', 'endMinute'];
       const [hour, minute] = format(new Date(), 'HH:mm').split(':');
       onChange(timestamp.set(hourKey, hour).set(minuteKey, minute));
+    };
+  }
+
+  handleRemoveCompleteTimestamp(header, planningItemIndex) {
+    return () => {
+      const planningItems = header.get('planningItems');
+      const changedPlanningItems = planningItems.remove(planningItemIndex);
+      header.set('planningItems', changedPlanningItems);
+      // TODO: the previous line seem to have no effect.
+      console.log(planningItems.toJS());
+      console.log(changedPlanningItems.toJS());
+      // TODO: hide/close modal
     };
   }
 
@@ -292,7 +305,8 @@ class TimestampEditor extends PureComponent {
   }
 
   render() {
-    const { timestamp, planningItemIndex } = this.props;
+    const { timestamp, planningItemIndex, header } = this.props;
+    console.log(this.props);
     const {
       isActive,
       year,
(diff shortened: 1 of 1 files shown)


munen:
@schoettl Thanks for taking a crack at this!

I'm sorry to say, but I did actually implement this during a [Hackathon](https://200ok.ch/posts/hack4glarus_hackathon_2019_winter_edition.html) we organized three weks ago. The PR is here: https://github.com/200ok-ch/organice/pull/125

It's my fault - there were duplicate issues and I closed only the more recent one (#83), but forget about about #30.

If it helps at all, it was quite hard to do - mostly the testing. Native timestamps aren't amazing - also in features. That's why in the long run, I'd like to replace them with a more feature rich date picker (for example one where you can see which day of the week you're selecting). There's a feature ticket for it here https://github.com/200ok-ch/organice/issues/36 

Voting to close this PR and issue #30

munen:
Having said this, it's obviously not good UX wise that you didn't see that it's possible to delete timestamps(;

That's another negative of the native date picker. Deleting a timestamp is possible by clicking the 'x' inside the date picker: 
image

Interestingly pretty much the same as your initial UI(;

schoettl:
Ah ok, I overlooked this tiny button :)

But in fact it is not displayed on my android phone... ah, it's not a button as in your screenshot but when I click on the picker, then there is a delete option in the dialog.

Can you merge only the first commit of this branch? It's refactoring, I don't know why there is the `<i>` tag in use, but it's deprecated in html5 in favor of `<em>` which does not make sense here either ^^

munen:
Yeah, it’s tricky. On iOS you have to click the timestamp, then a menu comes up and there it has a “clear” action. 

06EF5955-4818-4FF1-8A06-C51B8B8E71FA

It’s pretty complicated, but the benefit of it being native is that it’s at least the same for every native app^^

munen:
> Can you merge only the first commit of this branch? It's refactoring, I don't know why there is the <i> tag in use, but it's deprecated in html5 in favor of <em> which does not make sense here either ^^

I'll happily merge the first commit if an `<i>` tag is deprecated. I didn't know about that know. It doesn't mention it on [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i) and out of my head, I thought that other popular projects fontawesome use them and checked that. They also still do:

image

I assume you're referring [this the w3org article](https://www.w3.org/International/questions/qa-b-and-i-tags). The issue with w3org sometimes is that browser vendors and w3org don't always agree.

So, I'm not saying we can't use `span` in favor of `i` tags, but I want to understand the issue first(;

schoettl:
Well, right, I concluded too quickly. I always read, that I'm supposed to use semantic tags (like `<em>`) and because `<i>` was traditionally italic font, it should not be used anymore to to make text italic.

> Use the `<i>` element only when there is not a more appropriate semantic element, such as: [1]

And `<span>` is a generic inline containers without the smell of italic. And I often saw icons to be included with spans. But I guess some big "icon providers" gave the `<i>` a new meaning (-> icon) :) I did not notice that. No need to merge the commit! We better stick to the icon provider standard.

[1] https://www.w3schools.com/tags/tag_i.asp

munen:
I agree. It’s confusing stuff. The whole semantic web and federation battle seems to be lost in the big, but in the small I’m of course in favor we keep fighting for them^^

For an SPA like organice, the whole topic is tricky anyway. The semantics of the used tags in organice are not relevant for SEO, but they might at some point be for blind users. However, I doubt that organice could be used from a semantic browser for blind people atm. Emacs on the other hand would be a great option for that. Using organice with the build in accessibility features from iOS works quite ok, though. But that ignores all the markup and just speaks what it can see^^

PR #164, opened 2019-12-27
Place TODO keyword of previous header in the new header

### LINKED ISSUE #20: When creating a new header with the `+` action, use the same Keyword as prefix (like TODO) as the one from where the + was clicked
This would be the same behaviour has `S-C-RET` in Emacs.

schoettl:
This should close #20 

@munen When editing the title of a header (pressing the title edit icon), shouldn't we place the curser at the end of the line by default? At least for new headers (with a TODO keyword) this would make very much sense. And for existing headers with much text it wouldn't hurt, what do you think?

### DIFF
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -76,96 +76,7 @@ Some description content
     store.dispatch(displayFile('fixtureTestFile.org', testOrgFile));
   });
 
-  describe('Actions within an Org file', () => {
-    test('Can select a header in an org file', () => {
-      const { getByText, queryByText } = render(
-        <MemoryRouter keyLength={0}>
-          <Provider store={store}>
-            <OrgFile path="fixtureTestFile.org" />
-          </Provider>
-        </MemoryRouter>
-      );
-
-      expect(queryByText('Scheduled')).toBeFalsy();
-      expect(queryByText('Deadline')).toBeFalsy();
-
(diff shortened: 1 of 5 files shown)


schoettl:
Setting the curser position in the text input works on Arch Chromium and Firefox and Android Chrome.

Can you test on iPhone before merge?

munen:
I really like the implementation of "Place TODO keyword of previous header in the new header". Nice that you put your first code into a reducer(;

I'll write an acceptance test for it and merge.

As for:

> When editing the title of a header (pressing the title edit icon), shouldn't we place the curser at the end of the line by default?

I understand that you like this. However, it would be breaking UX change. Plus, for me personally, I don't think it would be a good default. My headlines are not all short, so when having the cursor at the end, I might not be able to read the beginning of the headline.

Having said this, I really do understand that you like this. I removed your commit from this PR (as it is a separate feature) and opened a new [PR](https://github.com/200ok-ch/organice/pull/168) for it. We can totally add the feature, but my proposal is that we introduce a [setting](https://github.com/200ok-ch/organice#other-customizations) for the behavior with the default set to `false`. Would that work for you?

munen:
Great new feature, thanks for implementing it! :pray: 

I've written an [acceptance test](https://github.com/200ok-ch/organice/pull/164/commits/4f20a3231626a2550baafd37aa461ad3ce909115) and updated the [changelog with attribution](https://github.com/200ok-ch/organice/commit/2c7852c59ef2d158ea76b5218df4a0b1d2e9e479).

PR #165, opened 2019-12-27
Proper coloring of items with DONE state

### LINKED ISSUE #16: When a todoKeyword is the last in its set, render the headline with the "DONE" colors
- [ ] Write correct CSS class for 'done' type keywords
  - Instead of `todoKeyword.toLowerCase()` it would be best to render `todo-keyword--done` if the keyword is the last of a keywordSet. Then it would get rendered with the appropriate color, no matter what the keyword is.

The relevant code is in `TitleLine.render`.

schoettl:
This closes #16 

@munen In case you know at the first glance where to get `todoKeywordSet` from, I'd be happy about a hint. Otherwise I'll figure it out, shouldn't be too hard.

### DIFF
--- a/sample.org
+++ b/sample.org
@@ -19,6 +19,8 @@ Try advancing the todo state of this header a few times!
 organice also supports [[http://orgmode.org/manual/Workflow-states.html#Workflow-states][custom todo states]] (if declared at the top of the file). Swipe right on this header a few times.
 
 Note that when the cycle restarts, it defaults to the first set of todo states. Manually edit the header to get back to a different todo state cycle (more on editing headers below!)
+*** FINISHED See that custom TODO states are colored correctly
+Custom =org-todo-keywords= keyword sequences can have finish states. They are colored appropriately. You can see that TODO and START headlines are colored red whereas DONE and FINISHED headlines are colored green.
 ** Editing headers
 When you select a header the "header action drawer" appears. The first two buttons in this drawer edit the header and description respectively. Try editing this header.
 ** Tags
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -32,6 +32,9 @@ describe('Render all views', () => {
   });
 
   const testOrgFile = `
+#+TODO: TODO | DONE
+#+TODO: START | FINISHED
+
 * Top level header
 ** A nested header
 ** TODO A todo item with schedule and deadline
@@ -47,6 +50,7 @@ Some description content
   This is an e-mail foo.bar@baz.org in a line of text.
 
   +Don't+ call me on: +498025123456789.
+* FINISHED A header with a custom todo sequence in DONE state
 `;
 
   let store;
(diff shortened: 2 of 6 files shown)


munen:
@schoettl You can get the `todoKeywordSets` as in this discussion: https://github.com/200ok-ch/organice/pull/154#pullrequestreview-336798858

A single headline doesn't have a todoKeywordSet. Those are set on a buffer level with these in-buffer settings: https://github.com/200ok-ch/organice#in-buffer-settings

schoettl:
Thanks, I got it :)

I hope this fix (especially `list.every(... list2.includes(...))`) does not slow down the rendering too much. If JS is smart, it would compute it only once for the same input...

Or should I add a line `FIXME:` comment with a note on speed optimization?

munen:
Yes, please add the FIXME. I’ll add memoization when I get to the PR^^

munen:
From looking on the diff on my phone: This looks very nice. I like you implemented it in a functional way. This will be easy to memoize, too!

I’ll probably also add an acceptance test as safeguard against regressions before merging, because it’s a user visible feature.

schoettl:
That sounds good!

> I like you implemented it in a functional way. This will be easy to memoize, too!

... coming from Haskell in my last projects :)

I will watch your changes, I have to learn about acceptance and integration tests...

munen:
> ... coming from Haskell in my last projects :)

It shows! The best JS devs I know in person or have seen in FOSS projects are the ones with some background in proper FP languages - often from Clojure or Haskell. Happy to have you here^^

Regarding your proposed FIXME: I really like that you're thinking of performance! And in the other search feature, we likely will have to think of it some time. But back to this feature: When I was commenting before, I was only on my phone and thought too quickly. For the particular function that you wrote, memoization would be overkill. The function you wrote checks if a value is a member of a small set. `todoKeywordSets` will never be many. By default it's two (TODO, DONE) and even if people add some, it'll be a handful, 10 or let's even say 20 (which I have never seen even in other PM apps). Memoization adds some overhead as it needs not only to remember values for inputs, but it also needs to look up based on what criteria the value should be looked up. Hence, I think that firstly your function will always be proper fast whilst memoization might even introduce a small penalty. To be fair, before I started to use my brain, I actually looked into the React profiler to benchmark your code against master. And there's no difference to be seen, because your function is proper fast(;

However, I did the other thinks we talked about - I made a refactoring, documented the feature in the sample.org and and wrote an integration test.

If you were curious on how I would have done the memoization: If the component would have been a functional component, we could have used a React feature called hooks - specifically the [useMemo](https://reactjs.org/docs/hooks-reference.html#usememo) hook. Since this is not a functional component, I would have used lodashs' [memoize](https://lodash.com/docs/4.17.15#memoize) function.

Really great feature, thank you very much for building it! 🙏 🙇

munen:
I have updated the changelog with the new feature and attribution: https://github.com/200ok-ch/organice/commit/d7db53ca486bbb440e0e53d89bea6e7c1213d0a8

schoettl:
Thanks for the information. Perfect, that it doesn't compromise performance! We can remove the comment when we come to that place again. In the meantime, I'm confident that nobody else will work on this and adds unnecessary complexity :D

munen:
> Perfect, that it doesn't compromise performance! 

:rocket:

> We can remove the comment when we come to that place again.

:+1:

PR #167, opened 2019-12-28
Headlines with active timestamps should appear in Agenda

### LINKED ISSUE #42: Headlines with active timestamps should appear in Agenda
Currently, only headlines with SCHEDULED and DEADLINE attributes show in the agenda. However, any headline with an active timestamp should appear exactly on that date (https://orgmode.org/manual/Timestamps.html).

schoettl:
This should close #42


src/lib/parse_org.js
```diff
@@ -598,6 +598,13 @@ const parseLogbook = rawText => {
   };
 };
 
+// TODO: Active timestamps in the headline, and (at least) active
```
munen: In Org mode, an active timestamp anywhere in the description of the headline will add it to the agenda, for example:

```
*** TODO Foo
    :PROPERTIES:
    :foo:      bar
    :END:

<2019-12-28 Sat>
```

That includes adding it to the headline, but it's likely seldomly done.

```
*** TODO <2019-12-28 Sat> Foo
```

schoettl: Actually, my task lists for work and private look all like your second example ^^

That has to do with the inferior vim plugin for orgmode but I also prefer the slimmer syntax.

schoettl:
So I did refactoring and also included active timestamps in the headline in the planning items. They should show up in the agenda (but I guess I have to fix something here).

Active timestamp in the description are not yet included in the planning items. They must be merged into the planning items after the description is fully parsed.

schoettl:
@munen maybe you can give me a hint on this:

- I update a timestamp in the headline/title
- After closing the timestamp modal, I expect that when I open the agenda modal, the changed item appears on the new date.

In fact: The item in the agenda has the updated date but is placed on the old day.

e.g.
- I set the date to 2019-12-31
- I open the agenda:
   - the displayed date is 2019-12-31
   - the day it is displayed is 2019-12-30 (today)

However, the new date seems to apply when I sync and then reload the page.

schoettl:
I think I noticed a problem in parsing and org export:

**The org export swallows an empty line after a line with planning items.**
I'm trying to fix this, but it might entail much more fixes in org parser and export.

It boils down to on how we define what the `rawDescription` should be in edge cases:

- for headline 1, it must be `''`
- for headline 2, it must be `'\n'`

```
* headline 1
* headline 2

* headline 3
```

That is, `rawDescription` must always end with `'\n'` unless the description is completely empty, right?

BTW, that is also what GNU/Linux expects from a decent "line of code". For example, take `echo -n xxx | wc -l`, it prints 0 because there no newline. Or git, it complains about "missing newline at eof". Or vim, it always writes a newline at every end of line.

As a result, we cannot safely use `trimRight()` on `rawDescription`. And we must not print a `'\n'` after we print out `rawDescription`. I think these two rules are violated in some places.

schoettl:
I fixed the problems mentioned in the last comment and added a lot of new tests (to track down the problems).

Only one snapshot test is failing now. I don't understand these kind of tests. @munen ?

schoettl:
After a fix for [this comment](https://github.com/200ok-ch/organice/pull/167#issuecomment-569761506) and [the previous comment](https://github.com/200ok-ch/organice/pull/167#issuecomment-570079125), I'd suggest we:

- do a review
- merge master into this branch
- merge this branch back into master

(or do you prefer a rebase?)

From the initial issue #42, the implementation status is

- [x] recognize active timestamps in title of headline
- [ ] recognize active timestamps in the description text

I will do the missing part in a separate PR.

munen:
Removing WIP due to [request to do a code review](https://github.com/200ok-ch/organice/pull/167#issuecomment-570080810).

I'll get to it as soon as I can.

munen:
Initial check: I've got 16 failing tests locally - same as on CI. Are you still working on this PR?

schoettl:
> Initial check: I've got 16 failing tests locally - same as on CI. Are you still working on this PR?

Sorry, yes, I merged with master and something went wrong. I'll check tomorrow and mention you again when it's ready for review.

munen:
👍

munen:
@schoettl I finally got around to looking at your PR.

First things first: It looks great. There's a lot going on, but you covered _a lot_ with new tests. Great refactoring job and great new tests!

I wouldn't be against merging this now, because it already improves the current situation. But I'll answer your questions first and then happily defer the decision to you.

> @munen maybe you can give me a hint on this:
> 
> - I update a timestamp in the headline/title
> - After closing the timestamp modal, I expect that when I open the agenda modal, the changed item appears on the new date.
> 
> In fact: The item in the agenda has the updated date but is placed on the old day.
> 
> e.g.
> - I set the date to 2019-12-31
> - I open the agenda:
>    - the displayed date is 2019-12-31
>    - the day it is displayed is 2019-12-30 (today)
> 
> However, the new date seems to apply when I sync and then reload the page.

The culprit is this: When setting a SCHEDULED or DEADLINE timestamp, this code is run: https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/components/Header/index.js#L500-L502

You can see that it has quite some logic and explicitly triggers a new Redux Action "ADD_NEW_PLANNING_ITEM". 

We could add a similar callback to the click to "+ Insert timestamp". Of course that would only start to be helpful when we fix the regression to have this clickable.

Interestingly, adding a DEADLINE manually to the description (and not via the DEADLINE icon), also adds it immediately to the AGENDA (without reload). This is because the corresponding reducer (https://github.com/200ok-ch/organice/blob/master/src/reducers/org.js#L243) parses the description. I can see you made a change to that reducer. However, as I understand you implemented parsing from titles. Editing a title triggers a different action "UPDATE_HEADER_TITLE" with the reducer function "updateHeaderTitle". I think this would be the place where we need to add a call to parse potential new planning items.

Long story short:

- This PR already improves on the status quo
- Hence, we can merge it if you want
- If you prefer, you can iterate on this PR with the feedback from above
- If you want to merge without the next iteration, let's open an issue so that we track the current state

In any case, can you please merge `master` into this branch, again? I'd give it a try, but iirc the conflicts are with other parsing code of you - so the changes to not mess up are likely better for you(;

Last but not least: Sorry for the late reply. There's been a lot going on this week (back to work, start of the monthly 7-day retreat). I'll likely be a little faster from today on, again. 🤞

schoettl:
Thank you for the feedback and answers! I will further work on this before we merge it.

I also have to think again about the design... about passing the parsed title line to `parseDescriptionPrefixElements` or better merge the `planningItems` in the end after parsing
- headline
- "description prefix" with planning items (i.e. `DEADLINE: ...`, ...)
- rest of description

munen:
:+1: 

Good luck! Let me know if I can help.

schoettl:
Note for me:
- edit parse_org.js
- get `mergePlanningItems()` working
- fix TODOs in parse_org.js and reducer org.js

src/lib/parse_org.js
```diff
@@ -794,10 +778,21 @@ export const parseOrg = fileContents => {
       logBookEntries,
     } = parseDescriptionPrefixElements(header.get('rawDescription'));
 
+    const parsedDescription = parseRawText(strippedDescription);
+
+    const mergedPlanningItems = mergePlanningItems(
+      planningItems,
+      extractActiveTimestampsForPlanningItemsFromParse(
+        'TIMESTAMP_TITLE',
```
schoettl: Is there a good alternative to this `'STRING_CONSTANTS'`? I use them in multiple files and they are also used as values in hashs (the parse results).

I could define `export const STRING_CONSTANT = 'STRING_CONSTANT'` but that to be an overhead. I would always have to import them. And another developer might not know about the constant and just use the string literal then...

src/lib/parse_org.js
```diff
@@ -809,3 +804,38 @@ export const parseOrg = fileContents => {
...
+    .filter(x => x.get('type') === 'timestamp' && x.getIn(['firstTimestamp', 'isActive']))
+    .map(x => createTimestamp({ type: type, timestamp: x.get('firstTimestamp') }));
+};
+
+// Merge planningItems from parsed title, description, and planning keywords.
+const mergePlanningItems = (...planningItems) => {
+  return planningItems[0].concat(...planningItems.slice(1));
+};
+
+export const updatePlanningItems = (planningItems, type, parsed) =>
```
schoettl: Is there a style guide or best practice to always use hashs for more than 1 or 2 arguments?

Like `({planningItems, type, parsed}) => ...`

Or what is your opinion on this, @munen ?

schoettl:
TODO: changing timestamp in title or description does not yet change the state -> agenda is not updated.

src/lib/parse_org.js
munen: That's a really good question. Some languages do offer good support for these "strings" that aren't really supposed to be strings, but some kind of identifier. Ruby calls them [symbols](https://ruby-doc.org/core-2.2.0/Symbol.html) (example: `:timestamp_title`), Clojure calls them [keywords](https://clojuredocs.org/clojure.core/keyword) (example: `timestamp_title`) - and both actually use the same syntax(;

JavaScript also has symbols and you can generate them with `Symbol()`. However, in JS a `Symbol` is always unique. Hence:

```
> Symbol('foo') != Symbol('foo')
true
```

You can use Symbols, if you want to, but then you'd have to import them everywhere like you said.

I'd say that many JS devs don't bother and use constant strings like you do. I'd opt for that, as well, and would add good jsdoc documentation for the functions parameters.

src/lib/parse_org.js
munen: Yes, using destructuring for function parameters is very common. It has a couple benefits:

- JS doesn't have 'named parameters', but with destructuring you get the same names everywhere
- Changing position doesn't require refactoring in potentially many places (which is always hard in a dynamically typed language)
- It's easy to add new properties to data and it's explicit to remove them

It's also common in other languages like Clojure.

Personally, I prefer to use destructuring when it's more than two parameters, for these reasons. And some others do, too. For example there's an eslint rule for that: https://eslint.org/docs/rules/prefer-destructuring

Having said that, I'm not nitting on a function with 3 parameters. If that's what you prefer, it's fine with me.

munen:
@schoettl Does your last comment supercede the one before? Atm, I'm interpreting it like that and don't review/merge. Or is it 'merely' a reminder for future work? I'm fine with both(;

munen:
@schoettl Notification bump: https://github.com/200ok-ch/organice/pull/167#issuecomment-576278988

I understand you currently don't have spare time right now which is totally fine, of course! But if you can find the time to answer the question:

> Does your last comment supercede the one before? Atm, I'm interpreting it like that and don't review/merge. Or is it 'merely' a reminder for future work? I'm fine with both(;

Then I know the state of the PR and can handle it accordingly^^

Thanks and I apologize for the bother :pray:

schoettl:
Oh, I'm sorry @munen, I overlooked your question! I guess there was a bug and I wanted to fix it first before further review and merge. I'll look into it soon.

munen:
@schoettl No worries - in the age of notification overload, it's super easy to overlook any number of them^^

Thanks for picking the PR up! :pray: :bowing_man:

schoettl:
@munen No new changes except the assembly of my refactoring work in this branch and your refactoring work from end of January. Works perfect together :)

But the one bug still persists (the reason why we did not yet merge):

Changing an active timestamp (in title or description) *using the timestamp modal* does not update the data model properly. The agenda then still shows the header at the old date.

If this it's easy to fix, I' be glad, if we can do it this weekend. Otherwise, I'd merge this branch. It's a lot of work, to keep it up-to-date with master, because it already has so many changes...

schoettl:
> Changing an active timestamp (in title or description) _using the timestamp modal_ does not update the data model properly. The agenda then still shows the header at the old date.

I'm working on this... the `TimestampEditorModal` takes the property `planningItemIndex`, so it should be possible to update not only the timestamp but also the `planningItems`...

schoettl:
Turns out, that I probably made a bad design decision by adding active timestamps to the `planningItems`:

- When updating an active timestamp (in parsed header or description), the timestamp in the `planningItems` must also be updated. This is not possible, because the timestamps in `planningItems` don't know to which parsed active timestamps they belong. Same problem vice versa.
- Also, updating `planningItems` can only work, when raw title and the full, "unstripped" raw description are given. But for existing headers, the description in this form does not even exist, it is "stripped", i.e. without "prefix elements" like "SCHEDULED:".

I will refactor:
- use `planningItems` (as before) only for "SCHEDULED:" and "DEADLINE:"
- extract all active timestamps (including `planningItems`), when the agenda is rendered.

munen:
The existing `planningItems` have a weird data model. Some components (`TimeStampEditorModal`, `TimeStampEditor`) only know of a `timestamp` and it's `planningItemIndex`. Meaning there's the assumption that there's always at most two planningItems.

This is very weird. It's especially weird, because the necessary metadata is there. The problem starts in the `OrgFile` component where only one timestamp called `firstTimestamp` will be pbassed down, even though more is known:

image

I have no idea why it is built that way - especially because we have all the information in the Redux store, already.

This concept is back from org-web. I have no problem if we want to get rid of it. Adding more workarounds on top of it will not make it easier for us in the future^^

munen:
I'm sorry that adding necessary functionality turns out to be such a hassle here. Thank you very much for your continued effort! :pray:

schoettl:
I have to correct my last comment:

When updating an active timestamp (in parsed header or description), the timestamp in the `planningItems` must also be updated. This is not possible *for a single timestamp*, because the timestamps in `planningItems` don't know to which parsed active timestamps they belong. Same problem vice versa.

My mentioned design decision was not so bad: Not only the agenda but also `updatePlanningItemsWithRepeaters` work with `planningItems`. Maybe more places in code work with `planningItems`. So `planningItems` still seem to be a good place to put *all* active timestamps.

munen:
Adding active timestamps to `planningItems` doesn't seem like a bad decision for me, as well.

It's only that `planningItems` work weirdly with an index instead of a type in some places (see my comment above).

schoettl:
Thanks for your comments @munen !

> The existing `planningItems` have a weird data model. Some components (`TimeStampEditorModal`, `TimeStampEditor`) only know of a `timestamp` and it's `planningItemIndex`. Meaning there's the assumption that there's always at most two planningItems.

I don't quite understand this. I think, if `planningItemIndex` is undefined or not is handled as different cases in these `TimestampEditor*` classes.

> It's especially weird, because the necessary metadata is there. The problem starts in the `OrgFile` component where only one timestamp called `firstTimestamp` will be pbassed down, even though more is known:

For all normal active timestamps, I also added only the `firstTimestamp` part to conform to current `planningItems`. We could change this, but at the moment, the "end timestamp" is not handled in the agenda anyway. So I'd say, this is good enough for now.

> I have no idea why it is built that way - especially because we have all the information in the Redux store, already.
> 
> This concept is back from org-web. I have no problem if we want to get rid of it. Adding more workarounds on top of it will not make it easier for us in the future^^

Could you maybe refactor this at some point in time? I don't feel confident enough on this redux topic yet.

I will do two (hopefully) small changes to fix the problem of "`planningItems` not updated".

munen:
> I don't quite understand this. I think, if `planningItemIndex` is
> undefined or not is handled as different cases in these
> `TimestampEditor*` classes.

I think we're talking about different things here. I'm talking generally about the very weird data modelling. In my last comment, I shared a screenshot that shows that many components only get a slice of the information (a `planningItem` and it's index). If we'd pass a a proper planningItem with a type, there would be fewer shenaningans happening further down. I personally don't understand the decision to only pass this weird data structure.

But that's not on you, that code is inherited from org-web.

>> I have no idea why it is built that way - especially because we have
>> all the information in the Redux store, already.
>> 
>> This concept is back from org-web. I have no problem if we want to
>> get rid of it. Adding more workarounds on top of it will not make it
>> easier for us in the future^^

> Could you maybe refactor this at some point in time? I don't feel
> confident enough on this redux topic yet.

Yes, that's likely a bigger issue.

I agree it makes sense to add your PR without these refactorings, as it is already quite large.

schoettl:
OK, I think we're ready to merge after a review. Well, the last review is only three commits ago and those commits essentially merge, refactor, and fix the "agenda not updated" bug.

munen:
@schoettl Great work, this PR has been a mammoth effort!

I've added last final changes before merging:

- Fixed ability to insert active timestamps for headers and descriptions
- Updated the changelog and the sample with the new feature

Thank you very much for this huge effort :pray: :bowing_man:

PR #168, opened 2019-12-28
WIP On editing a headline, place cursor at the end of the line

munen:


schoettl:
Fine, a separate PR is good.

Regarding:
https://github.com/200ok-ch/organice/pull/164#issuecomment-569421419

Maybe we find a way, to only place the curser to the end of the input, when the line was newly created and only contains the TODO keyword of the previous headline (i.e. after pressing the + button).

munen:
> Maybe we find a way, to only place the curser to the end of the input, when the line was newly created and only contains the TODO keyword of the previous headline (i.e. after pressing the + button).

I apologize, but I don't understand. I tried re-reading a couple of times, but it didn't help-_-

I read your statement above as a description of the original [PR](https://github.com/200ok-ch/organice/pull/193) you did which is already merged.

Can you try explaining again and let me know if/how I can help? 

I apologize that I missed the notification on your last message on this PR and haven't answered in two weeks. I only saw that the PR is WIP and only saw that there's a question while grooming.

schoettl:
My last comment is already implemented :)

The original intent of this PR (see title) is not favored by all users. It may be annoying for long header titles.

- It could made be an option, but in my opinion it's not important enough to add another option for this.
- The cursor could be placed at the end of the title input only for short titles (e.g. length < 80 characters). But this does not conform with the principle of least surprise.

Hence, I close this issue.

PR #173, opened 2019-12-29
WIP POC: Search in fullscreen (not in a drawer)

munen:
After your mail the question on whether we should ditch the modal or go fullscreen, I've been creating a small POC just to see how a fullscreen search would be possible instead of the drawer based approach that we have used so far. Generally speaking, it's possible.

The approach has proper URLs (`/search` is appended to the opened resource), hence URL sharing and reloading is possible without losing focus. All of the PR is in a rough POC shape. I just wanted to make a quick spike and see where the pros and cons are.

This is what it looks like:

Actually synchronized file:

image

Sample file with a search:
image

Upsides:

- More visual space for searching
- Dedicated URL for search per file

Drawbacks:

- With the current synchronization approach, the Org file will be downloaded again when changing to the new search route
- After searching, and selecting a header, we need a new mechanism to again load the regular Org file view, but this time with a selected header which has these implications:
  1. With the current synchronization approach, we would load the file a third time (starts to become overkill - plus, in the worst case the file has changed in the meantime)
  2. We need a mechanism to address a header in a 'foreign' file. There is no such thing, yet. The headers currently do have an `id`, but it's generated on the fly.
  
Alternatives:

- Rethink the synchronization strategy to save the file locally so we can switch between search and Org file view.
- Implement the search feature in the regular Org file view which would come with it's own downsides (complicating the already complicating code, but more importantly it would be weird UX if the headers constantly switch order).

Having made this POC, my proposal is to not go further along this path at this time, but instead use the drawer approach in the original PR. We could enforce the drawer size to be always at maximum size to get most of the 'more space' and 'less distraction' benefits. We would gain speed by not loading the file multiple times and we can reuse the components for future features better (I was planning on building '[refile](https://orgmode.org/manual/Refile-and-copy.html#Refile-and-copy)' on top of the new search feature which would be yet another _major_ productivity boost for organice users.

I'm not saying it's not possible to go fullscreen, but if we wanted to do that, my proposal is to do that another time. Let's first get the search feature in and working properly, then let's think about if it's worth the time to do the necessary refactoring for a little more screen estate.

What do you think @schoettl?


schoettl:
Great, I'm now read the end of your writing and found that we agree very much on staying with the modal or – you called it – drawer. Still:

From your spike and as I saw in your CSS class and variable naming, we have a small misunderstanding, in what the new feature is. Nevermind, but we must be clear about this.

- I thought it as a *task list* as an addition to the agenda.
- You thought it as the missing *search* feature.

I had in mind to maybe integrate the search feature later in the main view as a sparse-tree-feature (re-using my new library functions).

I already hesitated a bit integrating the "search all headlines" checkbox because searching all headlines fits more in the concept of a global *search* and not in the *task list*.

Having it as a *task list* in addition to the agenda, it should conform to
the agenda and also be a modal. I had an full-screen / maximum-size **modal** in mind, not a new page with URL.

I like the idea of an shareable URL (which could even include the search string) and a dedicated search feature.

But I currently like my concept of a *task list* even more because it's simple and clear how it's supposed to be used. There is even a fitting icon for this.

Well, now that I think of an future dedicated *search*: search and a filterable task list are not at all orthogonal features. But I still like the sparse tree idea, which only exands matching headlines and makes non-matching headlines gray. But would be defenitely a new feature request.

----

To the drawbacks you mentioned:

They sound pretty drastic to me. My task list should be available very quickly! And also, going back to edit the task should go quick.

PS: I don't really understand the refile feature yet. And cannot really test it in Emacs ^^

munen:
> From your spike and as I saw in your CSS class and variable naming, we have a small misunderstanding, in what the new feature is. Nevermind, but we must be clear about this.

I agree, let's talk a bit about the feature on a higher level. Maybe there's a misunderstanding on the Org resource model on your side or there's a misunderstanding on my side what your plans are and how they conflict with search.

Let's first talk a bit about the resource model of Org mode. I'll elaborate a little, because you mentioned that you're not an Emacs user(;

"Task lists" are only a sub-feature of what is possible in Org files. It's a common misconception for people new to Org mode, though. Also it's a great use case for Org. So I understand your position absolutely! Org files are more general - they are for "organizing" things. That's a pretty broad statement^^ 

For example, I use it for these things:

- time tracking and time exports
- writing offers to clients
- project management (private and business for teams)
  - even pm is bigger than just todos, because it includes other things like estimations and controlling which are both possible and great to do in Org mode
  - also not everything that's scheduled is a todo, for example a meeting should be on the agenda, but it's not directly actionable
- presentations/slides (pdf)
- documentation (html)
- books, papers (pdf)
- meeting minutes
- spreadsheets
- configuration management
- bookkeeping

Likely I'm even forgetting things(; In any case: Todos are just a fraction of what you can get out of Org mode and for what it is designed for.

Hence adding a "task list" as a "main action" to the bottom would be weird. "Main actions" should imo relate to all the features of Org mode.

> I thought it as a task list as an addition to the agenda.

I think this is where the misunderstanding between the two of us is: The Agenda isn't constrained to Tasks at all. It's just headers with either a Schedule, Deadline or Active Timestamp (the latter missing from organice, of course). This header could be an upcoming meeting or appointment - both of which are not actionable and not a todo.

> But I currently like my concept of a task list even more because it's simple and clear how it's supposed to be used. There is even a fitting icon for this.

*hihi*, fontawesome tries hard to follow it's name - it has icons for everything - search has 42 icons, too: https://fontawesome.com/icons?d=gallery&q=search

> PS: I don't really understand the refile feature yet. And cannot really test it in Emacs ^^

TL;DR: This is one of the major productivity boosts in Org mode. You can move any sub-tree to another sub-tree. That makes organizing and cleaning up super fast. 

> Well, now that I think of an future dedicated search: search and a filterable task list are not at all orthogonal features. 

I think so, too! Composable Search (as you wrote it) is a superset of a task list where every task list is a filtered set of headers.

I have no problem, however, with having a flat search that doubles as a task list in organice. In Emacs, searching headers would be a sparse tree, so that would be different. But Emacs search different anyway - it's neither composable (your search is superior imo) and there's also loads more space in Emacs on the Desktop.

Ok, now for a different pace if you're fearing for your wish to have a feature that is "just" a task list as a primary action: It is very much in the spirit of Emacs to have a basic feature and then hook into it and configure it any way you like! If you want to use the new feature primarily as a task list, we can totally do that, no problem. I propose we add a setting "search acts as task list" and then the 'checkbox' in the feature goes away and we always filter down to only tasks. We can even make the icon be a task list icon, then. It would just be the thing that you want it to be. If people don't check this setting, it's a search feature on all headers with an optional flag to filter down to just TODOs (or alternatively this could also be syntax in the parser).

You can also roll out this (and other) setting(s) to your company easily, because settings are persisted in a JSON file, which you can give to people, so they have a pre-configuration.

What do you think of that? If we're a little off, let's continue discussing this in text. If we're way off, I propose that we have a phone call to discuss our goals. The farther off we are, the closer our communication medium should be. We could also have a video call if you like! I have time tomorrow to work on this feature and I planned to do so anyway. I'll happily take the time to talk to you instead so that you're happy with the final feature exactly the way it's merged.

P.S.: In the meantime I might have thought of a way to not re-load the Org file in fullscreen mode all the time(; I will try it out tomorrow.

P.P.S.: In any case, please don't worry. You've put a lot of work into this feature and I want you to be happy about the outcome. Let's talk about where we want to go and find a solution which makes both of us happy, ok?

munen:
Note to self: Give fullscreen another try by not triggering the 'close file action'. This action would clear the headers whereas for this feature, we're only switching between views of the same file and don't need to clear the headers at all.

munen:
Closing, because we're going with the Drawer for the time being and not with the fullscreen approach.

munen:
Let's not delete the branch, though, in case we want to look back at the POC at some point.

PR #176, opened 2019-12-30
Support single- and double-quoted strings in the search filter

schoettl:
[This commit](https://github.com/200ok-ch/organice/pull/176/commits/3fd26615cd64add6a1b985a6b662e45130f18cb2) breaks `yarn start` but not `yarn test`

Update: [Fixed](https://github.com/200ok-ch/organice/pull/176/commits/f68d469e6280d4bd06e8a851ff31a33caae96727)

```
Failed to compile.

./src/reducers/org.js
Attempted import error: '../lib/headline_filter_parser' does not contain a default export (imported as 'headline_filter_parser').
```

The generated file `headline_filter_parser.js` – with or without the last commit – exports:
```
module.exports = {
  SyntaxError: peg$SyntaxError,
  parse:       peg$parse
};
```

It is imported like this:

```
import headline_filter_parser from '../lib/headline_filter_parser';
```

What is the problem? Does it work on the CI here?

### DIFF
--- a/src/lib/headline_filter.js
+++ b/src/lib/headline_filter.js
@@ -16,17 +16,18 @@ export const isMatch = filterExpr => header => {
 
   const filterFilter = (type, exclude) => x => x.type === type && x.exclude === exclude;
   const words = x => x.words;
+  const wordsLowerCase = x => x.words.map(y => y.toLowerCase());
 
   const filterTags = filterExpr.filter(filterFilter('tag', false)).map(words);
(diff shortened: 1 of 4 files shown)


schoettl:
Ready for merge.

munen:
Lgtm, nice improvement! :pray:

munen:
I've updated the changelog with the attribution: https://github.com/200ok-ch/organice/commit/84121069a51f55a5509b7892f27326dcc4754ff7

PR #177, opened 2019-12-30
Trim headline title at both sides before saving

schoettl:
... when capturing via template and when adding/editing a headline.

Makes sense?

I think it does not make sense to have trailing whitespace in any source code. OK, markdown uses trailing whitespace to declare a `<br>` but a headline in emacs does not need trailing space, right?

### DIFF
--- a/src/lib/parse_org.js
+++ b/src/lib/parse_org.js
@@ -685,6 +685,9 @@ const concatRegexes = (...regexes) =>
   );
 
 export const newHeaderFromText = (rawText, todoKeywordSets) => {
+  // This function is currently only used for capture templates.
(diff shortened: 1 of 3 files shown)


munen:
Good thinking!

I completely agree that trimming whitespace for a newly created headline is a good idea and that we should do it. There's no semantics to trailing whitespace and having it is likely an input error.

I have added some documentation on where we do trim and where we don't trim trailing whitespace.

Nice addition, thank you @schoettl :pray:

munen:
I have updated the changelog with the attribution: https://github.com/200ok-ch/organice/commit/b7f7a0081146d63ac546d1d209396f3c7902aa08

Issue #183, opened 2019-12-30
Fix Docker build


munen:
Adding a new build step broke the Docker build. For the new search feature, we have a parser geneator which is integrated in `package.json` [here](https://github.com/200ok-ch/organice/blob/master/package.json#L33). It's a simple call to `./bin/compile_search_parser.sh`.

However, when building the Docker container said file cannot be found.

I looked into it for a moment, but didn't see why the file is not included in the image. The `bin` folder is added and includes the other file in the directory. 

Since a broken build/push to Docker breaks the build for the whole project, I disabled creating the Docker image and pushing it here for the time being: https://github.com/200ok-ch/organice/commit/012f00fa295913e17c8b7ee7e34b3e4d1752e284

munen:
@dotcs Do you want to look into why the Docker image fails to include said file and hence the build fails?

dotcs:
Hi @munen
I quickly had a look into it and found the root cause. The alpine image does not include bash, so the scripts could not be executed because they have a `#!/bin/bash` shebang line.
There are two ways to fix the issue. Either rely on `/bin/sh` or install bash. I chose the latter because you already mentioned that you're more familiar with good old Debian based images and I think at least bash should be there. ;)

munen:
Hi @dotcs 
cc @schoettl 

Thanks a bunch for jumping quickly to the task! I must have made a mistake when I build the image locally. It looked to me as if the whole shell script was missing in the `bin` folder which seemed weird.

Thank you for properly debugging and fixing the issue^^

I've merged the PR and confirm that Docker images are again being built and pushed! (https://circleci.com/gh/200ok-ch/organice/514).

Please enjoy the new Search / TaskList feature(;

🙏 🙇

dotcs:
Thank you @munen for confirming that the build works again. That makes me happy! :)

PR #187, opened 2019-12-31
Bugfix: Docker build with bash scripts

dotcs:
This PR fixes issue #183. It installs `bash` in the docker container, so that the newly introduced bash scripts work without an error. It reverts also 012f00f, but keeps the `.dockerignore` file.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,17 +51,17 @@ jobs:
 
       - run: yarn test
 
-  # push-image:
-  #   machine: true
-  #   steps:
-  #     - checkout
-  #     # Build docker image and push it to DockerHub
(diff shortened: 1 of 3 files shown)


schoettl:
@dotcs Good idea. Is it necessary to also install `sed` or is it already included?The one bash script uses `sed`. It is not a GNU core utility but it's a standard tool that might be useful in future.

dotcs:
@schoettl Thanks for the hint. It's not necessary to install `sed`, because alpine comes with a compatible version. You can find out by overriding the entrypoint and run a shell within the container as such:

```bash
$ docker run --rm -it --entrypoint /bin/sh twohundredok/organice:latest
$ sed --version
This is not GNU sed version 4.0
```

munen:
Great job, thank you very much! 🙏 🙇

PR #193, opened 2020-01-01
Place cursor after the TODO keyword when creating a new header

schoettl:
This PR supersedes #168. As discussed there and in the linked comment, placing the cursor at the end of the input is problematic. But not for newly added headers, that only contain the TODO keyword.

PR #168 can be closed without merge.


munen:
Lgtm, too^^

munen:
Updated the changelog with the attribution: https://github.com/200ok-ch/organice/commit/d1e09eeab21c72fb881c0a4289e0b4ae719da03b

Thank you for adding the feature. It's a very nice UX improvement :pray:

PR #195, opened 2020-01-01
Button in headline menu to open property editor

### LINKED ISSUE #170: As a user, when editing a header which is currently without properties, I want to be able to open the property editor
As a workshop supervisor, working with properties to assign tasks to employees or to define why a task is blocked, **I want to be able to open the property editor for a headline even if there are no properties yet**.

- When in description edit mode, there should be a further button below the text input:
- Next to "+ Insert timestamp" there could be "Edit properties" which would open the property editor.

**Alternatives:**
This option could also be in the header toolbar next to the tags button, but there are already so many buttons. So we better keep it in the description edit mode.

schoettl:
This wants to close #170 

But I don't understand the framework good enough to implement this in a proper way.

There must be a better way as duplicating the `case 'property-list-editor'` in `OrgFile/index.js`.


munen:
@schoettl How about https://github.com/200ok-ch/organice/pull/195/commits/d9e833aba8dd30e33bfcd503069c92bf269415e8 ?

As far as I can tell, there's always a `selectedHeader` when calling the `PropertyListEditorModal`. It doesn't matter if we call it from the `HeaderActionBar` (you're new implementation) or by clicking on existing properties.

Unfortunately we don't have tests for this feature, yet, but it works for my manual tests(;

In any case: _Very_ nice addition :pray: 

If you're happy with my change, let's remove the WIP, and pull the change in!

schoettl:
Thanks, that looks good. I just fixed a variable name.

munen:
Nice, let’s merge the feature in, then!

munen:
A little late, but I did update the changelog with attribution: https://github.com/200ok-ch/organice/commit/8ef4e6c049b410a5267b99743031fa8e21e8c84f

Thank you, again :pray:

schoettl:
No problem and thank you for getting it to work ;)

PR #201, opened 2020-01-02
feat: Don't update search results when using an invalid search filter

### LINKED ISSUE #200: As a user, when using the Task List view and I enter an invalid filter syntax, I don't want the current view to be changed.
On invalid filter syntax, all items are displayed. Instead, the current view should not be changed.

Example of an invalid search: `bar:`

munen:
Closes #200

### DIFF
--- a/src/reducers/org.js
+++ b/src/reducers/org.js
@@ -906,12 +906,9 @@ export const setSearchFilterInformation = (state, action) => {
   }
 
   state.setIn(['search', 'searchFilterValid'], searchFilterValid);
+  // Only run filter if a filter is given and parsing was successfull
   if (searchFilterValid) {
-    let filteredHeaders = headers;
-    if (searchFilterExpr) {
-      // Only run filter if a filter is given and parsing was successfull
-      filteredHeaders = headers.filter(isMatch(searchFilterExpr));
-    }
+    const filteredHeaders = headers.filter(isMatch(searchFilterExpr));
     state.setIn(['search', 'filteredHeaders'], filteredHeaders);
   }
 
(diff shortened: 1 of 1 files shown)


src/reducers/org.js
```diff
@@ -893,23 +893,28 @@ export const setSearchFilterInformation = (state, action) => {
...
   }
 
-  state.setIn(['search', 'searchFilterValid'], !!searchFilterExpr);
-
-  let filteredHeaders = headers;
-  if (searchFilterExpr) {
-    // Only run filter if a filter is given and parsing was successfull
-    filteredHeaders = headers.filter(isMatch(searchFilterExpr));
+  state.setIn(['search', 'searchFilterValid'], searchFilterValid);
+  if (searchFilterValid) {
```
munen: @schoettl After moving more code out of the try/catch block, we introduced issue #200. I'm adding an additional conditional to remedy. Not sure if more conditionals are better than bigger try/catch blocks, though^^

schoettl: Oh, I considered that case, but apparently had a wrong implementation! I thought `searchFilterExpr` evaluates to false if parse fails, because it should be unassigned in that case. So now there are two `if`s to check for the same thing. And I don't understand why my solution not works. See line 911/912 in this diff.

munen: You're right, there's a redundant conditonal. I've remove it.

> And I don't understand why my solution not works. See line 911/912 in this diff.

It didn't work, because in 907 `filteredHeaders` where set to `headers`. They where only conditionally filtered, but in any case they were assigned to the state as `filteredHeaders`. Hence, if the parsing failed, we'd end up with all the headers.

Thanks for checking the diff and hinting to the redundant conditional. I think this is good to merge now^^

PR #206, opened 2020-01-03
Introduce Search next to TaskList

munen:
A TaskList should just be what the name says - a task list. It totally makes sense to have task specific feature that filters down to tasks and orders them by various task related critieria. Complecting this feature with a more generic "Search" will complicate the interface, the code and the understanding of the user.

Instead of going down that route, I propose to have a TaskList which does just that. Next to it, there is a standalone Search which has fewer features and only filters down on headers.

For the GUI, I've harmonized them into one collapsible FAB, because even though they are doing very different jobs - they are pretty aligned in terms of how they operate.

Open tasks:

- [x] Introcuce SearchModal
- [x] Harmonize TaskList and Search in the GUI
- [x] Refactor Task List and remove "Search all headlines" logic
- [x] Think if it makes sense to extract common markup from Agenda/TaskList and Search
- [x] Write acceptance tests for Search

@schoettl What do you think - does this work for you?

### DIFF
--- a/CONTRIBUTING.org
+++ b/CONTRIBUTING.org
@@ -44,7 +44,7 @@ These are the steps we take to ensure a high quality in the backlog:
    behavior, we ask a question or link to the documentation. We also
    add a 'question' label to the issue. If the issue creator does not
    respond to the question within 14 days, we close the issue. If she
-   does chose to respond at some point, she can re-open the issue.
+   does choose to respond at some point, she can re-open the issue.
 
 *** Development collaboration
 
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -483,8 +483,3 @@ export const setSearchFilterInformation = (searchFilter, cursorPosition) => ({
   searchFilter,
   cursorPosition,
 });
-
-export const setSearchAllHeadersFlag = searchAllHeaders => ({
-  type: 'SET_SEARCH_ALL_HEADERS_FLAG',
-  searchAllHeaders,
-});
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -298,16 +298,45 @@ Some description content
         });
       });
 
+      describe('Search', () => {
+        test('renders Search for an Org file', () => {
+          expect(queryByText('Search')).toBeFalsy();
+          expect(queryByText('A todo item with schedule and deadline')).toBeFalsy();
+
+          fireEvent.click(getByTitle('Show search'));
+          const drawerElem = getByTestId('drawer');
+          expect(drawerElem).toHaveTextContent('Search');
+          expect(drawerElem).toHaveTextContent('A todo item with schedule and deadline');
+        });
+
+        test('searches in all headers', () => {
+          fireEvent.click(getByTitle('Show search'));
+          const drawerElem = getByTestId('drawer');
+          const input = getByPlaceholderText(
+            'e.g. -DONE doc|man :simple|easy :assignee:nobody|none'
+          );
+
+          // All kinds of headers are visible
+          expect(drawerElem).toHaveTextContent('A todo item with schedule and deadline');
+          expect(drawerElem).toHaveTextContent('A header with tags');
+          expect(drawerElem).toHaveTextContent('Another top level header');
+
+          // Filter down to headers with tag :tag1:
+          fireEvent.change(input, { target: { value: ':tag1' } });
+
+          expect(drawerElem).toHaveTextContent('A header with tags');
+          expect(drawerElem).not.toHaveTextContent('Another top level header');
+        });
+      });
+
       describe('TaskList', () => {
         test('renders TaskList for an Org file', () => {
           expect(queryByText('Task list')).toBeFalsy();
-          expect(queryByText('Search all headlines')).toBeFalsy();
           expect(queryByText('A todo item with schedule and deadline')).toBeFalsy();
 
           fireEvent.click(getByTitle('Show task list'));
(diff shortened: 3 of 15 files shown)


schoettl:
I don't understand what you mean with
> Next to it, there is a standalone Search which has fewer features and only filters down on headers.

You mean the search should only search headlines without TODO keywords?

I also think it's good to not complicate or bloat the task list with features.

But to be honest, I like the idea of a search icon in the title bar (top-most) more than the FAB. One reason is that the FAB icon is what we know from "zoom-in".

And the other reason is, that I still think, that a task list has more in common with the agenda than with the search. Agenda and tasklist are both for "day/work planning". Search is not.

I refer to Google Calendar. When you click the search icon in the title bar, a search field appears right next to the search icon. How about this?

screenshot

BTW: is there any decent free image sharing service?

If we have a modal like the task list or a new page ("full screen") – I don't mind. Anyway, a flat list is good, it is more clear than a sparse tree, I think.

munen:
> I don't understand what you mean with
> > Next to it, there is a standalone Search which has fewer features and only filters down on headers.

In the PR, they are both within a collapsible FAB.

> You mean the search should only search headlines without TODO keywords? 

No. Search should be generic and just search for headlines. It should by default behave as the Emacs Org mode search would - with the exception to work on flat lists instead of sparse trees, because we're dealing with a smaller screen.

> And why "fewer features"?

Because the Task List as you have implemented it has implicit logic like re-ordering the tasks according to various criteria. This should not be in a standard search. It's a _very specific_ filtering that makes sense for a Task List, but not for search. When I'm generically searching for headers, it would be weird for TODOs to appear ahead of other headers for example.

> I also think it's good to not complicate or bloat the task list with features.

Well, I've been quoting you here anyway^^ The whole PR is to try to harmonize your requirement of having a Task List as a primary action whilst we still need a more generic search(;

> But to be honest, I like the idea of a search icon in the title bar (top-most) more than the FAB. 

> One reason is that the FAB icon is what we know from "zoom-in".

What do you mean by "what we know from zoom" that? 

According to the [material design doc]( https://material.io/components/buttons-floating-action-button/#): "A floating action button (FAB) represents the primary action of a screen."

> And the other reason is, that I still think, that a task list has more in common with the agenda than with the search. Agenda and tasklist are both for "day/work planning". Search is not.

I do believe that this is a different thing for you. However, organice strives to be easy to use for an Emacs Org mode user. And
in Emacs filtering for TODOs and headline content is behind the same command: image

That's why I asked you on the first day of the implementation whether it will be possible to search for headers without todoKeywords, tags or the like^^

> I refer to Google Calendar. When you click the search icon in the title bar, a search field appears right next to the search icon. How about this?

Great example! Let's do it like Google.

Google Maps starts with a very generic search (that would be headers for us):

7E977840-7325-42FC-845F-D3BF60A5D039

Then you can drill down to a more specific search like directions (that would be Task List for us).



Then you can set even more information like the type of transportation used (that could be ordering and the like for us).


I'm in favor of taking your proposal to do it like Google Search(;

The implication would be to start with a generic search which can be turned into a Task List on click. This would be akin to how Emacs filtering behaves and also how Google turns a generic search into a more specific search.

> BTW: is there any decent free image sharing service?

For this particular medium, I'd go with the built-in image sharing from Github? Just drag'n'drop an image to a comment or click "Attach files" below a comment.

munen:
01E335F2-6FCA-42C7-835E-1F9394F52DC4
9A39D3FE-95DB-438B-97C0-6CD2A1962C38
8B7CF9B6-4DBB-49CF-B68C-5C38376E7FAB

schoettl:
> > One reason is that the FAB icon is what we know from "zoom-in".
> 
> What do you mean by "what we know from zoom" that?

The icon you chose for the FAB is the icon for "zoom in". It is used to enlarge font size. You see it for example when you enlarge the size of a web page in chrome.


> According to the [material design doc](https://material.io/components/buttons-floating-action-button/#): "A floating action button (FAB) represents the primary action of a screen."

Yes, but Google Notes, Google Maps, and Google Calendar all have the search field and search button in the title bar, not as a FAB.


Thanks for the hint on how to share images – I didn't notice that GitHub feature :)

munen:
> The icon you chose for the FAB stands for "zoom in". It is used to enlarge font size. You see it for example when you enlarge the size of a web page in chrome.

I removed it in favor of a regular 'search' icon. The problem now is that the action is 'search' -> 'search' which is weird. An alternative would be 'filter' -> 'search'/'task-list', but I'm not so sure if people will recognize the filter action.

> Yes, but Google Notes, Google Maps, and Google Calendar all have the search field and search button in the title bar, not as a FAB.

Not on my iPhone, they don't. Calendar has search behind a burger menu. Notes and Maps don't have a header bar, but a hovering search bar. 

> Thanks for the hint on how to share images – I didn't notice that GitHub feature :)

You're welcome^^

munen:
I'm not very happy with having the Task List and the Search behind a collapsible FAB. It's a very bad compromise. But since you said off band that you're happy with having two separate modals with differerent feature sets, I'm willing to make this compromise for the time being.

For the long haul, I neither like the collapsible FAB and I don't like that they are two disjunct actions that behave very similarly and share a lot of code. It reduncant code, but more importantly it complicates the UI for the user. Additionally it differs on how Emacs behaves. 

But I do want search to land in the deployed version of organice rather sooner than later so that people can get some experience with it so that we can have a more substantiated discussion.

I'm merging this compromise with which we're both not happy, but on top of which we can practice in the real app and see what we like and what we dislike. 

Hence, this is not a decision to keep it this way. It's a spike on which we gain experience and then iterate on it. I sincerely hope that's in the best interest for all.

PR #207, opened 2020-01-03
Fix sort order in task list

### LINKED ISSUE #205: Filter DONE task states from Task List
The top tasks in the Task List in a 'real world' file are only tasks that are long DONE. For me that would be hundreds of DONE todos before seeing an actionable todo.

schoettl:
This may close #205 

Tasks in done state are sorted last now.

### DIFF
--- a/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
+++ b/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
@@ -101,7 +101,6 @@ function TaskListView(props) {
         return [earliestPlanningItem, header];
       })
       .sortBy(([planningItem, header]) => {
-        console.log(todoKeywordSets);
         const doneState = todoKeywordSets
           .flatMap(x => x.get('completedKeywords'))
           .includes(header.getIn(['titleLine', 'todoKeyword']));
@@ -127,7 +126,4 @@ const mapDispatchToProps = dispatch => ({});
 const getTimeFromPlanningItem = planningItem =>
   dateForTimestamp(planningItem.get('timestamp')).getTime();
 
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(TaskListView);
+export default connect(mapStateToProps, mapDispatchToProps)(TaskListView);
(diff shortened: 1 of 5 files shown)


src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
```diff
@@ -101,11 +101,15 @@ function TaskListView(props) {
         return [earliestPlanningItem, header];
       })
       .sortBy(([planningItem, header]) => {
+        console.log(todoKeywordSets);
+        const doneState = todoKeywordSets
```
munen: Could this be done with the `createIsTodoKeywordInDoneState` in parse_utils.js?

munen: I propose we add a test to document the functionality. Refactoring can be done with more certainty, too.

I'll add at least one.

schoettl: Right, should be possible to use that function. I thought I considered that... I probably did not use it because the documentation of this function was misleading. It has a mistake in the second line
...
if said `todoKeyword` is in `todoKeywordSets`.

I think it should be "if said `todoKeyword` is in any `todoKeywordSets`s done states."

munen: The test revealed that the current sort order for the `testOrgFile` in the integration test is:

1. A todo item with schedule and deadline
1. A header with a custom todo sequence in DONE state
1. A headline that's done a day earlier even
1. A headline that's done since a loong time

It's the same after using the function from `org_utils`, so I feel confident we can use it(;

> I think it should be "if said todoKeyword is in any todoKeywordSetss done states."

Right! I chaged it, thanks for the pointer!

munen:
Nicely done, I'll happily merge the PR^^

:pray:

munen:
I have updated the changelog with the attribution: https://github.com/200ok-ch/organice/commit/5562169831983743bf684a79e1a1fa2e3ec1e56a

Thanks, again^^

PR #48, opened 2019-09-21
WIP Planning items with following checkmarks are parsed correctly

munen:
Related issue: #47

Issue #47, opened 2019-09-21
Parser bug: SCHEDULED ahead of a checkmark isn't getting parsed properly


munen:
This will not properly get parsed:

``` org
 *** WAITING Proofing "Announcement für EmacsConf"
   SCHEDULED: <2019-09-16 Mon>
 - [X] Write draft announcement
```

It makes no difference if the checkmark is marked or not.

![image](https://user-images.githubusercontent.com/505721/65373179-692a2000-dc7a-11e9-90a8-6b1bf32984ce.png)

munen:
Fixed by https://github.com/200ok-ch/organice/pull/222

PR #222, opened 2020-01-11
Bugfix for timestamp parsing

schoettl:
I found a bug in the timestamp parsing: the day part was defined as `[^0-9]+`. This includes `>` and whitespace (including newlines).

The test in my first commit fails with:
```
    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

    - Tue
    + Tue>
    + - [ 
```

Since I cannot find a spec for orgmode timestamps, I just fix it with `[^0-9\s]`.


munen:
@schoettl Hell, yeah!

FYI, you just fixed PR https://github.com/200ok-ch/organice/pull/48 and related issue https://github.com/200ok-ch/organice/issues/47

Great catch and even greater PR. Happily pulling it in right away! :pray: :bowing_man:

munen:
I have updated the changelog with attribution: https://github.com/200ok-ch/organice/commit/2116e00df0f2ac979eda9146c3389143341a89c3

schoettl:
Good that you remembered the related issues!

munen:
Every now and then I would run into the bug again and would see it in the diffs. Great that nobody has to see that anymore :tada:

schoettl:
> Since I cannot find a spec for orgmode timestamps, I just fix it with `[^0-9\s]`.

PS: There is no character class for "all letters" in JS and `[a-zA-Z]` might not work for day names in other languages (if that is allowed), that is why I did not fix it with `[a-zA-Z]`

munen:
Org uses ISO 8601 (source https://orgmode.org/manual/Custom-time-format.html).

Hence, it can have localized day names (here's the [PR](https://github.com/DanielDe/org-web/issues/69) that added it to org web).

Issue #126, opened 2019-11-30
Validate capture templates


tru2dagame:
## Here is my capture template:

![image](https://user-images.githubusercontent.com/1673006/69906538-c1f70100-13ff-11ea-8b2a-c748431fe78a.png)

## When I try to use this template, I got this error:
![image](https://user-images.githubusercontent.com/1673006/69906543-ce7b5980-13ff-11ea-85fd-5704b162a31e.png)

munen:
Hi @tru2dagame 

Thank you for the report! It's a good catch - there is a validation missing here. At this point, organice assumes that you'll capture your entries into a header - hence, you'll have to specifiy one.

To be more concrete: Atm, you need to add a "Header path" for the capture template and this header needs to be set in your file. For example your file could look like:

```
* Inbox
```

Then your "header path" would just be "Inbox". And all your captures will go there.

There's three things we can improve here:

1. Trivial: Add documentation that this is required
2. Easy: Add a validation to the form
3. Medium: Refactor the feature so that it works without requiring a header for the capture entries

tru2dagame:
@munen Got it. Thanks. Hope we can improve to just add it without any requiring a header.

munen:
@schoettl I'd say https://github.com/200ok-ch/organice/pull/229 closes this issue. Rationale: Validating the capture template on input isn't necessarily better. Imo it's reasonable for a user to first want to set up a capture template and then adapt the Org structure. What do you think?

schoettl:
I'd vote for closing this, too.

Also, validation is not trivial, because capture templates can be defined for all files, while not all file might have the header defined in the capture template...

munen:
👍

PR #229, opened 2020-01-12
Bugfix: don't crash on using capture template when header path is invalid

schoettl:
I also replaced `!!targetHeader` with `targetHeader` because as far as I know, this `!!` doesn't make sense in many cases. An object never evaluates to false and null always eveluates to false. That is all we are testing here.

Or is there any good reason to keep it?

### DIFF
--- a/src/components/OrgFile/components/CaptureModal/index.js
+++ b/src/components/OrgFile/components/CaptureModal/index.js
@@ -84,9 +84,9 @@ export default ({ template, onCapture, headers, onClose }) => {
   useEffect(() => {
     if (textarea.current) {
       textarea.current.focus();
+      // Set Cursor position to the %? part of the template.
+      textarea.current.setSelectionRange(initialCursorIndex, initialCursorIndex);
     }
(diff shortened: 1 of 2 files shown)


munen:
Great fix! 

> Or is there any good reason to keep it?

Not at all. Your reasoning is spot on. I have seen people overuse it, because they are used to statically typed systems. So the notion to using data structures with their truthyness seems foreign - whilst the `!!` converts to a boolean. However, not only is it often not required, it often is senseless as in this example.

There's many more in this codebase. There is even 'weirder' checks like `if (!!headerId)`. Whenever you see it and are bugged by that, feel free to just get rid of it(;


Thank you for the fix! :pray: 

Now the user even get's nice feedback:

image


I'd say this PR closes https://github.com/200ok-ch/organice/issues/126. Rationale: Validating the capture template on input isn't necessarily better. Imo it's reasonable for a user to first want to set up a capture template and then adapt the Org structure. What do you think?

munen:
I have updated the changelog with attribution: 7861ecd

PR #232, opened 2020-01-13
Notify on changes for changelog

munen:
This is done with a similar UX than other productivity apps (like
Slack) do it. The user will see a 'gift' icon whenever there's a
change to the last changelog header which she has not seen, yet.


munen:
This has a bug recognizing changes. I’m aware of it and will push a fix out next. 

Apologies for shipping a broken feature🙏🏻

munen:
Fixed by: https://github.com/200ok-ch/organice/pull/233

PR #240, opened 2020-01-18
Fix relative time distance for future dates

### LINKED ISSUE #175: Relative dates (in words) in agenda are wrong, especially for future timestamps
In the agenda view when clicking on a date, the view toggles to relative dates, e.g. "3 days ago".

There are two bugs:

- For future dates, it also says "x days ago" while x is OK but the direction "ago" is wrong.
- For all timestamps without time, the difference is bad. E.g., at the time of this writing, 2019-12-29 is "about 23 hours ago". Better would be "today". Same problem for here: 2019-12-28 is "2 days ago".

schoettl:
Should close #175


munen:
@schoettl Thank you for your effort on this issue! :pray:

Issue #188, opened 2019-12-31
UX: Header toolbar causes content to change position


aspiers:
[This previously arose as discussion in #153 and #169; splitting out into its own issue for clarity.]

Tapping a header causes the toolbar to appear which allows editing the header, contents, deadline etc.  However because this toolbar appears above the selected header, it causes the header to shift vertically down, and then back up again if a different header further down is selected.  (The latter would also happen in the future if deselection of the header is ever implemented as per #152.)

In #153, [**@munen**](https://github.com/munen) commented on [December 25, 2019 9:08 AM](https://github.com/200ok-ch/organice/issues/153#issuecomment-568870778 "2019-12-25T09:08:14Z - Replied by Github Reply Comments"):
> This is a recent change. Before, the action bar was below the header. Meaning, it was in between the header and its content. People rightfully complained that the action bar was now in the way in between related content. Hence, we moved it on top of the header. A few weeks later, you come along^^

I believe the issue tracking that change was #100.  I agree with the referenced concerns from EmacsConf which it was trying to address; however I'm not at all convinced that the change implemented was the right one, and it seems I'm not the only one.  Going back to #153, [**@schoettl**](https://github.com/schoettl) replied on [December 26, 2019 9:42 AM](https://github.com/200ok-ch/organice/issues/153#issuecomment-569025279 "2019-12-26T09:42:09Z - Replied by Github Reply Comments"):
> To be honest, I also liked it better when the toolbar was below the headings. My reason for that is:
> 
> *   I often click on a header to see its sub headers
> *   Then I want to "close" and hide the sub headers again (because I've seen what I need)
> *   To hide them, now I have to click on a different point in the UI
> 
> So, placing the toolbar above the header line makes the important UI element (the header line) jumping around. That was also unintuitive for my dad ;)
> 
> I just tested [org-web.org/sample](https://org-web.org/sample) and really like that more. One finger inbetween header line and content is no problem for me and better than a jumping header line.

Unfortunately I can't immediately offer a perfect solution, but hopefully we can all work together in this issue to design one.  My current thinking is that if there are problems putting the toolbar below the header (as per #100) but also above it (as per #153, #169, and this issue), then maybe we need to find an entirely different approach.  For example:

- Make it a lot smaller by changing `DEADLINE` / `CLOCK IN` / `SCHEDULED` into icons.  I can't see any good reason why those buttons should be text but the others should be icons; am I missing something?
[...]

PR #254, opened 2020-02-14
WIP feat: double tap titles and descriptions to edit

williamconnolly:
As a user, I kept finding myself double-tapping titles and descriptions to edit them. I think it comes from me being used to a file explorer's "double click file to rename it" functionality. 

Also, on my tablet (and desktop screen as well), the body of the application was too narrow. Rather than 35em, I changed to 80% and found it was more usable. I'm open to removing this of course, if others' opinions differ.

### DIFF
--- a/bin/compile_search_parser.sh
+++ b/bin/compile_search_parser.sh
@@ -6,5 +6,10 @@ npx pegjs -o src/lib/headline_filter_parser.{js,grammar.pegjs}
 # code that we're writing and we want to keep eslint warnings at 0,
 # we're telling eslint to ignore the generated file.
 PARSER_FILE=src/lib/headline_filter_parser.js
+TMP_FILE=/tmp/headline_filter_parser.js
+
 # Prepend the generated parser code with a `eslint-disable` comment.
-sed -i '1i \/* eslint-disable */' "$PARSER_FILE"
+echo '/* eslint-disable */' > "$TMP_FILE"
+cat "$PARSER_FILE" >> "$TMP_FILE"
+mv "$TMP_FILE" "$PARSER_FILE"
+
--- a/bin/compile_search_parser.sh
+++ b/bin/compile_search_parser.sh
@@ -6,7 +6,5 @@ npx pegjs -o src/lib/headline_filter_parser.{js,grammar.pegjs}
 # code that we're writing and we want to keep eslint warnings at 0,
 # we're telling eslint to ignore the generated file.
 PARSER_FILE=src/lib/headline_filter_parser.js
-
 # Prepend the generated parser code with a `eslint-disable` comment.
 sed -i '1i \/* eslint-disable */' "$PARSER_FILE"
-
--- a/bin/compile_search_parser.sh
+++ b/bin/compile_search_parser.sh
@@ -6,10 +6,7 @@ npx pegjs -o src/lib/headline_filter_parser.{js,grammar.pegjs}
 # code that we're writing and we want to keep eslint warnings at 0,
 # we're telling eslint to ignore the generated file.
 PARSER_FILE=src/lib/headline_filter_parser.js
-TMP_FILE=/tmp/headline_filter_parser.js
 
 # Prepend the generated parser code with a `eslint-disable` comment.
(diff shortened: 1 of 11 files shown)


bin/compile_search_parser.sh
```diff
@@ -6,5 +6,10 @@ npx pegjs -o src/lib/headline_filter_parser.{js,grammar.pegjs}
 # code that we're writing and we want to keep eslint warnings at 0,
 # we're telling eslint to ignore the generated file.
 PARSER_FILE=src/lib/headline_filter_parser.js
+TMP_FILE=/tmp/headline_filter_parser.js
+
 # Prepend the generated parser code with a `eslint-disable` comment.
-sed -i '1i \/* eslint-disable */' "$PARSER_FILE"
+echo '/* eslint-disable */' > "$TMP_FILE"
```
williamconnolly: This sed command failed to execute on my computer. I have had issues with sed incompatibilities with linux/macOS etc, so for this specific action, "pre-pend a line to the file", I got it working this way. I'm specifically on macOS Mojave  10.14.15. I open to reverting this change as well.

schoettl: Can you tell me the output of `sed --version`? Maybe you can just install a modern GNU sed on your system?

I'd generally prefer this sed one-liner. If it have to be using a temp file, it should be done with `mktemp`, which is also hard to make platform-independent. But a static temp file can randomly make problems to some users, e.g. when they have a local CI system running.

schoettl:
I tested it, but double clicks to edit doesn't seem to work. I tried in chromium, normal mode and mobile emulation mode.

What is the behavior for a normal click, now? Still collapse/expand header? Isn't this delayed, when double clicks are also handled?

Having a width of 80 % makes the toolbar look not good. But it may indeed be more usable. And if we change the UI concept anyway (material design), it's maybe OK.

munen:
I tested it, too. The feature does work for me (after enabling it in the settings) - tested on FF (Linux) and iOS.

However, I'm not convinced of the UX of this feature atm. Since clicking a header will toggle an expand/collapse, a double click will almost always miss it's mark. I've created a demo with three headlines. I'm always double clicking, but it almost always doesn't work, because of expanding/collapsing headlines.

demo

Apart from that, on a mobile device, it's easy to trigger any action from the `HeaderActionBar` on a 'double-tap'. The first tap will open the header up and the second will click anything.

As of this state, I'm afraid we cannot merge it. @williamconnolly if you can address these UX features, this assassment can change, of course. Personally, I wouldn't know how to solve it, though. It's hard to do two things with one action.

Apart from the UX, the code is nicely done. Good job on building your first neat PR! :pray: 

I'm very sorry I don't have better news on the review, though.

williamconnolly:
Sounds good, I'll look into fixing this. I started compensating for the quick-header-collapse behavior and got consistent title double-taps when the titles were scrolled high enough in the body - I don't yet have a solution for this. All online tutorials speak of debouncing the taps for collapsing, to give room for onDoubleTap to fire. Thanks for feedback. What would you think instead of Tap-And-Hold?

bin/compile_search_parser.sh
williamconnolly: Ahh, I see, the local CI system that may not have a tmp dir is a good argument. I'll futz around with `brew install gsed` or something - I know that macOS default unix utils are woefully out of date.

munen:
"Tap-And-Hold" sounds good for editing the title.

Having said that, at some point "Tap-And-Hold" might trigger a new context menu which opens non-primary actions for headers. There has been previous discussion going on around this. The rationale is that the current `HeaderActionBar` is already filled with actions and it'll be an issue adding more to that. The whole paradigm around how to trigger actions against headlines may change in the future. There's nothing decided, though. This is an issue where some open discussion has happened: https://github.com/200ok-ch/organice/issues/188

bin/compile_search_parser.sh
munen: :+1:

williamconnolly:
Closing for now as I dont have much time to complete due to obvious circumstances

munen:
All the best for your other commitments! :pray:

PR #271, opened 2020-04-04
Add support for org-log-into-drawer TODO state logging

cpbotha:
Orgmode has a setting [org-log-into-drawer](https://www.gnu.org/software/emacs/manual/html_node/org/Tracking-TODO-state-changes.html) which logs TODO state changes into the `:LOGBOOK:` drawer by default.

This is especially useful for org-habits, where this list of repeating `TODO -> DONE` state changes can become quite long. (`:LOGBOOK:` is by default collapsed in organice and in orgmode proper)

I've added a new persistent setting (search for `shouldLogIntoDrawer` in the code). This is by default OFF so the old behaviour of putting these changes in the body of the element is currently the default.

In other words, existing users will see nothing different until they try this new setting.


cpbotha:
`yarn test` is failing on circleci because: "[BABEL] /home/circleci/repo/src/components/Landing/Landing.test.js: Cannot find module '@babel/helper-validator-identifier'"

On my dev machine, I'm seeing all tests passing:

```
Test Suites: 10 passed, 10 total
Tests:       236 passed, 236 total
Snapshots:   3 passed, 3 total
Time:        7.18s
Ran all test suites.
```

Could it be a misconfiguration on the circleci?

schoettl:
My local tests fail with similar errors `[BABEL] ... Cannot find module 'js-levenshtein'`.

Thanks for your PR! Feedback:

- Would you rename the settings header to something "human-readable" to conform with the other settings? Like your first sentence of the setting description.
- I activated the setting but changing the TODO state of a header (without existing logbook), does not create the logbook. I haven't tested, if it works with an existing logbook.

munen:
Curious regarding the failing tests. I tried to reproduce the issue on my machine with a new installation of organice - but the tests still pass.

The only difference I see in local vs CIt: We have no `yarn` pinning. On CircleCI, we use an older yarn version (1.19.1) whilst I'm using 1.22.4 locally. In theory, they could use different algorithms for resolving dependencies. Having said that, we've been using the same CircleCI setup for months. And the tests broke on CI without us introducing changes to the dependencies.

@schoettl That the tests also fail on your machine, this could be helpful(; What version of yarn are you using?

munen:
I'm going to SSH into the container running the tests on CircleCI to see if I can learn more.

schoettl:
> @schoettl That the tests also fail on your machine, this could be helpful(; What version of yarn are you using?

With system node:
```
$ nvm use system
Now using system version of node: v13.12.0 (npm v6.14.4)
$ yarn test
yarn run v1.22.4
error organice@0.2.0: The engine "node" is incompatible with this module. Expected version "^12.13.1". Got "13.12.0"
error Commands cannot run with an incompatible environment.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

With other node:
```
$ nvm use
Now using node v12.13.1 (npm v6.12.1)
$ yarn test
...
[BABEL] /home/jakob/projects/organice/src/reducers/org.unit.test.js: Cannot find module 'js-levenshtein'
...
```

Hope that helps.

munen:
My hunch could be right. After forcing an update of `yarn` in the SSH session, the tests run on CircleCI: 

image

munen:
@schoettl Thanks a lot, it does help!^^

munen:
Tests on CI are green, again: https://github.com/200ok-ch/organice/pull/274

Now, on to the PR at hand.

munen:
I merged `master` into the PR to get tests on CI to be green.

cpbotha:
> * I activated the setting but changing the TODO state of a header (without existing logbook), does not create the logbook. I haven't tested, if it works with an existing logbook.

The existing behaviour is to log TODO state changes only for repeating tasks. See `reducers/org.js::updatePlanningItemsWithRepeaters` -- on master, this is the only place where state change is logged.

In other words, it's not about there being a logbook or not, it is about whether it's a repeating task or not (with scheduled set).

I will change the wording in settings to reflect this.

(If we want general TODO state logging, that would be a separate project.)

munen:
And I tested the feature locally. Unfortunately, it was not a successful test. I made a short screencast of the testing procedure:

demo

What did I do wrong?

cpbotha:
> What did I do wrong?

:) Please see my comment https://github.com/200ok-ch/organice/pull/271#issuecomment-609408224

munen:
I've added your contribution to the changelog: https://github.com/200ok-ch/organice/commit/7db53a19e9de5c0a4fe05e683689584384fad4c2

Thank you very much for your time and effort :pray: :bowing_man:

PR #287, opened 2020-05-04
Bugfix for #286: append new tags correctly to title

### LINKED ISSUE #286: new tags gone after browser refresh
**Describe the bug**
The tag icon works beautifully in allowing me to assign selected newly created tags to each header. I really enjoy the display of the assigned tags. 
However, when the browser is refreshed, these tags become part of the header. For example, "this is header 1:TAG:"
When touching the tag icon again, it doesn't recognize this head has any tag at all. And the previously created tag list no longer exists. New tags have to be created. 
Each browser refresh appends the newly assigned tag to the end of the header.

**To Reproduce**
1. Assign new tags using the tag icon
2. Save 
3. Refresh browser

**Expected behavior**
I would expect the tag function to work across browser reloads.

 - iPhone 11
 - IOS 13.4.1
 - Browser: DuckDuckGo

schoettl:
This closes #286 

New tags were appended without a space before. Always inserting a space would change the original org file (parse-export-diff).

Solution: Only if title does not end with space, insert correct number of spaces (at least one)

### DIFF
--- a/src/lib/export_org.js
+++ b/src/lib/export_org.js
@@ -225,20 +225,28 @@ export const attributedStringToRawText = parts => {
 
 // Takes a plain JS object
 export const generateTitleLine = (header, includeStars) => {
-  let contents = '';
-  if (includeStars) contents += '*'.repeat(header.nestingLevel);
+  // Remove stars afterwards if not includeStars - because of :tag: alignment
+  let titleLine = '*'.repeat(header.nestingLevel);
 
   if (header.titleLine.todoKeyword) {
-    contents += ` ${header.titleLine.todoKeyword}`;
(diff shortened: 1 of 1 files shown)


schoettl:
Hi @munen,

I have again minor problems with prettier:

```
$ npx prettier --version
1.19.1
$ prettier --version
2.0.5
```

- The newer system prettier wraps every single lambda arg in parens `(x) => x`.
- `npx prettier` does the opposite.

I personally think, the parens shouldn't be mandatory. What do you think?

I guess, we should edit `.prettierrc.json` to fix this version difference.

munen:
@schoettl Great PR, as always!

I'll merge, add the changelog entry and will open a separate issue so we can discuss your absolutely legit input regarding the prettier situation.

PR #289, opened 2020-05-05
chore: Use new prettier default to always have parens in arrow fns

munen:
I've skimmed the changes from prettier for mistakes or outliers. In very few occasions, I have reverted a chunk to the previous indentation, but generally speaking it looks like this PR doesn't introduce a bad style regression.

Feel free to make any changes, close or merge^^


munen:
>  Did you sneak in some easter eggs? ^^

😁

PR #293, opened 2020-05-06
Fix problem with indent

### LINKED ISSUE #292: Planning items line does not respect the new hard-indent setting
**Describe the bug**
When the option "Disable hard-indent" is set in the settings, new planning items are still indented. Even if I edit the description and remove the space manually and click "Done", it stays indented.

schoettl:
This wants to close #292

### DIFF
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -64,8 +64,10 @@ const doSync = ({
   successMessage = 'Changes pushed',
   shouldSuppressMessages = false,
 } = {}) => (dispatch, getState) => {
-  const client = getState().syncBackend.get('client');
-  const path = getState().org.present.get('path');
+  const state = getState();
+  const presentOrgState = state.org.present;
+  const client = state.syncBackend.get('client');
+  const path = presentOrgState.get('path');
   if (!path) {
     return;
   }
@@ -78,7 +80,7 @@ const doSync = ({
   // recursively enqueue the request to do a sync until the current
   // sync is finished. Since it's a debounced call, enqueueing it
   // recursively is efficient.
-  if (getState().base.get('isLoading')) {
+  if (state.base.get('isLoading')) {
     // Since there is a quick succession of debounced requests to
     // synchronize, the user likely is in a undo/redo workflow with
     // potential new changes to the Org file in between. In such a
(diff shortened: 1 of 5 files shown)


src/components/OrgFile/components/HeaderContent/index.js
```diff
@@ -78,8 +78,10 @@ class HeaderContent extends PureComponent {
   }
 
   calculateRawDescription(header) {
-    // this is for display only, se we keep the default indentation behaviour
-    return createRawDescriptionText(header, false, false);
+    // This generates the text that appears in the description text field.
+    const dontIndent = this.state.base.get('shouldNotIndentOnExport');
+    // const dontIndent = this.props.dontIndent;
```
schoettl, line 83: @munen would you help me with this:
- I added `dontIndent` to the props (see below `mapStateToProps`).
- I can use the new prop (commented line)

But why can't I use the state directly? In the same file, `this.state` is used in many places. But using it here, I get an error, state is undefined. Maybe I used to understand this already, but I forgot it ^^

schoettl:
@cpbotha Thanks again for your setting "dontIndent", I use this feature!

Just for your information: I found a bug #292 and fixed it with the last WIP commit. It's currently failing because I have to switch two lines – but it's a pending question for me.

src/lib/export_org.js
```diff
@@ -259,13 +259,13 @@ export const generateTitleLine = (header, includeStars) => {
  * @param {boolean} dontIndent Default false means indent drawers according to
  * nesting level, else keep everything flush-left. Description is kept as is.
  */
-export const exportOrg = (
+export const exportOrg = ({
```
munen, line 262: I like this change. Less PLOP :smile:

src/components/OrgFile/components/HeaderContent/index.js
munen: @schoettl I'll happily try to help(; You cannot access `this.state` in `calculateRawDescription` for two reasons:

1. The local `state` is defined dependent on the `calculateRawDescription`: https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/components/HeaderContent/index.js#L46-L47
2. Only functions that are bound have access to `this` of the component (see https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/components/HeaderContent/index.js#L25).

Having said that, I don't think that you want to access the local state of the component, anyway. It's not the right place containing the information whether or not `dontIndent` is set. The `props` are the right place for that and you've already set that up correctly.

So, if I'm not mistaken, you want to use the props and not the state(;

schoettl: Thanks! That exactly answers my question.

src/actions/org.js
```diff
@@ -64,8 +64,10 @@ const doSync = ({
   successMessage = 'Changes pushed',
   shouldSuppressMessages = false,
 } = {}) => (dispatch, getState) => {
-  const client = getState().syncBackend.get('client');
-  const path = getState().org.present.get('path');
+  const state = getState();
+  const presentOrgState = state.org.present;
```
munen, line 68: @schoettl I can absolutely understand why it looks like a good idea to minimize the duplication of typing `state.org.present` all the time! 

Having said that, to my knowledge, it's an anti-pattern to hide the structure of the reducers data shape. Maybe anti-pattern is too strong a word and maybe I'm wrong. But to the best of my knowledge, the data shape in a redux reducer is akin to a 'table' in SQL[1]. Hence, the same structure is repeated in all places in the code where a query happens. If we hide the structure here, it adds cognitive load that the structure in this scope is called differently than in other scopes.

Of course, I might be wrong and can be persuaded that this is indeed a good practice. Until then, I feel pretty certain it's better to repeat the structure in the code.

1. See 'key concepts': https://redux.js.org/recipes/structuring-reducers/prerequisite-concepts

src/components/OrgFile/components/HeaderContent/index.js
munen: Happy to be of help :pray:

src/actions/org.js
schoettl: OK, I revert this. I understand the point.

munen: :pray:

schoettl:
@munen Ready for final review and merge, I think.

munen:
@schoettl Looks good to me! I'll happily merge and add the fix to the changelog.

Thank you for the good work :pray:

PR #298, opened 2020-05-14
Make script Mac-compatible

schoettl:
MacOS has no decent sed installed, but ex editor seem to work.

Ready to merge after review.

### DIFF
--- a/bin/compile_search_parser.sh
+++ b/bin/compile_search_parser.sh
@@ -8,4 +8,6 @@ npx pegjs -o src/lib/headline_filter_parser.{js,grammar.pegjs}
 PARSER_FILE=src/lib/headline_filter_parser.js
 
 # Prepend the generated parser code with a `eslint-disable` comment.
-ex -sc '0s/^/\/* eslint-disable *\//|x' "$PARSER_FILE"
+# We don't use sed or ex because some Linux/UNIX distros don't support them.
+tmp=$(mktemp)
+echo '/* eslint-disable */' > "$tmp" && cat "$PARSER_FILE" >> "$tmp" && mv "$tmp" "$PARSER_FILE"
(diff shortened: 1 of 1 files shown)


munen:
Thank you very much for the PR!

I'm writing this from the Mac, because I wanted to test the PR here and then on my Linux machine. Only now, I realize that the CI server failed, because `ex` is not installed. Of course, we could install `ex` there, but I don't know if that improves the overall situation. It seems it's not a default for all Linux distributions - otherwise CI wouldn't have failed (where we didn't have to install `sed` manually).

I see two 'sustainable' ways forward:

1. Use a tmp file instead of sed
2. Find a check that works for 'GNU sed' and ask people to install it if they don't have it

Option 1 seems dirty for someone used to GNU sed, but it's probably the more user friendly.

What do you think?

schoettl:
OK, let's see if this works. I deliberately did not amend the previous commit to retain the history.

However, I think I darkly remember (?), that `mktemp` is special, too. Not sure, if it works on all distros and Mac without a template string. But the template string is a problem, too, because some distros might not have a `/tmp` dir.

schoettl:
Just from `mktemp`s synopsis on a Mac OS man page, the template or a prefix is required. So the current solution may not work under Mac.
https://ss64.com/osx/mktemp.html

And I don't know, if we can trust in that every possible build server has `/tmp`... I think that was the reason why I originally went with `sed -i`.

Using a "static" temp file in the current directory is also a bad idea because of possible concurrent builds or manual calls.

munen:
Tested under MacOS Majave 10.14.6 with stock BSD `mktemp`.

munen:
Tested under Debian Linux.

Tests are also green, hence it works on CI, too.

schoettl:
OK, let's merge then :)

munen:
> OK, let's merge then :)

Indeed!

I just wanted to add a comment with regards to `/tmp`. Afaik 'all' currently used Unix flavors have a `/tmp` folder - they just behave differently. Some might survive a reboot, others will not.

Let's see what the first Windows contributor will do. But if she's using WSL or cygwin or a VM, then she'll be good anyway.

Good job, @schoettl! :pray:

PR #301, opened 2020-05-25
Couple more minor doc tweaks

aspiers:
- Fix formatting of references to doc files
- Sanity check for building docs from the right directory


munen:
:+1:

PR #302, opened 2020-05-27
Doc/android install

shombando:
I staged this tiny PR into three smaller commits with increasing amount of commentary and opinion. Please pick any or all commits that make the most sense for the project.


munen:
Great addition to the docs, I'm happily merging all of it!

Welcome to the contributors list :pray:

PR #306, opened 2020-05-29
More doc improvements

aspiers:
A ragbag of doc improvements; should be self-explanatory from each commit message.

### DIFF
--- a/README.org
+++ b/README.org
@@ -59,17 +59,48 @@ capabilities]].
 
 Installing to the homescreen on:
 
-- iOS: Open organice in Mobile Safari. Tap the "share" button and
-  select "Add to Home Screen".
-- Android: Open organice in Chrome or Firefox. Tap the "menu" button
-  and select "Add to homescreen" or "Install". Some browsers may also
-  present a popup banner with the option to "Add to homescreen" or
-  "Install".
-  Note: Any file can be added as a separate icon on the homescreen, so 
-  several organice app icons can be added to get direct access to individual 
-  files.
+- *iOS:*
 
+  Open organice in Mobile Safari. Tap the "share" button and select
+  "Add to Home Screen".
+
+- *Android:*
+
+  The exact procedure may differ depending on your browser and Android
+  version.  If you discover improvements to the following procedure,
+  please [[#contributing][let us know]]!
+
+  First open the organice web page in your mobile browser.
+
+  - On Chrome, tap the "menu" button (three vertically stacked dots)
+    and select "Add to homescreen".
+
+  - On Firefox, tap the home icon with the plus sign inside it which
+    is immediately to the right of the URL in the address bar.
+
+  - Other browsers may have a similar procedure to one of these.
+
+  At this point, most browsers will present a popup banner with the
+  option to "Add to homescreen" or "Install".
+
+  If you want the organice app icon on your home screen to jump to a
+  specific =.org= file, and/or if you want multiple app icons on the
+  home screen to jump to different files, things can get a bit more
+  tricky.  For instance, Chrome only offers "Add to homescreen" the
+  first time; once it has been added, the option changes to "Open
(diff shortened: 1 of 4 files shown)


README.org
```diff
@@ -59,10 +59,23 @@ capabilities]].
 
 Installing to the homescreen on:
 
-- iOS: Open organice in Mobile Safari. Tap the "share" button and
+- **iOS:** Open organice in Mobile Safari. Tap the "share" button and
```
munen, line 62: Here's a conflict to the PR that was just opened before yours: https://github.com/200ok-ch/organice/pull/302

bin/compile_doc.sh
```diff
@@ -5,6 +5,11 @@ if ! [ -e README.org ]; then
     exit 1
 fi
 
+if ! which pandoc >/dev/null 2>&1; then
```
munen, line 8: :+1:

bin/compile_doc.sh
```diff
@@ -14,3 +19,11 @@ pandoc CODE_OF_CONDUCT.md -o CODE_OF_CONDUCT.org
 cat CODE_OF_CONDUCT.org >> documentation.org
 rm CODE_OF_CONDUCT.org
 emacs documentation.org -l ./doc/org2html/init.el --batch  --funcall org-html-export-to-html
+
```
munen, line 22: :+1:

README.org
```diff
@@ -59,10 +59,23 @@ capabilities]].
...
   select "Add to Home Screen".
-- Android: Open organice in Chrome or Firefox. Tap the "menu" button
-  and select "Add to homescreen".
+
+- **Android:** Open organice in Chrome or Firefox. Tap the "menu"
+  button (three vertically stacked dots) and select "Add to
+  homescreen".
+
+  However, you can only do this once.  If you want multiple organice
```
munen, line 69: I don't know if that's relevant doc for organice. iOS and Android move so fast. Also, Android is quite fragmented. To me, it sounds like a piece of doc that in the best case might be correct now, but could soon be outdated. In any case, I cannot keep it up to date as I'm not an Android user.

@schoettl What do you think? Iirc you use Android. If you think my impression is wrong, I'll happily defer to you!

README.org
aspiers: OK, fixing.

README.org
aspiers: Android doesn't change *that* fast :stuck_out_tongue_closed_eyes:  But the current text in `master` about multiple icons actually looks somewhat inaccurate to me on both Chrome and Firefox:

- On Chrome, I'm pretty sure there is no way to add multiple icons without a third-party app; at least, there is no obvious way I can find.
- On Firefox, there are two different ways to install:
    1. The "home/plus" icon in the address bar, but this creates an app icon which launches the app, and doesn't jump directly to the page from which the icon was created.
    2. Menu (three dots) > Page > Add Page Shortcut, which *does* create an icon linking to the current page, but it's just a browser bookmark which launches the browser not the frameless PWA.

Of course, any/all of this might be different for different versions of Chrome / Firefox / Android.

However, regarding the debate of what to include, let's consider two scenarios:

- We keep the documentation minimal, for fear of the information we give being wrong or becoming out of date.  Result: users are completely on their own, and might incorrectly conclude that it's impossible to achieve one or more app icons which jump directly to specific pages.
- We give a best attempt to document the current state, highlighting that the info may be incomplete or out of date, and asking readers to let us know if so.  Result: the docs are probably helpful for some/most Android users, and in the worst case, readers still know that it *was* possible to achieve multiple/specific icons in some circumstances, and are encouraged to work with us to improve the docs where needed.  With luck the existing docs may even provide them with the hints / ideas they need to figure it out.

I prefer the second scenario, so I've force-pushed a new version which takes this approach.  Hope you like it!

munen:
👍 

Good call on the `CUSTOM_ID`, too!

PR #307, opened 2020-05-29
Make local file:... links navigate to the relevant file

aspiers:
Make a link like `[[file:foo/bar.org][bar]]` into `href="foo/bar.org"` without `target="_blank"`, so that it opens in the same organice window.  This allows more convenient navigation around a network of `.org` files, and in particular an `index.org` file which facilitates quick jumping to the most commonly used other `.org` files.

### DIFF
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -20,9 +20,14 @@ export default ({ parts, subPartDataAndHandlers }) => {
           case 'link':
             const uri = part.getIn(['contents', 'uri']);
             const title = part.getIn(['contents', 'title']) || uri;
-
+            const isFileLink = uri.startsWith("file:");
             return (
-              <a key={part.get('id')} href={uri} target="_blank" rel="noopener noreferrer">
+              <a
+                key={part.get('id')} 
+                href={isFileLink ? uri.substr(5) : uri}
+                target={isFileLink ? "" : "_blank"}
+                rel="noopener noreferrer"
+              >
                 {title}
               </a>
             );
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -216,7 +216,7 @@ describe('Render all views', () => {
 
     describe('Renders everything starting from an Org file', () => {
       test('renders an Org file', () => {
-        expect(getAllByText(/\*/)).toHaveLength(6);
+        expect(getAllByText(/\*/)).toHaveLength(7);
         expect(container).toMatchSnapshot();
       });
 
@@ -449,6 +449,7 @@ describe('Render all views', () => {
           expect(elem[0]).toHaveAttribute('href', 'tel:+498025123456789');
           expect(elem[0]).toHaveTextContent('+498025123456789');
         });
+
         test('recognizes URLs', () => {
           fireEvent.click(
             queryByText('A header with a URL, mail address and phone number as content')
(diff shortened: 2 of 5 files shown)


aspiers:
I looked at writing tests for this, but got completely confused by the large chunk of test code duplicated between two files.  Here is just a small subset of the duplication:

https://github.com/200ok-ch/organice/blob/591b13c0e7a1cdd227644d22729ced9965e3bdc7/src/lib/parse_org.unit.test.js#L104-L124

vs.

https://github.com/200ok-ch/organice/blob/591b13c0e7a1cdd227644d22729ced9965e3bdc7/src/components/OrgFile/OrgFile.unit.test.js#L233-L255

What's going on here?

schoettl:
I'm confused, too. I remember that I wrote many of these tests. But I certainly duplicated this code. As far I remember, I worked mostly in parse_org.unit.test.js.

Is the second one some auto-generated artefact? I can check the commit history later...

src/components/OrgFile/components/AttributedString/index.js
```diff
@@ -20,12 +20,19 @@ export default ({ parts, subPartDataAndHandlers }) => {
...
-            );
+            if (uri.startsWith("file:")) {
+              return (
+                <a key={part.get('id')} href={uri.substr(5)} rel="noopener noreferrer">
+                  {title}
+                </a>
+              );
+            } else {
+              return (
+                <a key={part.get('id')} href={uri} target="_blank" rel="noopener noreferrer">
```
schoettl, line 31: I'm pretty sure there is a way to simplify this. I think there must be a syntax to make rendering an HTML attribute depend on a bool `isFileLink`. And with a `uri = isFileLink ? uri.substr(5) : uri` or something we could avoid the duplication of the three lines of `<a>`.

munen: Yes, that sounds like a good refactor.

aspiers: OK sure.  I didn't go for that initially because I didn't know how to do that whilst avoiding an empty `target=""` or even `target` attribute due to `target={isFileLink ? "" : "_blank"}`, but I guess that tiny cosmetic issue hidden in the HTML source is still preferable to duplicated code.

aspiers:
[**@munen**](https://github.com/munen) commented on [May 29, 2020 3:43 PM](https://github.com/200ok-ch/organice/pull/307#issuecomment-636012238 "2020-05-29T14:43:13Z - Replied by Github Reply Comments"):
> Good catch on the duplication of tests. Complete duplications can definitively go.
> 
> Feel free to pick and chose the correct location depending on the context of the test.

I don't really know enough about the tests yet to make the call on this :thinking:

munen:
Sweet addition. It works well for me on a first try!

Apart from the duplication of the tests, do you plan to write an integration test for this?

aspiers:
Duplication fixed in #309.  I'll look at writing a test on top of that PR.

aspiers:
New version pushed with integration test.

schoettl:
Nice!

How should Organice handle non-org file links? I think I didn't see a test like `uri.match(/\.org$/)` in the changes...

What does Organice do, if we try to open a non-org file?

munen:
Sweet, great addition to the test-suite!

Good job, @aspiers! :pray: 

@schoettl 

> What does Organice do, if we try to open a non-org file?

It'll show a message that it cannot parse the file or that there was an error downloading the file. In any case, the user can go back to the original file.

munen:
@aspiers I've added the attribution to the changelog: 67e1390

Great to see you contributing new features with tests, now! :pray: :bow:

munen:
@aspiers (cc @schoettl): I deeply apologize, but I have merged this PR too eagerly without proper testing. After merging, I found multiple bugs immediately with using my main Org files (which have file links). I have opened an issue for that: https://github.com/200ok-ch/organice/issues/311

Tbh, the three bugs I found should all have been merge blockers. Of course it's on me, that I didn't test properly. But if we can't fix those soon, I'm inclined to revert the merge. It's very easy to get organice into a broken state with the current implementation. And, on `master`, we usually try to include only features that work. They may be a subset of what the feature could be, but they should behave consistently. This new feature, unfortunately, doesn't.

I'm afraid the implementation would have to be changed to:

- Parse the Org file with `file://` links
- On clicking such a link, create an event
- Then, in a new function that will have to be implemented for all sync back-ends, check if that file can be reached
  - If so, open it (not by redirecting, but through an action/reducer)
  - If not, show a message to the user that this file is not reachable through the sync-backend

Of course, there might be other options to implement this feature. I'm just writing this solution down here, because that's how I envisioned the feature before. Since it's not trivial, I haven't gotten around to building it.

PR #308, opened 2020-05-29
doc: Add a button on the landing page linking to the docs

aspiers:
As discussed on Matrix.  Not sure whether to ditch the `.view-sample-button` CSS rule; to me it looks best with the same `margin-top` as the other two buttons.


munen:
Definitively a great change, I'll happily merge this in!

The magenta color indicates the main call to action. That's why I made a minor adjustment and:

- Moved the "sign in" to the top, so that it will be found more easily
- Colored the 'doc' links 'sample' and 'documentation' differently

PR #309, opened 2020-05-30
Remove unit tests duplicated in parse_org.unit.test.js

aspiers:
There were a bunch of tests identically or near identically duplicated between `parse_org.unit.test.js` and `OrgFile.unit.test.js`, so remove them from one of the two files.  The former seems to be a more specific and low-level set of tests; based on the file name presumably it's the best place for testing `parseOrg()`, but probably less suitable for tests which also exercise `exportOrg()`.

### DIFF
--- a/src/components/OrgFile/OrgFile.unit.test.js
+++ b/src/components/OrgFile/OrgFile.unit.test.js
@@ -366,7 +366,7 @@ ${description}`;
             expect(exportedFile).toEqual(testOrgFile);
           });
 
-          test('For basic files', () => {
+          test('For files with multiple headlines with timestamps', () => {
             const testOrgFile = readFixture('multiple_headlines_with_timestamps');
             const exportedFile = parseAndExportOrgFile(testOrgFile);
             expect(exportedFile).toEqual(testOrgFile);
--- a/src/components/OrgFile/OrgFile.unit.test.js
+++ b/src/components/OrgFile/OrgFile.unit.test.js
@@ -9,6 +9,22 @@ import readFixture from '../../../test_helpers/index';
 import { noLogRepeatEnabledP } from '../../reducers/org';
 import { fromJS } from 'immutable';
 
+/**
+ * This is a convenience wrapper around parsing an org file using
+ * `parseOrg` and then export it using `exportOrg`.
+ * @param {String} testOrgFile - contents of an org file
+ * @param {Boolean} dontIndent - by default false, so indent drawers
+ */
+function parseAndExportOrgFile(testOrgFile, dontIndent = false) {
+  const parsedFile = parseOrg(testOrgFile);
+  const exportedFile = exportOrg({
+    headers: parsedFile.get('headers'),
+    linesBeforeHeadings: parsedFile.get('linesBeforeHeadings'),
+    dontIndent: dontIndent,
+  });
+  return exportedFile;
+}
+
 describe('Tests for export', () => {
   const createSimpleHeaderWithDescription = (description) =>
     fromJS({
--- a/src/components/OrgFile/OrgFile.unit.test.js
+++ b/src/components/OrgFile/OrgFile.unit.test.js
@@ -66,20 +66,6 @@ describe('Tests for export', () => {
 });
 
 describe('Unit Tests for Org file', () => {
-  describe('Test the parser', () => {
-    const expectType = (result) => expect(result.map((x) => x.type));
-    describe('Parsing inline-markup', () => {
-      test('Parses inline-markup where closing delim is followed by ;', () => {
-        const result = parseMarkupAndCookies('*bold*;');
-        expectType(result).toEqual(['inline-markup', 'text']);
-      });
-      test('Parses inline-markup surrounded by text', () => {
-        const result = parseMarkupAndCookies(' *bold*;');
-        expectType(result).toEqual(['text', 'inline-markup', 'text']);
-      });
-    });
-  });
-
   describe('Parsing and export should not alter the description part', () => {
     const expectStrippedDescription = (description) => {
       const {
@@ -244,6 +230,7 @@ ${text}`;
         const exportedFile = parseAndExportOrgFile(testOrgFile);
         expect(exportedFile).toEqual(testOrgFile);
       });
+
       test('Parses all valid URLs starting with www', () => {
         const parsedFile = parseOrg(testOrgFile);
         const firstHeader = parsedFile.get('headers').first();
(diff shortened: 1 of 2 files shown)


schoettl:
Right, `parse_org.unit.test.js` is for tests of functions in `parse_org.js`. The other unit tests are higher-level to test parse + export.

Looks like (at least some of) these dups were introduced during refactoring by commit 7c966b23a29065bcbdf9ff3411805e342d2e69d6 

Thanks for the clean-up! It's hard to review for me so I hope you deleted tests very carefully :)

aspiers:
Yes, I was very careful - but this is no guarantee of anything ;-)

src/components/OrgFile/OrgFile.unit.test.js
```diff
@@ -66,20 +66,6 @@ describe('Tests for export', () => {
 });
 
 describe('Unit Tests for Org file', () => {
-  describe('Test the parser', () => {
```
munen, line 69: :+1:

src/lib/parse_org.unit.test.js
```diff
@@ -6,26 +6,8 @@ import {
   _parsePlanningItems,
   parseMarkupAndCookies,
 } from './parse_org';
-import { exportOrg } from './export_org';
 import readFixture from '../../test_helpers/index';
 
-/**
```
munen, line 12: :+1:

src/lib/parse_org.unit.test.js
```diff
@@ -65,124 +47,9 @@ describe('Test the parser', () => {
 });
 
 describe('Parsing and exporting should not alter the original file', () => {
-  test("Parsing and exporting shouldn't alter the original file", () => {
```
munen, line 68: @aspiers I agree that those tests are redundant in both files. However, you're deleting the parser tests from the "parser test" file. Mostly you're reverting one of my cleanup commits: https://github.com/200ok-ch/organice/commit/7c966b23a29065bcbdf9ff3411805e342d2e69d6

I'd opt to say that we leave 'parser related' tests in the parser unit test file. Things that are not parser related (for example reducer tests) should stay in OrgFile.unit.test.js.

Does that make sense for you?

aspiers: Oh dear, we should laugh at ourselves a little bit for the comedy of errors here :laughing: :stuck_out_tongue_winking_eye: 

[**@aspiers**](https://github.com/aspiers) commented on [May 29, 2020 5:14 PM](https://github.com/200ok-ch/organice/pull/307#issuecomment-636058795 "2020-05-29T16:14:08Z - Replied by Github Reply Comments"):
> [**@munen**](https://github.com/munen) commented on [May 29, 2020 3:43 PM](https://github.com/200ok-ch/organice/pull/307#issuecomment-636012238 "2020-05-29T14:43:13Z - Replied by Github Reply Comments"):
> > Good catch on the duplication of tests. Complete duplications can definitively go.
> > 
> > Feel free to pick and chose the correct location depending on the context of the test.
> 
> I don't really know enough about the tests yet to make the call on this :thinking:

So you gave me the freedom to choose, which I then declined ... and then we both went in opposite directions to our words :laughing: :laughing:  Sorry for the confusion!  Of course it's a trivial thing so hopefully no harm done at all :grin:  

> I'd opt to say that we leave 'parser related' tests in the parser unit test file. Things that are not parser related (for example reducer tests) should stay in OrgFile.unit.test.js.
>
> Does that make sense for you?

Yes it does, and in fact I tried to be very careful to place tests strictly according to this criteria.  This is how I concluded that tests which exercise not only `parseOrg()` but also `exportOrg()` go beyond the more narrow scope of `parse_org.unit.test.js`, and therefore are better placed in `OrgFile.unit.test.js`.  @schoettl [seemed to agree with me](https://github.com/200ok-ch/organice/pull/309#issuecomment-636367215 "2020-05-30T18:26:00Z - Replied by Github Reply Comments"):
> Right, `parse_org.unit.test.js` is for tests of functions in `parse_org.js`. The other unit tests are higher-level to test parse + export.

We could be even more pedantic about it and note that anything which tests both `parseOrg()` and `exportOrg()` is probably too complex to be a unit test and is more of a functional test, therefore wouldn't belong in `OrgFile.unit.test.js` either... But I'd prefer to avoid that kind of bike-shedding for now - I guess we probably all agree that there are more pressing problems to solve :wink: 

Of course it's your call if you think this is important enough to block merging.  However I need to strictly ration my time on Organice due to RSI symptoms, so it would take me longer to rework this, which would in turn delay #310.

aspiers: Sorry, that last paragraph might have sounded a little passive-aggressive, which I didn't want.  If merges get delayed a bit, that's fine too :-)  I know you're very busy anyway.

aspiers: Upon further thought, maybe it makes more sense to move all the `parseAndExportOrgFile()` tests to a new file.

schoettl: > Upon further thought, maybe it makes more sense to move all the `parseAndExportOrgFile()` tests to a new file.

Hm, I'm not so convinced of this idea because I'd say most tests in `parse_org.unit.test.js` are *meant* to test the parser. They just use the export function because this is often simpler than writing down the expected parse structure. But the focus is still on testing the more complex parts of the parser, not the export.

munen: I agree.

I'd vote to close this PR without merging. I'll do the cleanup in a separate PR. In any case, knowing that there's test duplication **and** having a list of all tests that are duplicated is already quite helpful, so your work is not in vain at all, @aspiers. Re-doing it is just a schlepptask, I'll happily do it.

Does that work for you, @aspiers?

munen: NB: I do see that your next PR #310 is based on this (which is perfectly fine and I consider good practice^^). It shouldn't be too hard to hard-reset the two commits away, though. Or will it cause more trouble?

aspiers: [**@munen**](https://github.com/munen) commented on [June 1, 2020 5:45 PM](https://github.com/200ok-ch/organice/pull/309#discussion_r433356940 "2020-06-01T16:45:20Z - Replied by Github Reply Comments"):
> I agree.
> 
> I'd vote to close this PR without merging. I'll do the cleanup in a separate PR. In any case, knowing that there's test duplication **and** having a list of all tests that are duplicated is already quite helpful, so your work is not in vain at all, [@aspiers](https://github.com/aspiers). Re-doing it is just a schlepptask, I'll happily do it.
> 
> Does that work for you, [@aspiers](https://github.com/aspiers)?

The problem is that since the duplication occurred, some of the tests got enhanced in one version but not the other.  I very carefully went through every single test, making a decision about which one was newer and therefore which should be kept.  So if you repeat this exercise yourself, you'll need to do the same, otherwise you'll end up accidentally regressing some previous enhancements, and in one case.  For example, `parseAndExportOrgFile()` has an extra `dontIndent` argument in `OrgFile.unit.test.js`.  So it's not as simple as using my PR as a list of tests to remove and then removing them from `OrgFile.unit.test.js`.

Personally I would be more inclined to merge this and then move the tests around as you see fit.  But it's your call.  

[**@munen**](https://github.com/munen) commented on [June 1, 2020 5:47 PM](https://github.com/200ok-ch/organice/pull/309#discussion_r433357958 "2020-06-01T16:47:10Z - Replied by Github Reply Comments"):
> NB: I do see that your next PR [#310](https://github.com/200ok-ch/organice/pull/310) is based on this (which is perfectly fine and I consider good practice^^). It shouldn't be too hard to hard-reset the two commits away, though. Or will it cause more trouble?

Done.  There will be a merge conflict with #310 now, due to the change in `exportOrg()`'s call signature, but that's not a big deal.

src/lib/parse_org.unit.test.js
```diff
@@ -63,124 +63,9 @@ describe('Test the parser', () => {
...
-          expect(parsedFile.strippedDescription).toEqual(testDescription);
-        });
-
-        test('Parsing a list with planning items should not mangle the list', () => {
-          const testDescription = '  - indented list\n     - Foo';
-          const parsedFile = _parsePlanningItems(`SCHEDULED: <2019-07-30 Tue>\n${testDescription}`);
-          expect(parsedFile.strippedDescription).toEqual(testDescription);
-        });
-
-        describe('Parses planning item with following checkmark', () => {
```
munen, line 169: Note to self marker: I'm up to here with the review. Where's this test now?

munen:
Ok, so we already have the first merge conflict on this PR. It's time to tackle it for good.

@aspiers You've put a lot of time into it! I just spent close to an hour just to sanity check the changes. It's very finicky work, because it would be sad to lose good tests. Having duplicate tests isn't good at all, either.

> Personally I would be more inclined to merge this and then move the tests around as you see fit. But it's your call.

I agree. I tried to move the tests right now, but it's 10pm and it's beyond me. I'm afraid if we leave this open too long, we'll have more merge conflicts. Hence, after this second checkup that all is good (and all was good, great work aspiers! :pray:), I'm merging this. Imo it's preferable to have tests in arguably slightly off namespaces than to have redundant tests.

Thanks for the extra effort on the code and the discussion here!

PR #310, opened 2020-05-30
Parse TODO keyword sets correctly from anywhere in the file (#151)

### LINKED ISSUE #151: #+TODO lines not parsed unless before first headline
I have a fairly complex configuration for my `TODO` keywords.  I have `org-todo-keywords` customised to:

```elisp
((sequence "NEXT(n)" "STARTED(>)" "|" "DONE(d)")
 (sequence "PROJECT(p)" "PROJDONE(P)")
 (sequence "ONGOING(o)" "WAITING(w@)" "CHASE(C@)" "|")
 (sequence "SOMEDAY(s)" "MAYBE(m)" "|" "CANCELLED(c)"))
```

Since organice does not know about elisp configuration, I added this to near the bottom of my `TODO.org`:

```org
* org-mode settings
*** TODO keywords
#+TODO: NEXT(n) STARTED(>) | DONE(d)
#+TODO: PROJECT(p) PROJDONE(P)
#+TODO: ONGOING(o) WAITING(w@) CHASE(C@) |
#+TODO: SOMEDAY(s) MAYBE(m) | CANCELLED(c)
```

However it seems to have no impact on organice.  All the above keywords are treated like normal headline text.  Upon further experimentation I isolated some issues which contribute to the problem:

1. The `(x)` keyboard shortcuts are not parsed separately, and are treated as part of the keywords. (**UPDATE: filed as #184**)

2. The `#+TODO` lines need to be at the top of the file.  If I put them under a headline anywhere, they cease to take effect.  This differs to Org's behaviour, and prevents us from being able to tuck them out of sight in order to keep the file tidy.

On the plus side, I did note that `|` seems to be handled correctly.

 - OS: openSUSE Linux
 - Browser: Chrome 79

 - Device: OnePlus 5T
 - OS: Android 9.0.x
 - Browser Brave

aspiers:
~~**N.B. This PR is currently stacked on top of #309, so a) please do not merge yet, and b) you only need to review the last commit of the three.  Once #309 is merged, I can rebase this on top of the merge, and it will become a single commit.**~~

Move parsing of TODO keyword sets (i.e. `#+TODO:` and `#+TYP_TODO:` lines) and other `#+...` config lines into a separate pass which happens before the main parsing loop.  This means that these lines can appear anywhere in the file and will still take effect, just like they do in emacs.  Also add support for `#+SEQ_TODO:` which means the same thing.

Since these lines can now appear anywhere, after collecting them into `todoKeywordSets`, it no longer makes sense to use `todoKeywordSets` to regenerate them in the export.  So in order for `exportOrg()` to preserve the exact original file, store all lines before the first header in `linesBeforeHeadings` in the order they appeared, without filtering out TODO keyword sets or other config lines, and use this in the export.  This has a nice side effect of significantly simplifying the `exportOrg()` function.

However, other config lines beginning with `#+...` are still collected in `fileConfigLines`, to allow other types of config to have an effect in organice as before.

Fixes #151.

### DIFF
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -110,8 +110,6 @@ const doSync = ({
               path,
               exportOrg({
                 headers: getState().org.present.get('headers'),
-                todoKeywordSets: getState().org.present.get('todoKeywordSets'),
-                fileConfigLines: getState().org.present.get('fileConfigLines'),
                 linesBeforeHeadings: getState().org.present.get('linesBeforeHeadings'),
                 dontIndent: getState().base.get('shouldNotIndentOnExport'),
               })
--- a/src/components/OrgFile/OrgFile.unit.test.js
+++ b/src/components/OrgFile/OrgFile.unit.test.js
@@ -19,8 +19,6 @@ function parseAndExportOrgFile(testOrgFile, dontIndent = false) {
   const parsedFile = parseOrg(testOrgFile);
   const exportedFile = exportOrg({
     headers: parsedFile.get('headers'),
-    todoKeywordSets: parsedFile.get('todoKeywordSets'),
-    fileConfigLines: parsedFile.get('fileConfigLines'),
     linesBeforeHeadings: parsedFile.get('linesBeforeHeadings'),
     dontIndent: dontIndent,
   });
@@ -175,8 +173,7 @@ ${text}`;
     test('Parse file with one empty line', () => {
       const testOrgFile = '\n';
       const exportedFile = parseAndExportOrgFile(testOrgFile);
-      // if the exporter produces no output at all, the single newline character is discarded
-      expect(exportedFile).toEqual('');
+      expect(exportedFile).toEqual('\n');
     });
 
     test('Parse very basic file with description', () => {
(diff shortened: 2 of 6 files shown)


munen:
I've looked over commit 3 and tested it with my three biggest Org files (each in the hundreds of KBs). In my tests the feature works as advertised while I didn't see any regressions.

I quite like it; it's a huge addition with a net removal of LOC and complexity.

I'll wait on merging util #309 is resolved.

@schoettl If you want to get involved before merging, just give a ping and I'll wait until you had time to look over it.

From me, so far: A+. Less complexitiy, less code, more tests, more functionality. Best case PR, @aspiers! :bow: 

NB: I also checked on the performance, but that was totally m00t, as it's taking up so little time:

image

aspiers:
Rebased, and even more code deleted since `exportOrg()` doesn't need to know about `todoKeywordSets` and `fileConfigLines` any more.

schoettl:
I'll look through it now :)

munen:
I tested the functionality, again - and it still works like a charm(;

I've [amended the docs](02654ad) and [added the attribution](02654ad).

Great work, @aspiers :pray:

PR #314, opened 2020-06-01
Explain what an SPA is

aspiers:
It wasn't obvious to me, even though I've encountered that acronym before (and then forgot it).  So add "Single Page Application" just before the first mention.

This change only modifies the first half of the first sentence.  The rest is just refilling the paragraph.


munen:
Sweet, thanks!

It's a good point that this might be unclear. I'll pull in your change and add a link to the MDN page, too: https://developer.mozilla.org/en-US/docs/Glossary/SPA

PR #316, opened 2020-06-01
Use a more robust approach to enabling CORS pre-flight checks

aspiers:
Using the `R=200` trick in mod_rewrite seems problematic, because it can cause Apache to return a `200 OK` response with a very unhelpful message body including:

    The server encountered an internal error or misconfiguration and
    was unable to complete your request.

This SO answer suggests that the internal error is due to Apache not having an `ErrorDocument` configured for `200` responses by default:

- https://stackoverflow.com/a/48529566/179332

Apparently a `204` code (`No Content`) is a more appropriate response:

- https://stackoverflow.com/questions/27703871/return-empty-response-from-apache/

and indeed my experiments confirm that changing the `RewriteRule` to use `R=204` eliminates this error.

However, this approach doesn't guarantee that the CORS headers are included in the response.  A better approach is to revisit the original motivation for this rewrite rule, which was presumably to get around the fact that the browser doesn't send credentials with the CORS preflight check, therefore the server would typically give a `401` response.  So we can simply follow an alternative approach:

- https://serverfault.com/a/431718/79729

where we do not require authorization for `OPTIONS` requests.

### DIFF
--- a/README.org
+++ b/README.org
@@ -485,6 +485,9 @@ In the [[https://organice.200ok.ch/documentation.html#faq_webdav][FAQ]], you'll 
   - Documentation on Nextcloud sharing
 
 ** Routing
+   :PROPERTIES:
+   :CUSTOM_ID: routing
+   :END:
 
 Whilst organice is a true [[https://developer.mozilla.org/en-US/docs/Glossary/SPA][Single Page Application]] (SPA) and therefore
 has no back-end whatsoever, this does have an implication for
@@ -514,7 +517,29 @@ RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI} -f [OR]
 RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI} -d
 RewriteRule ^ - [L]
 
+# N.B. If this next rule is allowed to apply to a WebDAV directory, it
+# will break PUT requests to upload new files!
 RewriteRule ^ /index.html [L]
+#+END_EXAMPLE
+
+If your WebDAV directory happens to be not only on the same webserver,
+but also within a subdirectory of the directory containing this
+=.htaccess= file, then you will need to create another =.htaccess=
+file in the top-level WebDAV directory containing this:
+
+#+BEGIN_EXAMPLE
+RewriteEngine Off
+#+END_EXAMPLE
+
+Otherwise any attempts to use WebDAV to upload new files via =HTTP
+PUT= requests will fall foul of the =/index.html= rewrite rule above,
+resulting in a =403 Forbidden= response.
+
+Another way to avoid this more selectively is to precede that rule
+with:
+
+#+BEGIN_EXAMPLE
+RewriteCond %{REQUEST_METHOD} !PUT
 #+END_EXAMPLE
 
 * Capture templates
--- a/README.org
+++ b/README.org
@@ -474,7 +474,7 @@ like [[https://httpd.apache.org/docs/2.4/mod/mod_dav.html][Apache]] or [[https:/
 
 **** More information
 
-In the [[https://github.com/200ok-ch/organice/wiki#webdav][Wiki]], you'll find lots more information regarding WebDAV:
+In the [[https://organice.200ok.ch/documentation.html#faq_webdav][FAQ]], you'll find lots more information regarding WebDAV:
 
   - A screencast of how organice works when logging in to a WebDAV
     server
--- a/README.org
+++ b/README.org
@@ -517,30 +517,13 @@ RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI} -f [OR]
 RewriteCond %{DOCUMENT_ROOT}%{REQUEST_URI} -d
 RewriteRule ^ - [L]
 
-# N.B. If this next rule is allowed to apply to a WebDAV directory, it
-# will break PUT requests to upload new files!
 RewriteRule ^ /index.html [L]
 #+END_EXAMPLE
 
-If your WebDAV directory happens to be not only on the same webserver,
-but also within a subdirectory of the directory containing this
-=.htaccess= file, then you will need to create another =.htaccess=
-file in the top-level WebDAV directory containing this:
-
-#+BEGIN_EXAMPLE
-RewriteEngine Off
-#+END_EXAMPLE
-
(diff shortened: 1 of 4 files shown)


WIKI.org
```diff
@@ -137,7 +135,8 @@ to your Dropbox, here are your options:
  have HTTP ok status.
  #+END_EXAMPLE
 
- then the =RewriteRule= in the above example is either missing from, or ineffective in your webserver config.
+ then the =RewriteRule= in the above example is either missing from,
```
schoettl, line 138: which `RewriteRule` in the above example? You deleted it.

aspiers: Ooops haha good spot! I went through several iterations of the config before settling on `<LimitExcept>`, and forgot to update this.

README.org
```diff
@@ -471,7 +471,7 @@ about that - here's a [[https://github.com/200ok-ch/organice/wiki#missing-cors-h
 
 **** More information
 
-In the [[https://github.com/200ok-ch/organice/wiki#webdav][Wiki]], you'll find lots more information regarding WebDAV:
+In the [[https://organice.200ok.ch/documentation.html#faq_webdav][FAQ]], you'll find lots more information regarding WebDAV:
```
munen, line 474: :+1:

WIKI.org
```diff
@@ -54,11 +54,13 @@ to your Dropbox, here are your options:
 
 *** Missing CORS headers
 
- If your back-end doesn't have the necessary CORS headers set (more
- info in the [[https://github.com/200ok-ch/organice/blob/master/README.org][README]], this is what the error will look like:
+ If your back-end doesn't have the necessary CORS headers set, this is
```
munen, line 57: :+1:

WIKI.org
aspiers: Fixed.

WIKI.org
```diff
@@ -78,17 +80,15 @@ to your Dropbox, here are your options:
...
- RewriteRule ^(.*)$ $1 [R=200,L]
-
  <Location /webdav>
      Options Indexes
      DAV On
      AuthType Basic
      AuthName "webdav"
      AuthUserFile /srv/dav/.htpasswd
-     Require valid-user
+     <LimitExcept OPTIONS>
```
munen, line 89: It's great that you're putting a lot of effort into better docs and setup for WebDAV. It'll certainly be very much appreciated by future readers!

At this time, I cannot test this new configuration for Apache, though. When I wrote the original docs, I did set this up locally, so that I could test WebDAV. Since then, I have switched computers and haven't set this up, again.

I'm hesitant to merge config that I haven't checked. By this, I don't want to imply that it's not working for you, but as a maintainer I should be able to understand/reproduce the setup.

The best way to move forward would likely be to create this Apache setup in a `Dockerfile`. I've been thinking of doing that for some time now. It would ease a lot of discussions on the topic of WebDAV. Right now, everyone is using some very specific setup (NextCloud, HA Proxy, Apache, etc) and it's not really possible to give support for that. Such Ops topics are way out of scope of a SPA project. If we had an official example Docker container that works for us and we use in development, we can always point to that one.

What do you think? Maybe you're even already using a Docker based setup?(;

aspiers: I'm not using Docker but this is an excellent idea.

Totally understand your hesitation.  I just spent a frustrating 3-4 hours trying to really get to the bottom of this Apache voodoo.  I conducted countless experiments, and I'm now pretty confident `LimitExcept` is the better way to go.  It has clear logic to it (prevent a CORS preflight `OPTIONS` check from resulting in 401 response due to lack of credentials) whereas the "use `RewriteRule ^(.*)$ $1 [R=200,L]` to rewrite a URL to itself" trick is much weirder and hard to explain.  In fact, can you explain that?  If so, I'd love to hear the explanation so I can fully understand it.  If not then you have a choice between two configs you don't fully understand ;-)

BTW it makes no sense to me to have the rewrite rules applying more generally, outside the DAV `<Directory>` or `<Location>` section, because we only want CORS checks to succeed for DAV.

aspiers:
I've just pushed another commit for yet another WebDAV gotcha.  I really hope this is the last Apache / WebDAV pitfall I have to haul myself out of ... :man_facepalming:

WIKI.org
munen: @aspiers Very well, then. I've just spend some time and have setup such a Docker based setup and rewrote the documentation accordingly: https://github.com/200ok-ch/organice/pull/356

If you generally approve of the PR, I'll merge it and then I can add your proposed changes to the official test bench and see if it still works on my machine(;

NB: If you can get the authentication to work using the Docker image, I don't mind if you enable it and propose a change. I tried quite a few times, but failed. Since it's just a test bench, it can live without authentication.

aspiers: #356 looks good to me, please merge and then we can resume progress on this PR.  I'd be interested to hear what blockers you hit regarding authentication.  I wonder if this PR helps with any of them.

aspiers:
I have just rebased this on #356 but haven't pushed yet.

munen:
@aspiers I'll wait for you to push your rebased changes to this branch. Then, I'll test them locally. If logging in and making changes to files and config still works, I'll happily pull them in!

Thanks again, for the extra effort of well documented and robust WebDAV setup for organice. That'll prove very valuable for many users, I'm sure! :pray:

aspiers:
Now rebased.

WIKI.org
```diff
@@ -142,7 +142,31 @@ about that - here's a screencast showing you how to do it:
  have HTTP ok status.
  #+END_EXAMPLE
 
- then the =RewriteRule= in the above example is either missing from, or ineffective in your webserver config.
+ then something is wrong with your webserver config.  You can check
+ whether a CORS preflight check is returning the right headers via:
+
+ #+BEGIN_EXAMPLE
+ curl -v -X OPTIONS https://my.server/webdav
```
munen, line 149: NB: For the official Apache test bench, we need to have a trailing slash, otherwise there's no 200 ok, but a 301 permanently moved.

munen:
The proposed changes to the webdav config in Apache work for me!

I've slighly re-arranged your docs so that everything related to setting up WebDAV is grouped together, but still linked from all relevant locations in the docs.

munen:
Great stuff, the webdav config actually got slimmer!(;

Hence, good finds^^

PR #318, opened 2020-06-02
Fix rendering of file: links  (#311)

### LINKED ISSUE #311: Opening `file://` links only works in one easy case and breaks organice in others
**Describe the bug**

I have to apologize. I have merged PR #307 too quickly without proper testing. Initially, I thought that your solution to 'just' use the URL for opening a file is quite ingenius. Personally, I always envisioned this feature to be more complex. It turns out, the current setup is a little too easy and works only in one simple scenario: Opening files relative to the current file which is also accessible within the sync-backend. It doesn't work for:

1. Opening files with absolute paths (no matter whether the file is in the sync-backend or not)
2. Opening files which are not available in the sync-backend
3. On iOS, when in the PWA full-screen view, clicking any link will be treated as opening an external link which results in leaving the fullscreen PWA and opening the browser. It might well be that the user isn't logged in to Mobile Safari (just the PWA), and then the user would only see the login screen. Even if not, any open link would result in new tab and leaving the PWA.

Especially 1 and 2 leave the user in a state from which organice does not recover. The only solution is a reload which is hard to do on a PWA.   

**To Reproduce**

- Bug 1/2 (on the Desktop): https://www.dropbox.com/s/yaifnyioazlp548/2020-05-31%2011-08-05.mkv?dl=0
- Bug 3 (on iOS): https://www.dropbox.com/s/i4rf8kouae2xvz9/Video%2031.05.20%2C%2011%2022%2047.mp4?dl=0

**Expected behavior**

- File links that can be reached should be reached
- File links that cannot be reached are OK to show a 'file not found', but they shouldn't break organice
- Clicking a link shouldn't throw the user off the PWA

 - OS: iOS and Debian
 - Browser: Mobile Safari and FIrefox

aspiers:
The new logic for rendering `file:` links introduced by #307 had a number of issues:

- It created links for absolute paths, which does not make sense since there is no fully general mapping between the filesystem and any of the sync backends.

- It created links for relative paths pointing outside the backend (e.g. `../../../../foo.org`),

- It used `<a href="...">` rather than `<Link to=...>`, which meant the PWA treated them as external links.

So create a `<Link>` for relative `file:` paths pointing to `.org` files, don't link other `file:` paths, and render all other links with `<a ...>` as before.

Fixes #311.

### DIFF
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -8,6 +8,8 @@ import TablePart from './components/TablePart';
 import ListPart from './components/ListPart';
 import TimestampPart from './components/TimestampPart';
 
+import { orgFileExtensions } from '../../../../lib/org_utils';
+
 import classNames from 'classnames';
 
 export default ({ parts, subPartDataAndHandlers }) => {
@@ -19,13 +21,13 @@ export default ({ parts, subPartDataAndHandlers }) => {
     const id = part.get('id');
     const uri = part.getIn(['contents', 'uri']);
     const title = part.getIn(['contents', 'title']) || uri;
-    if (uri.startsWith("file:")) {
+    if (uri.startsWith('file:')) {
       const target = uri.substr(5);
       const isRelativeOrgFileLink =
-        !target.startsWith("/") &&
-        !target.startsWith("~") &&
-        !target.startsWith("../") &&
-        uri.endsWith(".org");
+        !target.startsWith('/') &&
+        !target.startsWith('~') &&
+        !target.startsWith('../') &&
+        uri.match(orgFileExtensions);
       if (isRelativeOrgFileLink) {
         const dir = location.pathname.match(/.*\//);
         return (
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -21,31 +21,49 @@ export default ({ parts, subPartDataAndHandlers }) => {
     const id = part.get('id');
     const uri = part.getIn(['contents', 'uri']);
     const title = part.getIn(['contents', 'title']) || uri;
+    let target = uri;
     if (uri.startsWith('file:')) {
-      const target = uri.substr(5);
+      target = uri.substr(5);
       const isRelativeOrgFileLink =
         !target.startsWith('/') &&
         !target.startsWith('~') &&
-        !target.startsWith('../') &&
         uri.match(orgFileExtensions);
       if (isRelativeOrgFileLink) {
-        const dir = location.pathname.match(/.*\//);
-        return (
-          <Link key={id} to={dir + target}>
-            {title}
-          </Link>
-        );
-      } else {
-        return title;
+        target = normalisePath(target);
+        if (!target.includes('/../')) {
+          // Normalisation succeeded, so we can safely return a <Link>
+          return (
+              <Link key={id} to={target}>{title}</Link>
+          );
+        }
       }
-    } else {
-      return (
-        <a key={id} href={uri} target="_blank" rel="noopener noreferrer">
-          {title}
-        </a>
-      );
+
+      // Best effort in other cases; the file:// href may or may not work,
+      // but either way it should help show the user what's going on.
+      target = 'file://' + target;
     }
+
+    return (
+        <a key={id} href={target} target="_blank" rel="noopener noreferrer">
+        {title}
+      </a>
+    );
   };
+
+  const normalisePath = target => {
+    let dir = location.pathname.match(/(.*)\//)[1];
+    let normalised = target;
+    while (normalised.startsWith("../")) {
+      if (!dir.match(/^\/file\/.+/)) {
+        // We're already at the top; can't break out of the area accessible
+        // via HTTP, so just the original non-normalised path
+        return target;
+      }
+      normalised = normalised.substr(3);
+      dir = dir.match(/(.*)\//)[1];
+    }
+    return dir + '/' + normalised;
+  }
 
   return (
     <span>
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -24,14 +24,22 @@ export default ({ parts, subPartDataAndHandlers }) => {
     let target = uri;
     if (uri.startsWith('file:')) {
       target = uri.substr(5);
-      const isRelativeOrgFileLink =
+      const isRelativeFileLink =
         !target.startsWith('/') &&
-        !target.startsWith('~') &&
-        uri.match(orgFileExtensions);
-      if (isRelativeOrgFileLink) {
+        !target.startsWith('~');
+      if (isRelativeFileLink) {
+        // N.B. Later on we may improve this conditional by performing
+        // an existence check on the backend if it allows that
+        // operation.
+
         target = normalisePath(target);
         if (!target.includes('/../')) {
           // Normalisation succeeded, so we can safely return a <Link>
+          if (!uri.match(orgFileExtensions)) {
+            // Optimistically assume that the link is pointing to a
+            // directory.
+            target = target.replace(/^\/file\//, '/files/');
+          }
           return (
               <Link key={id} to={target}>{title}</Link>
           );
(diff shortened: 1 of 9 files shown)


src/components/OrgFile/components/AttributedString/index.js
```diff
@@ -11,26 +13,46 @@ import classNames from 'classnames';
...
+      const target = uri.substr(5);
+      const isRelativeOrgFileLink =
+        !target.startsWith("/") &&
+        !target.startsWith("~") &&
+        !target.startsWith("../") &&
+        uri.endsWith(".org");
+      if (isRelativeOrgFileLink) {
+        const dir = location.pathname.match(/.*\//);
+        return (
+          <Link key={id} to={dir + target}>
```
schoettl, line 32: Do we need the `dir`? Relative (even `../`) would work without `dir` – at least in the `<a ...>`.

src/components/OrgFile/components/AttributedString/index.js
```diff
@@ -11,26 +13,46 @@ import classNames from 'classnames';
...
+  const renderLink = (part) => {
+    const id = part.get('id');
+    const uri = part.getIn(['contents', 'uri']);
+    const title = part.getIn(['contents', 'title']) || uri;
+    if (uri.startsWith("file:")) {
+      const target = uri.substr(5);
+      const isRelativeOrgFileLink =
+        !target.startsWith("/") &&
+        !target.startsWith("~") &&
+        !target.startsWith("../") &&
```
schoettl, line 27: Actually, this is also a relative link. But OK, it's not trivial to fix it against "sync directory break-out".

src/components/OrgFile/index.js
```diff
@@ -114,11 +114,16 @@ class OrgFile extends PureComponent {
...
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
     const { headers, pendingCapture } = this.props;
     if (!!pendingCapture && !!headers && headers.size > 0) {
       this.props.org.insertPendingCapture();
     }
+
+    const { path } = this.props;
```
schoettl, line 123: If we don't really need `path`/`dir`, we could revert this hunk.

test_helpers/fixtures/main_test_file.org
```diff
@@ -12,9 +12,12 @@ Some description content
 
 * A header with tags                                              :tag1:tag2:
 * A header with [[https://organice.200ok.ch][a link]]
-* A header with a link to a local .org file as content
-
-  [[file:schedule_and_timestamps.org][a local .org file]]
+* A header with various links as content
+  Relative link to [[file:schedule_and_timestamps.org][an existing .org file in the same directory]]
+  Relative link to [[file:subdir/foo.org][a fictitious .org file in a sub-directory]]
```
schoettl, line 17: For completeness, add relative `../` link?

src/components/OrgFile/components/AttributedString/index.js
```diff
@@ -11,26 +13,46 @@ import classNames from 'classnames';
...
+        uri.endsWith(".org");
+      if (isRelativeOrgFileLink) {
+        const dir = location.pathname.match(/.*\//);
+        return (
+          <Link key={id} to={dir + target}>
+            {title}
+          </Link>
+        );
+      } else {
+        return title;
```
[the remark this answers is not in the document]
aspiers: Sure, good point - will do!

src/components/OrgFile/index.js
aspiers: No we can't.  This is essential for making `<Link>` work in this context, otherwise the address bar changes but the displayed file doesn't.

test_helpers/fixtures/main_test_file.org
aspiers: That was supposed to be on the line immediate below this comment, but Orgmode auto-corrected it to my local absolute path without me noticing :-(  Will fix.

munen:
Looks great to me. I've tested it and it fixes the discussed issues. I've added support for `.org_archive` file suffixes (53d8ca5eb0137683c83e930855b604198f2e690b) and have updated the changelog with attribution (26f5af4f7b0b767aa62de4e0d4066929a653f11f).

I'm waiting for https://github.com/200ok-ch/organice/pull/318/files#r434764160 to be resolved.

@aspiers No need to rebase on your last commit. This way I can easily just review the last commit.

Again, great work! :pray: Very much looking forward to merging this PR!

src/components/OrgFile/components/AttributedString/index.js
aspiers: Not trivial, but not too hard either ;-)  I've done it.

src/lib/org_utils.js
```diff
@@ -657,3 +657,6 @@ export const getTodoKeywordSetsAsFlattenedArray = (state) => {
     .toSet()
     .toJS();
 };
+
+/** Regular expression of file extensions to validate a filename. */
+export const orgFileExtensions = /org$|org_archive$/;
```
schoettl, line 662: I'm not fully convinced of enforcing a filename extension for org files. But anyway, if we do, it should be an extension starting with `.`.

schoettl: For example, my emacs or whatever often creates files like `#my.org#` or `my.org~`. Or maybe a beginner saves a file as `.txt`. Is it a problem to just show the "cannot parse" error message for non-org files?

src/sync_backend_clients/dropbox_sync_backend_client.js
```diff
@@ -15,7 +16,7 @@ export const filterAndSortDirectoryListing = (listing) => {
     // Show all folders
     if (file['.tag'] === 'folder') return true;
     // Filter out all non-org files
-    return file.name.match(/org$|org_archive$/);
+    return file.name.match(orgFileExtensions);
```
schoettl, line 19: OK, I understand – we already have this file filter...

src/lib/org_utils.js
aspiers: Good point; fixed in the latest commit.

munen: > Why should we prohibit to open these files?

Your examples sound a lot like tmp files for unsaved buffers. I’d say it’s prudent not to open these, because any changes to it are potentially just going to the GC^^

My initial reason to limit for those two extensions was not links, though. I’m filtering all files in the file explorer to be of that type or of a folder. The rationale here is that I personally have about two zillion of files - and when using organice, I don’t care for any of them except for the ones that I can actually work on. 

There’s certainly other options for that:

- more powerful file explorer (search/sort/filter)
- make the file extensions configurable

schoettl: OK, makes sense.

src/components/OrgFile/components/AttributedString/index.js
```diff
@@ -1,36 +1,86 @@
...
+  let location = useLocation();
+
+  const renderLink = (part) => {
+    const id = part.get('id');
+    const uri = part.getIn(['contents', 'uri']);
+    const title = part.getIn(['contents', 'title']) || uri;
+    let target = uri;
+    if (uri.startsWith('file:')) {
+      target = uri.substr(5);
+      const isRelativeOrgFileLink =
```
schoettl, line 27: Rename this variable to `isRelativeLink` because it can also be a directory?

aspiers: Good catch! Fixed.

src/sync_backend_clients/dropbox_sync_backend_client.js
aspiers: Yes, and I've now fixed it to include the `.` :-)

munen:
@aspiers I like the changes. It's taking care of the issues we've discussed before.

It does add another kind of broken link, though. The optimistic check for directories yields funny things - for example when trying to open a `org.gpg` file.

But I've seen your commit comment that you're aware of that. I'd agree that a check is necessary.

In any case, the changes are certainly a big step into the right direction, I'm happily merging! :pray:

test_helpers/fixtures/main_test_file.org
munen: That's interesting behavior. I have had the same in my tests!

munen:
I have updated the changelog with the current state and attribution accordingly: 9d88a2c

PR #320, opened 2020-06-02
Add a new option to toggle whether .organice-bak files are pushed

aspiers:
Not all users want automatic backup files.  For example, a WebDAV backend which exposes a git repository with external processes handling automatic commit/sync has no need for backups, because there is already full git history.

So add an option to disable this feature.

### DIFF
--- a/src/util/settings_persister.js
+++ b/src/util/settings_persister.js
@@ -107,6 +107,13 @@ export const persistableFields = [
   },
   {
     category: 'base',
+    name: 'shouldBackup',
+    type: 'boolean',
+    shouldStoreInConfig: true,
+    default: true,
+  },
+  {
+    category: 'base',
     name: 'shouldShowTitleInOrgFile',
     type: 'boolean',
     shouldStoreInConfig: true,
@@ -209,7 +216,11 @@ export const readInitialState = () => {
         value = null;
       }
     } else if (field.type === 'boolean') {
-      value = value === 'true';
+      if (!value && field.default !== null) {
+        value = field.default;
+      } else {
+        value = value === 'true';
+      }
     } else if (field.type === 'json') {
       if (!value) {
         value = field.default || Map();
--- a/sample.org
+++ b/sample.org
@@ -297,7 +297,8 @@ file back up.
 
 Generally, when working with distributed Org files, we're recommending
 to put them under version control and to check for bugs and racing
-conditions between clients.
+conditions between clients. There is a setting available to disable
+backups in case you do not need them.
 * organice operates completely client side
 You don't log in to organice directly because organice doesn't have a
 back end - it operates completely client side using Dropbox, Google
--- a/sample.org
+++ b/sample.org
@@ -288,9 +288,9 @@ WebDAV. Click the "Sign in" button in the upper right hand corner to
 sign in with either of them and authenticate organice.
 
 ** Backups
-The first time you push changes from organice back up to your chosen
-sync service, organice will make a backup of the original file first.
-It'll be named {your-file-name}.organice-bak. Dropbox and Google Drive
+Whenever organice loads a file from your chosen sync service, before
+you can make any changes it will first make and push a backup of the
+file, named ={your-file-name}.organice-bak=. Dropbox and Google Drive
 also both keep a full version history of your files for you, but this
 is an additional precaution in case something goes wrong pushing the
 file back up.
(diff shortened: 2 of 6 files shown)


src/util/settings_persister.js
```diff
@@ -105,6 +105,13 @@ export const persistableFields = [
     type: 'boolean',
     shouldStoreInConfig: true,
   },
+  {
+    category: 'base',
+    name: 'shouldBackup',
+    type: 'boolean',
+    shouldStoreInConfig: true,
+    default: true,
```
munen, line 113: The PR makes a lot of sense, the implementation works and is exactly modeled after the existing flags.

Unfortunately (like in the literal sense), I cannot merge this PR, yet. The reason is that `default: true` doesn't work. There's no good `default` functionality at this time for settings.

It's possible to set a `default` via `||` in `mapStateToProps` in Settings/index.js, however we can't set the `default` to `|| true`, because that will always evaluate to `true`.

For a new flag/feature, I wouldn't mind merging without a default. If users don't enable it, it's not an issue. However, this flag would by default disable a feature that has been enabled. So, if a user is expeting for the feature to run, it would silently fail her.

Not having a default for Settings has been on my private list for a while. I guess now is the time to think about it. If you've got any ideas on how to implement it, you're welcome to beat me to the punch^^

aspiers: > It's possible to set a `default` via `||` in `mapStateToProps` in Settings/index.js, however we can't set the `default` to `|| true`, because that will always evaluate to `true`.

I'm pretty sure I already addressed that.  Please see https://github.com/200ok-ch/organice/pull/320/files#diff-1260fc7748c44c832b55e6f3369f6fabR219

I'm guessing you are thinking of the existing code which handles defaults for `json` values; however as you can see below, my new code for `boolean` defaults works correctly - or at least it did when I tested it ;-) 

> For a new flag/feature, I wouldn't mind merging without a default. If users don't enable it, it's not an issue. However, this flag would by default disable a feature that has been enabled. So, if a user is expeting for the feature to run, it would silently fail her.

Again, I'm pretty confident that it keeps the existing default behaviour, but ICBW of course.

> Not having a default for Settings has been on my private list for a while. I guess now is the time to think about it. If you've got any ideas on how to implement it, you're welcome to beat me to the punch^^

Well I think that with this PR it now should work for `boolean` and `json` types.  I'm not sure whether it makes sense for `nullable` types, since if the default isn't null, then can the user even nullify it later?

src/util/settings_persister.js
```diff
@@ -209,7 +216,11 @@ export const readInitialState = () => {
         value = null;
       }
     } else if (field.type === 'boolean') {
-      value = value === 'true';
+      if (!value && field.default !== null) {
```
aspiers, line 219: This is the key line.  If the default is `true` or `false`, the value will be set to that.

munen:
@aspiers Thank you for actually already thinking building this PR in a non-breaking way; that's much appreciated!

I actually did test - even before I looked over the code (and could have seen your change). For me, it doesn't work. Here's a screencast showing that I'm on your branch, I pulled the latest version, I restart `yarn start` and when the flag is not set in the JSON, the toggle is `off`: 

shouldBackup_test

munen:
When the initial state is read in `App.js` and the flag is not available in the config JSON, then the flag is set to `false`:

image

munen:
The reason is that `default` isn't set: 
image

munen:
I started to debug. In `settings_persister.js`, when doing `readInitialState()`, in the beginning it is set correctly:

image

munen:
It's a little weird.

When I clear my localStorage and do an initial load of the application, and then debug `readInitialState`, then the value is already set to "true".

When I login to Dropbox, then, the value in localStorage will be "undefined".

This is reproducible. I tried twice.

There's some code writing back to localStorage very early on and messing something up.

munen:
I'm giving up the hunt for today. It shouldn't be too hard to find the spot. There's not many LOC containing `localStorage.setItem()` and none containing `localStorage[]`. However, I don't find the spot where it breaks down.

On a very high level, this is what happens:

- On an empty localStorage, load organice the first time, the value is set to "true"
- Click on Sign In with Dropbox, the value is set to "true"
- After having logged in with Dropbox, the value is set to "undefined"

aspiers:
Many thanks for all your efforts so far!  I will have another look at this.

munen:
Adding a `blocked` label to show that we want to resolve the default settings issue, first.

Issue #321, opened 2020-06-03
Performance bottlenecks in headerWithPath / parentIdOfHeaderWithId


aspiers:
[As observed on Matrix](https://matrix.to/#/!DfVpGxoYxpbfAhuimY:matrix.org/$0yve1zLCmfGR1moQABe3vcEPPUpnH3UWwjhqW2_1ong?via=matrix.org), searching a large org file for a header within a capture can be extremely slow.  Chrome dev tool's Performance tab clearly illustrates that `headerWithPath` has an issue:

![image](https://user-images.githubusercontent.com/100738/83697378-becf0e80-a5f6-11ea-9cf0-fa2669d6494d.png)

Drilling down further reveals that the cost of `parentIdOfHeaderWithId()` calling `.reverse()` is certainly an issue, but there may be others.

## To Reproduce

1. Set up a large `.org` file
2. Set up a capture template where the header doesn't match anything
3. Initiate a capture

## Expected behavior

It displays the `The specified header path doesn't exist in this org file!` message quickly.

**Desktop (please complete the following information):**
 - OS: openSUSE Tumbleweed
 - Browser: Chrome 80

aspiers:
OK it seems another problem is that this code is `O(n^2)`:

https://github.com/200ok-ch/organice/blob/68f688dca719cc11285b4e6fcf355aa5bc746468/src/lib/org_utils.js#L191-L195

because `parentIdOfHeaderWithId()` is itself `O(n)`.

Furthermore,

https://github.com/200ok-ch/organice/blob/68f688dca719cc11285b4e6fcf355aa5bc746468/src/lib/org_utils.js#L81-L82

is twice as slow as it needs to be, because `headerWithId()` calls `indexOfHeaderWithId()`, and then throws away the result after using it to return the header.

aspiers:
#322 and #323 have already made a significant difference here, however we could still improve by an order of magnitude by maintaining a mapping from `headerId` to `headerIndex`, which would change `indexOfHeaderWithId()` from O(n) to O(1).

However it's getting late so I should probably stop here.

aspiers:
Worth noting that @munen [does not see the same performance issue on his iPhone Xs](https://matrix.to/#/!DfVpGxoYxpbfAhuimY:matrix.org/$t0C1TSsHXJTYINwLzQEOO13X0SjsabDkH0QrX7vdAUw?via=matrix.org), which I find utterly baffling.

munen:
[...]

PR #322, opened 2020-06-03
Use findLast() instead of reverse() and find()

aspiers:
Doing a `reverse()` just to `find()` the last header matching some predicate is needlessly expensive; so just use `findLast()` instead:

https://immutable-js.github.io/immutable-js/docs/#/List/findLast

Partially addresses #321.


munen:
Definitively a solid improvement!

PR #323, opened 2020-06-04
Double performance of obtaining header and index for given header id

aspiers:
Many places in the code had adopted the following pattern:

    const header = headerWithId(headers, headerId);
    const headerIndex = indexOfHeaderWithId(headers, headerId);

However this is twice as slow as necessary, because `headerWithId()` calls `indexOfHeaderWithId()`, and then throws away the result after using it to return the header.

So instead build a single `indexAndHeaderWithId()` function which obtains the index, does an O(1) lookup to obtain the header, then returns an object containing both.

Partially addresses #321.

### DIFF
--- a/src/lib/org_utils.js
+++ b/src/lib/org_utils.js
@@ -31,6 +31,11 @@ export const changelogHash = () => {
   });
 };
 
+export const indexAndHeaderWithId = (headers, headerId) => {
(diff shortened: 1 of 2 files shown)


munen:
Good catch and nice improvement! :pray:

PR #343, opened 2020-06-08
doc: Use `CUSTOM_ID` permalinks everywhere

### LINKED ISSUE #342: Docs should use CUSTOM_ID permalinks everywhere
Currently the docs use a mixture of `<<radio targets>>` and `CUSTOM_ID` properties to allow deep-linking to particular sections.  The latter are much better because:

1. they provide a stable permalink which will not break when the docs are regenerated, e.g. compare https://organice.200ok.ch/documentation.html#deployment vs. https://organice.200ok.ch/documentation.html#orgc7ecd7e which is randomly generated each time
2. the permalink is human-readable
3. GitHub's rendering of the `.org` file results in radio targets polluting the text, whereas `CUSTOM_ID` anchors are nicely hidden.

So I propose switching all radio targets to `CUSTOM_ID` properties.

munen:


aspiers:
Wow that was quick, awesome thanks! :tada:

munen:
@aspiers 

> Wow that was quick, awesome thanks! tada

Sure, after you mentioned that feature the first time (I didn't know about it before), I had it on my private todo list, but was lazy about it. The change itself took only a minute thanks to Emacs, Macros and Regexp replacement(; Sometimes, a public issue is a good motivator*lol*

PR #352, opened 2020-06-12
Get back on track with prettier

aspiers:
Fix prettier regressions, and run prettier and prettier-eslint checks in CircleCI.  Also add a `yarn lint` shortcut for easy local testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,13 +46,29 @@ jobs:
           name: Install Dependencies
           command: yarn install --frozen-lockfile
 
+      - run:
+          name: Show Dependencies
+          command: yarn list
+
       - save_cache:
           name: Save Yarn Package Cache
           key: yarn-packages-{{ checksum "yarn.lock" }}
           paths:
             - ~/.cache/yarn
 
-      - run: yarn eslint
+      - run:
+          name: eslint
+          command: yarn eslint
+
+      - run:
+          name: prettier-eslint
+          # Make the build output more useful by immediately showing full details of
+          # any style violations.
+          command: |
+            yarn prettier-eslint --write
+            GIT_PAGER= git diff
+            git reset --hard HEAD
+            yarn prettier-eslint --list-different
 
       - run: yarn test
 
--- a/CONTRIBUTING.org
+++ b/CONTRIBUTING.org
@@ -65,9 +65,13 @@ We have good quality assurance and an established workflow. This is it:
 4. New Branch in Git
 5. Naming: =(feature|fix|chore)/issue-id/short-description=
 6. Develop / Test locally until tests pass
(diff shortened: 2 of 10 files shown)


.circleci/config.yml
```diff
@@ -52,7 +52,26 @@ jobs:
...
+      - run:
+          name: eslint
+          command: yarn eslint
+
+      - run:
+          name: prettier
+          # Make the build output more useful by immediately showing full details of
+          # any style violations.
+          command: |
+            yarn prettier --write
```
munen, line 64: It looks like this is causing a problem: https://app.circleci.com/pipelines/github/aspiers/organice/90/workflows/1ed822e3-a87b-4132-bab8-247359674567/jobs/102/steps

`prettier` and `prettier-eslint` disagree on how to format a passage of code. I can preproduce that on my machine. The checked in code is what `prettier-eslint` generates. However, `prettier` wants it different. I propose we only use the `eslint-task`. It also accepts the `--write` flag.

aspiers: Good catch - adjusted. Let's see if the build works.

munen: :+1:

aspiers: Didn't work :-(  Trying to figure out why - just pushed again.

munen: :+1:

aspiers: Ohh, it was probably due to the merge from `master` which I only just spotted.  I'll rebase instead.

munen: :+1: 

I merged, because there were conflicts. And I chose the solution that `prettier-eslint` let me(;

aspiers: I've now rebased, and updated the commits/docs to reflect our preference of `prettier-eslint` over `prettier`.  Build succeeded, so I think this is ready to merge now.

aspiers:
Why `-eslint-config-path path_to_eslintrc` and not `-eslint-config-path eslintrc.yml`?

munen:
Because the command is called from Emacs, so it needs an absolute path to find the config file. Otherwise it’ll use defaults.

aspiers:
> Because the command is called from Emacs, so it needs an absolute path to find the config file. Otherwise it’ll use defaults.

Gotcha!  As mentioned on Matrix, the solution seems to be to include the `--stdin-filepath` option as per https://github.com/prettier/eslint-plugin-prettier/issues/195#issuecomment-493446593.  So can you make your emacs supply that?  Similarly, I need to somehow resolve https://github.com/jscheid/prettier.el/issues/23

munen:
@aspiers Sweet, you've found it! :partying_face: 

I can easily add that to my config: https://github.com/munen/emacs.d/commit/ce247e4ab968e04113da2a9f3509ab3fd93a5e68

PR #356, opened 2020-06-13
doc: WebDAV test setup

munen:
At this time, I cannot test any new proposed WebDAV configuration. When I wrote the original docs, I did set this up locally, so that I could test WebDAV. Since then, I have switched computers and haven't set this up, again.

The best way to move forward would likely be to create this Apache setup in a `Dockerfile`. I've been thinking of doing that for some time now. It would ease a lot of discussions on the topic of WebDAV. Right now, everyone is using some very specific setup (NextCloud, HA Proxy, Apache, etc) and it's not really possible to give support for that. Such Ops topics are way out of scope of a SPA project. If we had an official example Docker container that works for us and we use in development, we can always point to that one.

### DIFF
--- a/WIKI.org
+++ b/WIKI.org
@@ -52,79 +52,80 @@ to your Dropbox, here are your options:
 
  [[https://github.com/200ok-ch/organice/wiki/videos/demo-webdav.gif]]
 
-*** Missing CORS headers
+*** WebDAV test server
 
- If your back-end doesn't have the necessary CORS headers set (more
- info in the [[https://github.com/200ok-ch/organice/blob/master/README.org][README]], this is what the error will look like:
+For testing purposes, we use a Docker image with a proven and well
+documented server: Apache2 running on Debian. You can build the Docker
+image yourself - and customize the Apache configuration to your needs:
 
- [[https://github.com/200ok-ch/organice/wiki/videos/demo-webdav-failing-cors.gif]]
+#+BEGIN_SRC shell
+docker build -f doc/webdav/Dockerfile -t apache-webdav .
+#+END_SRC
 
-*** Setup your own WebDAV Server with Apache2 on Debian
+Then run the Apache2 WebDAV server with:
 
- For testing purposes, here are the instructions to setup WebDAV
- locally on your machine using Apache2 using Debian. These instructions
- are not meant to be used in production, though.
+#+BEGIN_SRC shell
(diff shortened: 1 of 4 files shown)


WIKI.org
```diff
@@ -52,79 +52,80 @@ to your Dropbox, here are your options:
...
+On your host machine, you can now login with any WebDAV client using
+the URL =http://localhost:8080/webdav=. There is no authentication
+configuration, so any user account works (including omitted user
+accounts). It goes without saying that if you wanted to use this for
+production, please enable authentication. Within the test image,
+you'll find the [[file:sample.org][sample.org]] file, so you can get started developing and
+testing right away.
+
+For testing WebDAV outside of organice, and you're an Emacs user, you
+can open [[/dav:localhost#8080:/webdav/][this]] link (=C-c C-o=). Then, you will get a [[https://www.gnu.org/software/tramp/][TRAMP]] session
```
schoettl, line 80: Very nice Org mode examples here :) I just discovered TRAMP and babel (shell) and now use it a lot in my org files, e.g. to generate hledger reports right in Org mode :)

munen: Thanks^^

Yes, TRAMP and Babel source blocks are mighty powerful - I'm using them all the time. At 200ok, we also use Org in combination with ledger(; We also use it for controlling, writing quotes and invoices. With such literate docs, we gain transparency and reproducability. Plus, it's just a lot more fun to do these chores like that(;

aspiers: Oooh... I knew about TRAMP and babel, but hledger is a very cool new find for me :laughing:  Thanks!

aspiers:
I have tested this docker image and verified your instructions work for me.

WIKI.org
munen: If you're into text based accounting (and who wouldn't want to use version control on serious money related stuff?), maybe I can interest you in another FOSS project of ours: easy.

(h)ledger only solves the data format and versioning problem. However, the entries are still mutable state. We've written a system that's adding event orientation to the game. Doing bookkeeping like that has been a game changer for us. Another additional benefit is that I (as a business owner) don't really want to think in accounting terms, I want to think in 'business event' terms (writing invoices, paying invoices, paying employees and so forth). With easy, I can actually focus on that and it takes care of the rest. 

https://github.com/200ok-ch/easy

munen:
Sweet, thank you all for the positive feedback. I'm pulling this in, then.

PR #359, opened 2020-06-14
doc: Add a bunch of CUSTOM_ID properties for permalinks

aspiers:
This continues on from the work in #343, for more fully resolving #342.


aspiers:
BTW there seems to be a pre-existing inconsistency regarding use of `-` vs. `_` in `CUSTOM_ID` properties.  Perhaps we should decide on one way?

munen:
Adding more CUSTOM_IDs makes sense. I'll happily pull your changes!

But first, where did you see inconsistencies? I quickly checked and can't find them:

```
munen@lambda:~/src/200ok/organice% ag CUSTOM_ID | awk {'print $3'} | ag "\-|\_"
faq_dropbox
faq_webdav
building_docs
offline_support
in_buffer_settings
synchronization_back_ends
capture_templates
media_capture
```

PR #360, opened 2020-06-14
CI: Add doc compilation to CircleCI config

aspiers:
This reduces the risk of merging broken docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
             - yarn-packages-{{ checksum "yarn.lock" }}
 
       - run:
-          name: Install Dependencies
+          name: Install Javascript Dependencies
           command: yarn install --frozen-lockfile
 
       - run:
@@ -71,6 +71,14 @@ jobs:
             yarn prettier-eslint --list-different
 
       - run: yarn test
+
+      - run:
+          name: Install Documentation Dependencies
+          command: sudo apt-get update -yq && sudo apt install -yq pandoc emacs
+
+      - run:
+          name: Compile documentation
+          command: ./bin/compile_doc.sh
 
   push-image:
     machine: true


munen:
Compiling the docs is already part of CI. However, it's not part of the tests (and it shouldn't be). When the tests run, it's part of the subsequent deploy step which requires the tests to be green: https://github.com/200ok-ch/organice/blob/317b2f75206e9243ca430627c3c02f2881c8ceb0/.circleci/config.yml#L100

Now, the installation of dependencies for that is twice in config.yml(;

aspiers:
[**@munen**](https://github.com/munen) commented on [June 14, 2020 5:29 PM](https://github.com/200ok-ch/organice/pull/360#issuecomment-643789944 "2020-06-14T16:29:37Z - Replied by Github Reply Comments"):
> Compiling the docs is already part of CI. However, it's not part of the tests (and it shouldn't be).

Why shouldn't it?  The docs are as important as the code IMHO, so they deserve testing too.

> When the tests run, it's part of the subsequent deploy step which requires the tests to be green:
> 
> https://github.com/200ok-ch/organice/blob/317b2f75206e9243ca430627c3c02f2881c8ceb0/.circleci/config.yml#L100

Maybe I'm misunderstanding something, but I'm 99% sure this `deploy` step doesn't run on PRs before they're merged.  For example look at PR #361 which I submitted to demonstrate this point.  It breaks `bin/compile_doc.sh`, and the CI doesn't spot it.  If that was a serious PR which looked good to all reviewers and you merged it, I assume the `deploy` step would have broken, which would be pretty nasty.  Hence this proposal to check compilation *before* merging, not just to run it after merges.

> Now, the installation of dependencies for that is twice in config.yml(;

Yes, that is unfortunate but unavoidable I suspect.  And it's hardly a big issue anyway.

munen:
> Why shouldn't it? The docs are as important as the code IMHO, so they deserve testing too.

Well, that has some merit to it..

However, does that imply that _all_ artifacts should be test created? That would include Docker images for organice and for Apache WebDAV, for example. 

My original logic was that it shouldn't be part of the tests, because compiling the docs was coupled to uploading them. You fixed that.

> Maybe I'm misunderstanding something, but I'm 99% sure this deploy step doesn't run on PRs before they're merged.

It doesn't run. After merging, I'd get it in the notifications firehose and via email and would fix it.

> Yes, that is unfortunate but unavoidable I suspect. And it's hardly a big issue anyway.

Well.. I wouldn't say that duplication and slower tests are generally not a big issue. For this particular situation, I don't know what the time penalty is, yet, so it might not apply.

munen:
Maybe you know already, but we did actually break building the docs and it's not showing as a failure (even on merge): https://app.circleci.com/pipelines/github/200ok-ch/organice/745/workflows/4004fca6-0ca5-4f48-a2f1-ceb3f27fedf1/jobs/1257/parallel-runs/0/steps/0-105

*lol*

aspiers:
[**@munen**](https://github.com/munen) commented on [June 14, 2020 7:12 PM](https://github.com/200ok-ch/organice/pull/360#issuecomment-643801996 "2020-06-14T18:12:26Z - Replied by Github Reply Comments"):
> > Why shouldn't it? The docs are as important as the code IMHO, so they deserve testing too.
> 
> Well, that has some merit to it..
> 
> However, does that imply that _all_ artifacts should be test created? That would include Docker images for organice and for Apache WebDAV, for example.

That's a good point and a nice idea.  They could even be set up to run but without a requirement for them to pass - i.e. as a warning only.  However let's walk before we run ;-)  

> My original logic was that it shouldn't be part of the tests, because compiling the docs was coupled to uploading them. You fixed that.

Ah yes, I forgot about that ;-)

> So, generally speaking, we can totally discuss adding it to the CI pipeline.
> 
> > Maybe I'm misunderstanding something, but I'm 99% sure this deploy step doesn't run on PRs before they're merged.
> 
> It doesn't run. After merging, I'd get it in the notifications firehose and via email and would fix it.

Right.  I think it's much more ideal if it gets caught in the review phase, because a) that reduces the risk of you personally having to fight fires, and b) it might well highlight a fundamental problem with the PR. 

> > Yes, that is unfortunate but unavoidable I suspect. And it's hardly a big issue anyway.
> 
> Well.. I wouldn't say that duplication and slower tests are generally not a big issue. For this particular situation, I don't know what the time penalty is, yet, so it might not apply.

28 seconds according to [this build](https://app.circleci.com/pipelines/github/aspiers/organice/96/workflows/4274e3af-90f4-4b46-8596-68e4eb8179e7/jobs/110).

aspiers:
[**@munen**](https://github.com/munen) commented on [June 14, 2020 7:15 PM](https://github.com/200ok-ch/organice/pull/360#issuecomment-643802295 "2020-06-14T18:15:03Z - Replied by Github Reply Comments"):
> Maybe you know already, but we did actually break building the docs and it's not showing as a failure (even on merge, it's still green, but it doesn't run): [app.circleci.com/pipelines/github/200ok-ch/organice/745/workflows/4004fca6-0ca5-4f48-a2f1-ceb3f27fedf1/jobs/1257/parallel-runs/0/steps/0-105](https://app.circleci.com/pipelines/github/200ok-ch/organice/745/workflows/4004fca6-0ca5-4f48-a2f1-ceb3f27fedf1/jobs/1257/parallel-runs/0/steps/0-105)
> 
> _lol_
> 
> Update: I tried fixing it, but it still fails: [8a894f0](https://github.com/200ok-ch/organice/commit/8a894f027fe469532e8b83102a67c293e526f3db)

Ouch :laughing:  That kind of proves the need for this, doesn't it?

munen:
> Ouch laughing That kind of proves the need for this, doesn't it?

Unfortunately, no. Like I said, the failing build doesn't show up. When the export fails, then the return value is still `0`, so CI doesn't pick up on a problem. I saw this manually, because I wanted to read some docs and saw that they were outdated. After confirming that it's not a caching problem, I investigated.

Update on that front, this should fix it: https://github.com/200ok-ch/organice/commit/0e3c8dd2147ebdc2089be757d188ad48530b6dca

munen:
> Right. I think it's much more ideal if it gets caught in the review phase, because a) that reduces the risk of you personally having to fight fires, and b) it might well highlight a fundamental problem with the PR.

That would be nice, if we didn't just have a failure that wasn't picked up by CI (after merging and publishing).

If it would actually be picked up, then it would be a benefit, I concur!

Without it, it would:

- add 28s to the 15s of running the tests
- add a false sense of security

I propose to find out how CI can actually catch an Emacs export problem. We can use the above failure situation as a baseline on what should be caught.

munen:
Update: That last CI run was successful, again: https://app.circleci.com/pipelines/github/200ok-ch/organice/747/workflows/a72f93ee-cd8a-43fe-9b7c-3c29f9f31f1a/jobs/1264/parallel-runs/0/steps/0-105

aspiers:
[**@munen**](https://github.com/munen) commented on [June 14, 2020 8:03 PM](https://github.com/200ok-ch/organice/pull/360#issuecomment-643808163 "2020-06-14T19:03:55Z - Replied by Github Reply Comments"):
> > Ouch laughing That kind of proves the need for this, doesn't it?
> 
> Unfortunately, no. Like I said, the failing build doesn't show up. When the export fails, then the return value is still `0`, so CI doesn't pick up on a problem.

This is trivially fixed by #363.  I don't think it negates any of the benefits I outlined.

> I saw this manually, because I wanted to read some docs and saw that they were outdated. After confirming that it's not a caching problem, I investigated.
> 
> Update on that front, this should fix it: [0e3c8dd](https://github.com/200ok-ch/organice/commit/0e3c8dd2147ebdc2089be757d188ad48530b6dca)

Even though I don't yet entirely understand that commit (and don't have time to dig into it, but don't need to anyway), I fully trust your judgement that that is a worthwhile change :-)  But I think #363 is still definitely needed as well.

[**@munen**](https://github.com/munen) commented on [June 14, 2020 8:08 PM](https://github.com/200ok-ch/organice/pull/360#issuecomment-643808686 "2020-06-14T19:08:06Z - Replied by Github Reply Comments"):
> > Right. I think it's much more ideal if it gets caught in the review phase, because a) that reduces the risk of you personally having to fight fires, and b) it might well highlight a fundamental problem with the PR.
> 
> That would be nice, if we didn't just have a failure that wasn't picked up by CI (after merging and publishing).
> 
> If it would actually be picked up, then it would be a benefit, I concur!

With #363, it should be picked up :-)

> Without it, it would:
> 
> *   add 28s to the 15s of running the tests
> *   add a false sense of security
> 
> I propose to find out how CI can actually catch an Emacs export problem. We can use the above failure situation as a baseline on what should be caught.

Sounds good!  Thanks for your time on this! :-)

aspiers:
I've submitted #364 as a time-saving alternative to this.

munen:
Closing since we merged #363 and #364 where aspiers implement the same functionality, but even better^^

PR #364, opened 2020-06-15
CI: Add parallel doc compilation to CircleCI config

aspiers:
This reduces the risk of merging broken docs.  Run this in parallel
with the normal build to save time.

If this works, it can supersede #360.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
             - yarn-packages-{{ checksum "yarn.lock" }}
 
       - run:
-          name: Install Dependencies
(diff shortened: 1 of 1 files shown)


aspiers:
Worked fine AFAICS.

aspiers:
Note that the `build-docs` job is not currently configured as a "Required" status check within GitHub.  If merging this, I guess it would make sense to change that.

munen:
Sweet, I tested #363 and #364 together in #365. 

Works perfectly, happy to pull both in! Good work :pray: 

NB: I followed your advice to check 'build-docs' as required check.

image

PR #423, opened 2020-08-09
Add note to header (org-add-note)

schoettl:
This adds a button for taking notes to tasks (`org-add-note`). They will be prepended to the header contents after `:PROPERTIES:` (and apparently before the `:LOGBOOK:`):

```org
* test header
- Note taken on [2020-08-06 Do 10:49] \\
  Passt jetzt
- Note taken on [2020-08-05 Mi 18:12] \\
  Neue Ergebnisse von ... in org anschauen.
- Note taken on [2020-08-05 Mi 09:34] \\
  Faktor angepasst -> ... nochmal testen

```

@munen 
I have some problems with understanding React (again)... where do I link the reducer function to the button?

(I know that I have to move code with side-effects out of the reducer function, i.e. `prompt()`. It's WIP currently.)

Is it OK, to use JS `prompt()`? It should be available on all platforms and is probably "good enough". A multi-line input dialog can still be added later.

The log message "Note taken on [TIMESTAMP] \\\\" (`org-log-note-headings`) can later be made a setting.

Should input lines be hard-wrapped? In orgmode, emacs wraps it while I type. So it's probably a good idea to also do that in Organice.

## TODO
- [x] line wrap for long notes
- [x] some minor whitespace problems?
- [x] refactor: note composition in reducer function
- [x] refactor reducer using `addNote`
- [x] refactor reducer getting current time (impure) -> marked with TODO comments
- [x] get clicks on timestamps in logNotes working
- [x] logNotes parser doesn't seem to stop at empty line after notes
- [x] get tests running (indentation problem)
- [x] fix new reducer tests that I've broken
- [x] newly created notes do not show up in the header immediately


munen:
@schoettl Cool that you started working on this! :+1: 

>  I have some problems with understanding React (again)... where do I link the reducer function to the button?

Not directly. Via the action that you've set up pretty well already(;

I just did the "React-style" wiring together of your code and left some TODO notes: https://github.com/200ok-ch/organice/pull/423/commits/eca2067fab4fdbf8f07e6df5003ee7e66c8232ae

Good job so far!

> Is it OK, to use JS prompt()? It should be available on all platforms and is probably "good enough". A multi-line input dialog can still be added later.

Works for me! Thanks for checking^^

> Should input lines be hard-wrapped? In orgmode, emacs wraps it while I type. So it's probably a good idea to also do that in Organice.

I haven't researched it, but I trust your judgment on that 100%.

Keep up the good work, thanks again for starting work on this feature! :pray:

schoettl:
Thanks for your feedback and help :)

munen:
Any time^^

src/components/OrgFile/components/Header/index.js
```diff
@@ -305,6 +308,26 @@ ${header.get('rawDescription')}`;
...
+    // Getting indentation, timestamp, and template string is impure, hence do it here:
+    const indentation = dontIndent ? '' : ' '.repeat(header.nestingLevel + 1);
+    const timestamp = moment().format('YYYY-MM-DD HH:MM');
+    // Generate note based on a template string (as defined in org-log-note-headings):
+    const noteText = `${indentation}- Note taken on [${timestamp}] \\\\
+${indentation}  ${input}`;
+    // TODO: wrap input; sth. like this https://codereview.stackexchange.com/a/171857
+
+    console.log(noteText);
+    this.props.org.addNote({ noteText });
```
schoettl, line 328: How can I pass the `action` to the `addNote` reducer? (And how is the `state` passed?)

In your the previous version, @munen , without `{ }`, the argument did not go through either.

src/components/OrgFile/components/Header/index.js
```diff
@@ -305,6 +308,26 @@ ${header.get('rawDescription')}`;
...
+  handleAddNoteClick() {
+    let input = prompt('Enter a note to add to the header:');
+    if (input !== null) input = input.trim();
+    if (!input) return;
+    const { header } = this.props;
+
+    const dontIndent = this.props.dontIndent;
+
+    // Getting indentation, timestamp, and template string is impure, hence do it here:
+    const indentation = dontIndent ? '' : ' '.repeat(header.nestingLevel + 1);
```
schoettl, line 320: @munen I'm not sure if I got your previous comment right (on composing the note in the reducer). Anyway, getting the indentation, template string and the timestamp is impure code and the last step, filling the template string can be done right here, I think.

src/reducers/org.js
```diff
@@ -518,6 +518,28 @@ const refileSubtree = (state, action) => {
...
+    // TODO: Don't do it like this, because this overwrites the
+    //       complete body of the header.
+    // TODO: I don't think it works like this. Probably it's a good
+    //       idea to extend `_updateHeaderFromDescription` from
+    //       lib/parse_org.js.
+    // It's not enough to set rawDescription; the whole header must be
+    // reconstructed. Either parse modified rawDescription or directy
+    // update description (which is parsed).
+
+    header.update('rawDescription', (rawDescription) => `${noteText}\n${rawDescription}`)
```
schoettl, line 539: I still have to think about how to best integrate the note in the header... it's not as trivial as I initially thought ;)

schoettl: I think I'll go with prepending/inserting to `rawDescription` and then parse the whole thing again.

I have to check, but I think the notes are better seen as part of the description, not as one of the "prefix elements"... because it should end up as "parsed description".

src/components/OrgFile/components/Header/index.js
schoettl: I only get `undefined` in the reducer...

munen: > How can I pass the action to the addNote reducer?

You don't pass `actions` to `reducers`. `actions` tie state and `reducer` together. 

>  (And how is the state passed?)

I added that bit to the wiring that I've provided: https://github.com/200ok-ch/organice/pull/423/files#diff-2b14c674e803e97fef65139e337183fbR317

In case it's not clear what an action and what a reducer does, I suggest to read the 'core principles' of Redux: https://redux.js.org/introduction/core-concepts or even do the tutorial: https://redux.js.org/basics/basic-tutorial/

On that note, we don't have to do some of the manual wiring of Redux, because we're also using react-redux which also has good documentation: https://react-redux.js.org/introduction/quick-start

munen: > I only get undefined in the reducer...

I apologize, I forgot to update one function signature in my last refactor before the push. I fixed it now: https://github.com/200ok-ch/organice/pull/423/commits/7ac512a9661b40bbf88a5927cb98f65e44637092

src/components/OrgFile/components/Header/index.js
munen: > Anyway, getting the indentation, template string and the timestamp is impure code and the last step, filling the template string can be done right here, I think.

You can do it right here. But then it's harder to test and you have coupling to the component.

Also all the things that you mentioned shouldn't be impure unless I misunderstand something.

In any case, it's not a big issue. You can do it in the component and when the PR is done, and if I'm right that the things you brought up are in fact pure, then I can easily refactor it into the reducer before merging the PR. I'll happily take over that part.

src/reducers/org.js
munen: > I still have to think about how to best integrate the note in the header... it's not as trivial as I initially thought ;)

It's definitively not trivial. It'll have to be included in the domain model that we have for a header. I've documented the way I'd do it here: https://github.com/200ok-ch/organice/pull/423/files#diff-d675f8aab42787975acbd1fe26257689R530

src/components/OrgFile/components/Header/index.js
munen: @schoettl You are correct that changes to the state should only be made with pure functions. That is the third of the [three principles of Redux](https://redux.js.org/introduction/three-principles).

However, that does not include a query for `dontIndent`. The reason is that this information is in the store - and the store is the single source of truth. That's the first principle of Redux. Think of it like that: It is as if the state was an implicit input to the function you're writing.

Creating the `noteText` should be done in the reducer. Otherwise the functionality of this feature would be tightly coupled to the component in which it is happening. However, components shouldn't hold this kind of business logic for multiple reasons. One: It's not possible to test the functionality outside the scope of a fully hydrated component.

But as I mentioned before: If this is unclear, you can totally build the `noteText` in the component. And when you're done, I'll refactor it to the Reducer and maybe add a test(;

schoettl: @munen 
Great, totally clear. I'll pass just the time and input text to the reducer. However, I already refactored and reuse the current `addNote` function from another reducer with a different text (the log when repeating tasks are done). I'll refactor this...

I also noticed that "get current time" is used in some other reducers. I'd refactor this to pass the time from the component, do you agree?

src/reducers/org.js
```diff
@@ -518,6 +518,18 @@ const refileSubtree = (state, action) => {
...
 
+const addNote = (state, action) => {
+  // See org-add-note (C-c C-z) and variable org-log-note-headings
+  const { noteText } = action;
+
+  const headerId = state.get('selectedHeaderId');
+  const headers = state.get('headers');
+  const headerIndex = indexOfHeaderWithId(headers, headerId);
+  return state.updateIn(['headers', headerIndex, 'logNotes'], (logNotes) =>
+    parseRawText(noteText + (logNotes ? `\n${attributedStringToRawText(logNotes)}` : ''))
```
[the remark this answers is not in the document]
munen: Sounds reasonable :+1:

munen:
I just read through the diff. Looks like this feature is coming together nicely :+1:

src/components/OrgFile/components/Header/index.js
```diff
@@ -305,6 +308,37 @@ ${header.get('rawDescription')}`;
...
 
+  handleAddNoteClick() {
+    let input = prompt('Enter a note to add to the header:');
+    if (input !== null) input = input.trim();
+    if (!input) return;
+
+    // Wrap line at 70 characters, see fill-column in "Insert note" window (C-c C-z)
+    const wrappedIndentedInput = this.formatTextWrap(input, 70).replace(/\n(.)/, '\n  $1');
+    // Getting timestamp and template string is impure, hence do it here:
+    const timestamp = moment().format('YYYY-MM-DD HH:MM'); // TODO: use getCurrentTimestamp and export timestamp to text?
```
munen, line 319: Maybe do it similarly like: https://github.com/200ok-ch/organice/blob/master/src/lib/timestamps.js#L79

src/components/OrgFile/components/Header/index.js
munen: >  Great, totally clear. I'll pass just the time and input text to the reducer. However, I already refactored and reuse the current addNote function from another reducer with a different text (the log when repeating tasks are done). I'll refactor this...

:+1: 

> I also noticed that "get current time" is used in some other reducers. I'd refactor this to pass the time from the component, do you agree?

That sounds like a great idea! It'll make testing them easier, too. Otherwise the only option to reliably test them are:

- Mock time
- Use Regexps to test for 'time and dates' in the results instead of a specific time and date

:pray:

munen:
Side note: I'm looking forward to this feature more and more as I start to use it in our project management Org files. It makes a lot of sense to have 'note updates' in a structured manner. Thank you for pointing me to this feature of Org mode :pray:

schoettl:
I broke some of the new reducer tests @necto ... although I didn't touch the `advanceCheckboxState`. I'll look into the tests. But refactoring the reducers right now, I'm very thankful for the new tests :)

@munen You said, settings are in the store, in the state...
https://github.com/200ok-ch/organice/pull/423/files#diff-d675f8aab42787975acbd1fe26257689R186
I guess same applies for the `state.selectedHeaderId`. So why is `selectedHeaderId` and `logIntoDrawer` passed at all? Do you think I can refactor this?

munen:
> But refactoring the reducers right now, I'm very thankful for the new tests :)

Just-In-Time-Tests! JITT?(;

> I guess same applies for the `state.selectedHeaderId`. So why is `selectedHeaderId` and `logIntoDrawer` passed at all? Do you think I can refactor this?

Generally speaking, it does apply for `state.selectedHeaderId`.

I have two theories as to why it's passed:

1. It might just be redundant and could easily be refactored.
2. There's places in the code where the `selectHeader` org action is not called before another action wants to use the selected header.

Without looking through all of the code, my first hunch would be that both are bad style. The first certainly is. For the second, I'd have to read through all of it to see if there's justification.

Maybe there's a third option, too, but I'm quite sure it's one of the two reasons above.

> Do you think I can refactor this?

That would be great!(;

schoettl:
OK, there is one problem left:

My current log notes parser parses lists (bullet point character `-`) at the beginning of headline contents. The reason for parsing log notes is that they come *before* the logbook. Therefore they cannot be part of the description because the description starts *after* the logbook.

If no logbook exists, log notes are still parsed as log notes, not as description. That's the basic reason why 9 tests are broken.

Example: The current log notes parser parses
```
* header
- [ ] simple list that is actually not log notes
```
as log notes with one item and an empty description!

To avoid this, I would change the parser to parse log notes only if a logbook is also present. Otherwise, the log notes or list is parsed as description. New log notes (`addNote`) will still be added to log notes but when exported and then parsed again, they will be description unless there is a log book.

munen:
@schoettl That _does_ sound like a tricky parser problem. I do have a suggestion.

You're implementing the default behavior of Emacs Org mode `org-add-note` right now. That's definitively the right way to go for a new organice feature. But if it turns out too hard and there's alternatives in Emacs Org mode, we can consider them.

If I understand you correctly, this new feature would be a lot easier if it emulated the behaviour of Emacs Org mode when `org-log-into-drawer` is set to `t` (the default is `nil`). That's it's documentation:

```
org-log-into-drawer is a variable defined in ‘org.el’.
Its value is t
Original value was nil

  This variable has an alias: ‘org-log-state-notes-into-drawer’.
  You can customize this variable.

Documentation:
Non-nil means insert state change notes and time stamps into a drawer.
When nil, state changes notes will be inserted after the headline and
any scheduling and clock lines, but not inside a drawer.

The value of this variable should be the name of the drawer to use.
LOGBOOK is proposed as the default drawer for this purpose, you can
also set this to a string to define the drawer of your choice.

A value of t is also allowed, representing "LOGBOOK".

A value of t or nil can also be set with on a per-file-basis with

   #+STARTUP: logdrawer
   #+STARTUP: nologdrawer

If this variable is set, ‘org-log-state-notes-insert-after-drawers’
will be ignored.

You can set the property LOG_INTO_DRAWER to overrule this setting for
a subtree.

Do not check directly this variable in a Lisp program.  Call
function ‘org-log-into-drawer’ instead.
```

If you wanted to go that route, and save some time building a crazier parser (which we'll make obsolete at some point with `org-parser` potentially), then I'm in favor to go that route. If there's downsides and you don't want to go that way, that's OK for me, of course. I just want to speak up in case it saves you work.

NB: All things being equal, I personally like the option to add the note into the logbook drawer. It somehow feels cleaner. Not saying that the other option is not valid, just saying that it wouldn't be bad at all imo to do it that way^^

munen:
Or, you could emulate the behaviour if `org-log-state-notes-insert-after-drawers` were set to `t` which will make the log notes part of the description.

schoettl:
Thanks @munen - damn, org mode has so many options... generally I'd like to stick to the standard/default behavior of org mode. It turned out, that the change I desribed in my previous comment actually simplified the parser :)

And now, all tests pass at my machine. Is it related to the other current issue, that they fail in circleci?

munen:
> It turned out, that the solution I described in my previous comment actually simplified the parser :)

Oh, sweet!

Of course, if the default behaviour can be implemented, that would be the best option(;

munen:
> And now, all tests pass at my machine. Is it related to the other current issue, that they fail in circleci?

No. I took the liberty to fix the error from CI: 29a77d0

munen:
But since you asked, out of curiosity: Have you experienced flaky test behaviour on your machine in organice? Tests that don't run, but then run on a retry?

schoettl:
> > And now, all tests pass at my machine. Is it related to the other current issue, that they fail in circleci?
> 
> No. I took the liberty to fix the error from CI: [29a77d0](https://github.com/200ok-ch/organice/commit/29a77d0462f58a7183e16b5c4b930838c7bd76ef)

Thanks. I think, sometimes, the linter errors only show up after restarting `yarn start`.

No, I didn't notice random behaviour of tests.

schoettl:
Alright, I think, I fixed the last point on the checklist. Ready to merge after review :)

munen:
> Thanks. I think, sometimes, the linter errors only show up after restarting yarn start.

That actually seems reproducible (not that it's a good thing). I'd say that it behaves the same for me. It's more reliable in Emacs through flycheck. It would be best if we had a command that we can run. Let's see if we find one^^

> No, I didn't notice random behaviour of tests.

That's good to hear! :+1:

schoettl:
> Alright, I think, I fixed the last point on the checklist. Ready to merge after review :)

Well, I should first review my own PR again ^^

munen:
> Well, I should first review my own PR again ^^

Perfect. I'll let you go over it and when you ping, I'll use my next organice slot to review it.

NB: I just did a first test on my machine and from there, it worked as advertised! So far so good^^

schoettl (commented):
Looks good to me. Refactoring of reducers/org.js regarding impure new Date() is not finished. I only added TODO comments, but I would postpone this.

schoettl:
Oh, I found one bug: The timestamps in log notes are not editable. The timestamp editor pops up but I cannot change the timestamp. Any idea?

munen:
> Oh, I found one bug: The timestamps in log notes are not editable. The timestamp editor pops up but I cannot change the timestamp. Any idea?

Not really a reason, but related to timestamps: I just started testing and saw that notes from organice lack the timestamp.

```
   - Note taken on [2020-08-29 Sat 19:59] \\
     from emacs
   - Note taken on [2020-08-29 Sat] \\
     from organice
```

munen:
I confirm: Changing the timestamp does not work for a new timestamp. When reloading the file, it works.

So far, I checked that changing a new timestamp works for: "scheduled", "deadline", "headline" and "description". I think that's all the other places that we have timestamps.

So, the reason will be in finding what's different for those(;

schoettl:
> So, the reason will be in finding what's different for those(;

Hm yes... the timestamps are generated here (by parsing the `noteText`).
https://github.com/200ok-ch/organice/pull/423/files#diff-d675f8aab42787975acbd1fe26257689R531
They have an id, the correct type, the firstTimestamp...

grafik

I don't see a difference in their parsed structure to other, changeable timestamps.

schoettl:
I think I got it...

schoettl:
Still one (very minor) problem: headlines with active timestamps in log notes don't show up in agenda. I don't know why. I added them to planningItems but still. It's probably not hard to fix but - to be honest -, I'm tired of this, I'd rather see this feature in production...

munen:
> Still one (very minor) problem: headlines with active timestamps in log notes don't show up in agenda. I don't know why. I added them to planningItems but still. It's probably not hard to fix but - to be honest -, I'm tired of this, I'd rather see this feature in production...

I confirm: I can reproduce and agree that it's a very minor issue and would like to see the feature in production(;

Starting with the review now.

src/reducers/org.js
```diff
@@ -183,15 +184,14 @@ const updateCookiesOfParentOfHeaderWithId = (state, headerId) => {
 };
 
 const advanceTodoState = (state, action) => {
-  const headerId = action.headerId || state.get('selectedHeaderId');
-  if (!headerId) {
+  const { headerId, logIntoDrawer, currentDate } = action; // TODO: isn't logIntoDrawer a setting in the store? then why to pass it?
```
munen: @schoettl Because it's not in _the_ store. Not in the `org` store that is. It's in the `base` store.

src/reducers/org.js
```diff
@@ -972,7 +972,7 @@ const addNewPlanningItem = (state, action) => {
   const newPlanningItem = fromJS({
     id: generateId(),
     type: action.planningType,
-    timestamp: timestampForDate(new Date()), // TODO: impure new Date()
+    timestamp: timestampForDate(action.timestamp),
```
schoettl, line 976: Is there a naming convention for date/time/timestamp, ...? `timestamp` in Organice is mostly used for the parse structure of a timestamp (see timestamp.js, `getCurrentTimestamp` and so on). So I'd suggest `currentDate` for a `new Date()` timestamp. Or is there another convention in JS on how to call these? Like `dateTime` or `time`?

src/reducers/org.js
```diff
@@ -1312,10 +1317,10 @@ function updatePlanningItemsWithRepeaters({
     currentTodoSet.get('keywords').first()
   );
   if (!noLogRepeatEnabledP({ state, headerIndex })) {
-    const lastRepeatTimestamp = timestampForDate(new Date(), {
+    const lastRepeatTimestamp = timestampForDate(timestamp, {
```
schoettl, line 1320: I was stuck at this point... what is the parameter `currentDate`? Isn't it also a `new Date()` timestamp, like your new parameter `timestamp`? Why was it not used in this place? I coudn't find out yet.

src/reducers/org.js
munen: I agree that `currentDate` would be a better name overall for this.

The reason I named it `timestamp` is because we're assining it to another attribute that already has the name `timestamp`. So I was working on removing the `TODO` regarding impurity in the reducer and didn't want to go down the rabbit hole of finding all the `timestamp`s that should be `currentDate`. That's better done in a second step. But I agree, it would be much more appropriate naming. In general in JS, but specifically in organice since a `timestamp` already is a well known entity.

schoettl:
(I'm just working again on the planningItem/Agenda problem.)

src/reducers/org.js
munen: I'm sorry, I might not fully understand the question. I'll try to answer: Yes, in this spot, there was a call to `new Date()` as it was in others. I replaced them all with the same `new Date()` call made in an action: https://github.com/200ok-ch/organice/pull/423/commits/cf6e2849c5fd40e73b6203e69a9401c4bef54c8a

schoettl: I mean, this function (line 1306) already has a parameter `currentDate`. Now it has `timestamp` and `currentDate` – are they the same thing?

munen:
> (I'm just working again on the planningItem/Agenda problem.)

Thanks for letting me know! I'm close to finishing with my review.

Let's not force push for the moment(;

src/reducers/org.js
munen: Ah, now I get it. Thank you for persisting^^

It's a good catch. `currentDate` is dead code which isn't even passed from the action - alongside other dead code:

image

munen: This even leads to buggy behaviour of code thinking the variables are defined:

image

I'll clean this up.

src/reducers/org.js
```diff
@@ -183,15 +184,14 @@ const updateCookiesOfParentOfHeaderWithId = (state, headerId) => {
 };
 
 const advanceTodoState = (state, action) => {
-  const headerId = action.headerId || state.get('selectedHeaderId');
-  if (!headerId) {
+  const { headerId, logIntoDrawer, currentDate, timestamp } = action; // TODO: isn't logIntoDrawer a setting in the store? then why to pass it?
+  const existingHeaderId = headerId || state.get('selectedHeaderId'); // TODO: why passing headerId at all?
```
munen: @schoettl For this reason: 64a092a

munen (approved):
@schoettl I have finished the review and made minor changes which I've added to semantic commits.

The feature is well implemented, documented and tested. It also works well with my real world Org mode files. I've added docs into sample.org and changelog.org, as well. Personally, I'm very much looking forward to using the feature in organice \\(^_^)/

You said that you've started working on adding the notes with active timestamps into the Agenda. If you want, you can do that in this PR. However, I wouldn't mind (at all) merging this PR. The feature and the code is in a good state. And the PR is already huge (58 entries in conversation, 43 commits and 20 files changed - congrats :partying_face:).

I'll leave the decision up to you. Feel free to merge yourself.

Thank you for the hard and thorough work, as always! :pray: :bowing_man: 

P.S.: I wanted to link to Org mode docs for `org-add-note` in the sample.org as for the other features, but didn't find appropriate docs. One place [seems outdated or is incorrect](https://orgmode.org/manual/Drawers.html), the other only documents the ability [to add notes from the agenda](https://orgmode.org/manual/Agenda-Commands.html). If you've seen Emacs Org mode docs that are appropriate, it would be nice to add them. If not, that's not a big issue for us, merely an opportunity to contribute to orgmode.org(;

schoettl:
Alright I just did some minor refactoring. One thing was that the new, third category of timestamps in notes was not handled.

I still couldn't fix the agenda problem. And I also don't find more documentation on that feature. I thought it's a good idea, to add the Org mode function name to `simple.org`. Just to show Org mode users who don't know the feature, that it's not our own invention.

munen:
Sweet, congrats on the merge! :confetti_ball: 

NB: I forgot to mention in the result of the review that I really appreciate you doing loads of refactoring and improvements along the way when you're working in the codebase. With this pathfinder principle the codebase will get better and better^^

In any case, I'm stoked to be able to use the feature from now on. Thank you for your continued effort!

PR #428, opened 2020-08-17
Fix the INSERT_CAPTURE 'is undoable' test

necto:
redux-undo has two stages for the state history:
first the result of an undoable action (the one that was not filtered by the `filter`
config parameter to the `undoable` function) gets into the `_lastUnfiltered`
transit field, then on the next undoable action it gets stored into the `past`
stack is used to retrieve the previous versions of the state.

The test did not provide any undoable action before running INSERT_CAPTURE
so when it was run redux-undo did not push the previous state into 'past'
and could not retrieve it when and UNDO was requested.

With this PR the test first performs an undoable action to warm up the redux-undo history,
and then checks the stack property of the state history.

### DIFF
--- a/src/reducers/org.unit.test.js
+++ b/src/reducers/org.unit.test.js
@@ -171,10 +171,18 @@ describe('org reducer', () => {
       expect(last.get('rawDescription')).toEqual('Some description\n');
     });
 
-    // FIXME: This test works, but only when run with `.only`. There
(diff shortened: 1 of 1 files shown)


munen:
Wow, what an amazing contribution - I didn’t get the test to work and I’m very grateful for your explanation and fix!

Thank you very much and a warm welcome to the organice contributors list 🙏🏻

PR #431, opened 2020-08-21
Increase the test coverage for src/reducer/org.js

necto:
I would like to propose a test suite for the org format transformers implemented in `src/reducer/org.js`.
This suite covers up to 99.36% of all the statements, which boosts confidence in refactoring and implementing new features.
This changeset touches almost exclusively the test code, except for the following changes in the non-test code:

- package.json: add the coverage target (running `yarn coverage` yields detailed test coverage statistics)
- src/actions/org.js: add a comment clarifying the expected values in the `updateTimestampWithId` arguments
- src/lib/timestamps.js: add a `timestampForDate` function to enable tests involving timestamps
- src/reducers/org.js:
  - fix the potential moveHeaderDown null-reference access
  - Eliminate dead code, provided `parentListItem.has('checkboxState') == parentListItem.get('isCheckbox')`. This is the most disrupting change, I will be grateful if you confirm that this assumption is correct.

The testing revealed one more bug that I would need your advice to fix:

On this line: https://github.com/200ok-ch/organice/blob/master/src/reducers/org.js#L90
the code intent is to close all the nested headers when closing a parent header.

However, the `index` variable on that line ranges over the headers themselves, not their indices (i.e. objects, not integers). That's why the code does not work.

As a result, the nested headers are never closed, but nevertheless they are hidden because their parent header is closed. So when I reopen the parent header the openness of the subheaders is preserved.

Personally, I think this is convenient, and I would preserve the current behavior and removed the erroneous code.
The alternative is to fix the erroneous code and make sure the subheaders are closed.

Shall I (a) remove the erroneous code, or (b) fix it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "start": "./bin/compile_search_parser.sh && react-scripts start",
     "build": "./bin/compile_search_parser.sh && react-scripts build",
     "test": "./bin/compile_search_parser.sh && react-scripts test --env=jsdom",
+    "coverage": "./bin/compile_search_parser.sh && react-scripts test --env=jsdom --coverage --watchAll=false",
     "eslint": "./node_modules/.bin/eslint --cache .",
     "nibble": "./node_modules/.bin/eslint-nibble --cache .",
     "prettier": "./node_modules/.bin/prettier \"**/*.js\"",
@@ -54,6 +55,11 @@
     "prettier-eslint-cli": "^5.0.0",
     "raw.macro": "0.3.1",
     "react-test-renderer": "^16.13.1"
+  },
+  "jest": {
+    "coverageReporters": [
+      "text"
+    ]
   },
   "browserslist": [
     ">0.2%",
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -485,6 +485,13 @@ export const reorderPropertyList = (fromIndex, toIndex) => (dispatch, getState) 
     dirtying: true,
   });
 
+/**
+ * Action to change the timestamp using a cross-cutting id.
+ *
+ * @param {*} timestampId cross-cutting id of the timestamp (might be in the title or description).
+ * @param {*} newTimestamp the new value for the timestamp;
+ *                         must have the form: {id:, type:, firstTimestamp:, secondTimestamp:}.
+ */
 export const updateTimestampWithId = (timestampId, newTimestamp) => ({
   type: 'UPDATE_TIMESTAMP_WITH_ID',
   timestampId,
--- a/src/lib/timestamps.js
+++ b/src/lib/timestamps.js
@@ -48,8 +48,10 @@ export const renderAsText = (timestamp) => {
 };
 
 export const getCurrentTimestamp = ({ isActive = true, withStartTime = false } = {}) => {
-  const time = new Date();
+  return timestampForDate(new Date(), { isActive, withStartTime });
+};
 
+export const timestampForDate = (time, { isActive = true, withStartTime = false } = {}) => {
   const timestamp = {
     isActive,
     year: format(time, 'yyyy'),
--- a/src/reducers/org.js
+++ b/src/reducers/org.js
@@ -816,6 +816,8 @@ const updateParentListCheckboxes = (state, itemPath) => {
           return false;
         case 'partial':
           return false;
+        default:
+          return false;
       }
     })
     .toJS();
--- a/src/reducers/org.js
+++ b/src/reducers/org.js
@@ -368,7 +368,7 @@ const moveHeaderDown = (state, action) => {
   const subheaders = subheadersOfHeaderWithId(headers, action.headerId);
   const nextSiblingIndex = headerIndex + subheaders.size + 1;
   const nextSibling = headers.get(nextSiblingIndex);
-  if (nextSibling.get('nestingLevel') < header.get('nestingLevel')) {
+  if (!nextSibling || nextSibling.get('nestingLevel') < header.get('nestingLevel')) {
     return state;
   }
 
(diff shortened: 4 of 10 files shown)


munen:
Hi @necto

I apologize for taking this long to respond and comment on your PR, I'm at the final stretches of getting a new product ready with a first big enterprise customer already signed up. That'll cross-finance future organice development in the future nicely(;

Having said that, I'd like to start on a completely different note: I'm blown away by this PR! It's a huge improvement on testing and refactorability of the codebase.

I've spent some time looking over the code and all looks good. That includes the change regarding `parentListItem.has('checkboxState')` - it looks good to me.

Regarding the error on openness: That's a great catch and a bug compared with how Emacs Org mode behaves. I would have voted vote to fix it now that you've found the cause for it. But I personally also think it's quite convenient. So now, I'm thinking to add a feature flag for it. What do you think?

Apart from that 'error on openness' discussion, I'll be stoked to merge your PR in. I have only one question: Is there a reason you've removed the [`default` case](https://github.com/200ok-ch/organice/pull/431/files#diff-d675f8aab42787975acbd1fe26257689L819) for `childCompletionStates` from `updateParentListCheckboxes()`? This triggers the following eslint errors for me:

``` shell
Compiled with warnings.
./src/reducers/org.js
  Line 811:18:  Expected to return a value at the end of arrow function  array-callback-return  Line 812:7:   Expected a default case                                  default-case
```

When adding the `default` case, again, the tests stay green.

It's such a minor issue compared to the huge contribution you're making, so I feel a bit bad about nitpicking. I'm only doing it, because those would be the first eslint warnings in the codebase^^

In any case, a **huge** thank you is in order 🙏🙇

necto:
I've removed the `default` case following the "avoid the catch-all cases" advice that goes along the lines (I couldn't easily find an online incarnation of this):

A catch-all case for a sum type often shelters a potential future bug. Any time a sum type is extended the developer needs to think of all the places it is dispatched upon. A catch-all default case might make it harder for the developer because it implicitly handles the new value of the type, but it was written when this new value didn't exist so the handling might be inadequate.

This argument applies better to statically typed languages where the match cases are checked at compile-time for exhaustiveness. However, it is valid for a dynamic language as well, except that the exception will arise in run-time, and might hurt user experience in production.

Also, frankly, I removed it because I couldn't cover it with a legit testing scenario so I considered it to be dead code.

Considering all the above, I think the best solution is to keep the `default` case as it is and throw an exception in the dev build. What would be the best way to add something like `if (not production) { assert(false); }`?

I am also fine with just keeping it as it was, as the argument above sounds weak for JS.

munen:
Thank you for taking the time to explain the rationale, and I understand it now. Having said that, I find it very interesting that there's obviously two diverging best practices on a binary decision.

I'm usually only programming in languages that are dynamically typed (mostly Clojure, Ruby and JavaScript). Those languages don't have an explicit contract. Instead, they use implicit contracts with the benefits of writing less tests and potentially having more re-usable code. The downside is that the implicit contract cannot be checked for during compile time. That's why 'defensive programming' is quite common in cases like `case` (sorry). 

There's an argument to be made if dynamically typed languages are harder to develop for in a bigger or disjunct team (which is typical for FOSS). Two of the possible remedies for this are 'tests' and 'static code analysis'. In organice, we try to make good use of both. JavaScript, by now, has great tooling for that. We're using  [eslint](https://eslint.org/) with the [eslint:recommended](https://github.com/200ok-ch/organice/blob/master/.eslintrc.yml#L5) settings. This will not only enforce a syntactical style guide (which is already nice, but more of a nice to have), but it will help not make common pitfalls. The pitfall for leaving out the `default` statement here is two fold:

1. There's no default behavior defined. Yet, for this specific instance it's quite reasonable to assume that a `childCompletionState` is `false` unless it follows the explicit expectation of being `checked`. Adding any other string (even by accident) would result in undefined behavior. And typos happen a lot more often in dynamic languages than in statically typed ones(;

2. There's no return value of the function on a fall-through. The rationale is the same as above.

Thankfully, these issues are trivially solved with static code analysis. For example, I call `eslint` in Emacs and will see the error directly when coding:

https://screenshots.200ok.ch/screenshot_2020_08_26-6ce69aba.png

Also, when starting `yarn start`, you'll see this:

https://screenshots.200ok.ch/screenshot_2020_08_26-6194194d.png

It didn't trigger an error on CI, even though we're running eslint on CI, because it's just a warning, not an error. Yet, it's the first warning on the codebase^^

I also understand you're argument that it's not trivial to find a good test case for it. The reason is that this defensive programming mechanism guards against a 'random' error like a typo or introducing a new specific state (which is encoded in a string, not in a static type). That's not really easy to test for. This closes the loop - in the beginning, I said that there's two remedies for the downside of dynamically typed languages. The first is tests (which you've obviously mastered! :rocket:). The other is to have code guidelines to save us from the other kind of 'random' errors - for example by programming defensively.

You can find the 'official' argument on why a `default` makes sense for a switch here: https://eslint.org/docs/rules/default-case

munen:
In conclusion, I'm super stoked about this PR! It's a hugely beneficial addition to organice. Thank you very much 🙏 🙇

munen:
I forgot to address

> To conclude, I think the best solution is to keep the default case as it is and throw an exception in the dev build. 

That's an interesting proposition.

> What would be the best way to add something like if (not production) { assert(false); }?

That would be `process.env.NODE_ENV === 'production'`. This is a [webpack feature](https://webpack.js.org/guides/production/).

munen:
I'll merge this PR to get the new tests in asap.

The option points "how to deal with closing" and the discussion of assertations can be added in future PRs.

Thank you, again!

schoettl:
>     * src/lib/timestamps.js: add a `timestampForDate` function to enable tests involving timestamps

Just now, I need your new function :) Good that I remembered that I read in in your PR.

PR #432, opened 2020-08-21
Make REMOVE_TABLE_ROW/COLUMN, UPDATE_TABLE_CELL_VALUE undoable

### LINKED ISSUE #430: Removing a table row or column should be undoable
**Describe the bug**
When I remove the column or a row of a table and then press undo, organice jumps to before the action that preceded the removal and does not restore the removed column.

**To Reproduce**
1. Sign in to a synchronization backend in organice service to enable the change-history functionality
2. Open an org file with a table and select a cell in this table
3. Click on "remove column"
4. Click on the "undo" button on the top of the screen
5. Observe that the column is not restored.

**Expected behavior**
The undo button in this context should restore the removed column and do nothing else.

**Additional context**
Discovered during the test development. Seems to me that a fix is to simply add the `REMOVE_TABLE_COLUMN` and `REMOVE_TABLE_ROW` into the `UNDOABLE_ACTIONS` list in `src/reducers/index.js`.
If it is that simple, I can do it myself by the end of the next week.

necto:
This renders the table editing and undoing experience more intuitive.

This closes #430


munen:
Good catch! Thank you for the fix :pray: 

I'll happily merge it - after I added the contribution to the changelog: 6ad7f1d

PR #434, opened 2020-08-24
transpose quality and quantity

chee:
i think it's meant to be this way round?


schoettl:
You're right, @chee ! Thanks for your contribution!

munen:
Whoops^^

PR #437, opened 2020-08-26
Fix external links

### LINKED ISSUE #435: External links in Organice settings open the page in the same window
**Describe the bug**
External links to help pages in Organice settings are not `target=_blank`. At least when the app is opened in the browser, clicking them will open the page in the same browser window.

In contrast links in the open document are, e.g.
```
<a href="nnnnn" target="_blank" rel="noopener noreferrer">nnnnn</a>
```

**Expected behavior**
All links to external resources should behave the same way.

I can fix this. But I first want to see if someone is expert on links :)

`target="_blank" rel="noopener noreferrer"` makes sense also for all http(s) links to external pages, right? Is there a better way in React to code links than to just write all this `<a href...>`? I think we have a `<Link>` element or so... I will look how it's done in the document...

schoettl:
Closes #435 

Wouldn't it be a good idea to write a component for this? `<ExternalLink url=url>` or so?

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -34,7 +34,7 @@
 
         If you're interested in inspecting the source code yourself (or even deploying your own
         version of the project), you can find the project's repo here:
-        <a href="https://github.com/200ok-ch/organice">https://github.com/200ok-ch/organice</a>
+        <a href="https://github.com/200ok-ch/organice" target="_blank" rel="noopener noreferrer">https://github.com/200ok-ch/organice</a>
       </div>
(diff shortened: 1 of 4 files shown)


munen:
@schoettl Good catch on those links. I found one that was broken and fixed, it too.

Happy to pul lthe changes in!

> Wouldn't it be a good idea to write a component for this? <ExternalLink url=url> or so?

Yes, that would make sense and would clean up the code quite a bit.

PR #440, opened 2020-08-29
Fix subheader fold propagation

necto:
Addressing the bug found in #431:

If a user folds a header, all its subheaders should collapse as well, so that when the user reopens it, they stay closed.

The previous behavior is buggy in a way that it keeps the subheaders open as they were, restoring their openness when the header is unfolded.

The previous behavior is useful at least for two users of organice (@munen and myself), so this change set also introduces a user setting.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,6 +4,13 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-08-29 Sat]
+** Fixed
+   - If a user folds a header, all its subheaders should collapse as well, so that when the user reopens it, they stay closed.
+     - The previous behavior is buggy in a way that it keeps the subheaders open as they were, restoring their openness when the header is unfolded.
+     - The previous behavior is useful, though. So this change introduces a user setting to toggle the behaviour.
+   - Thank you [[https://github.com/necto][necto]] for your [[https://github.com/200ok-ch/organice/pull/440][PR]] 🙏
+
 * [2020-08-25 Tue]
 ** Fixed
    - Updating table cell values and removing table rows or colums was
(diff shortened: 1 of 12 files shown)


necto:
Need help with this flaky test:
"Render all views › Org Functionality › Renders everything starting from an Org file › Agenda › Clicking the Timestamp in a TODO within the agenda toggles from the date to the time".
It has failed on my machine even on the current master. However when I started debugging it, in particular, when I inserted a bunch of prettyDOM logs, and

```
console.log(queryByText(/12 months ago/));
```

right before the failing line it succeeded and was succeeding ever since (on my machine).

Any idea what might be happening here?

munen:
> Need help with this flaky test:

Curious that you're saying it's flaky. That's good to know and I didn't so far. I'm not aware that this test ever failed on me or on CI.

But if it fails for you and then it works when you're debugging, that sounds like a timing issue to me. The test does open the agenda, which when run in the browser, does have an animation, at least. However, if that test fails for you sometimes, then other tests opening the agenda or similar modals (like search or the todo list) should also fail. Is that the case? Or is this the only test in the 'sometimes failing/flaky' category?

munen:
Iirc this is the first time that this test fails on CI. But since it does, I'll rerun the build. Not to say that all is good if it doesn't fail, but to see if it's a 'sometimes' failing test for CI.

munen:
NB: The test is green on my machine - on `master` and this feature branch.

munen:
Ok, `rebuild` is green. So it seems to be flaky, indeed.

munen:
Apart from the pre-existing flaky test, the new feature, implementation and tests look good to me!

I've added a bit of minor doc regarding what's the default in Org mode (503226f) and added your contribution to the changelog (dc557bc).

Since the test is green for me and on CI on a second try, and I've manually confirmed functionality of the agenda, I will merge this PR. I'll open an issue where we can discuss the options to improve the test reliability.

Great contribution as always, thank you for your continued effort, @necto :pray: :bow:

PR #444, opened 2020-08-30
Auto-focus filter input fields

### LINKED ISSUE #404: Autofocus on input for 'search' and 'refile'
As a user, when I open the 'search' or 'refile' modals on the desktop or mobile, I want the focus to be in the input field, so that I can start typing immediately.

schoettl:
Closes #404 

I thought, I already had an [`autofocus` attribute](https://developer.mozilla.org/de/docs/Web/HTML/Element/Input) on the input fields once, but they have no effect. In React the attribute must be `autoFocus`: https://blog.danieljohnson.io/react-ref-autofocus/


munen:
Good job, great feature! :+1: :pray:

PR #453, opened 2020-08-31
Replace some <a> tags with <ExternalLink>

schoettl:
The first examples looked like this interface is good: `<ExternalLink href="url" content="text" />`

But later, I stumpled upon links with HTML tags inside, so it's probably better to allow children as described here: https://reactjs.org/docs/composition-vs-inheritance.html

Any reason to keep the `content` property and only use its value instead of childen if no children? Maybe it's a little bit shorter? But I think it's cleaner to have only one "interface".

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,9 +50,9 @@
     "eslint-nibble": "^5.1.0",
     "eslint-plugin-jest": "^23.20.0",
     "eslint-plugin-react": "^7.20.6",
-    "eslint-plugin-react-redux": "^3.1.0",
-    "prettier": "^2.0.5",
+    "eslint-plugin-react-redux": "^3.3.0",
     "prettier-eslint": "^11.0.0",
+    "prettier": "^2.0.5",
     "prettier-eslint-cli": "^5.0.0",
     "raw.macro": "0.4.2",
     "react-test-renderer": "^16.13.1"
--- a/package.json
+++ b/package.json
@@ -50,9 +50,9 @@
     "eslint-nibble": "^5.1.0",
     "eslint-plugin-jest": "^23.20.0",
     "eslint-plugin-react": "^7.20.6",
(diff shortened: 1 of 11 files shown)


munen:
> Any reason to keep the content property and only use its value instead of childen if no children? Maybe it's a little bit shorter? But I think it's cleaner to have only one "interface".

I'd prefer to have only the children interface. It feels more like a 'normal' HTML tag.

munen (changes requested):
> Any reason to keep the content property and only use its value instead of childen if no children? Maybe it's a little bit shorter? But I think it's cleaner to have only one "interface".

I'd prefer to have only the children interface. It feels more like a 'normal' HTML tag.

That is, if it is OK for you. Otherwise, I don't mind to have the `content` property, as well. Would be less refactoring, so that's an upside^^

schoettl:
OK, ready for review and merge.

munen:
@schoettl Code looks good and manual tests also worked fine on the links.

Before merging, I wanted to run `yarn test` locally once and saw this:

image

It's reproducible and I don't have it on `master`. Unfortunately, I also have no first hunch on what the error entails.

schoettl:
Thanks, I didn't see this warning in the first run. `key` is a reserved property, I didn't knew that.

@munen Before we merge, I'd like to fix a dead link on the landing page (when not signed in). Where is the Documentation button intended to link to? Currently it is `/documentation.html` which doesn't exist and is also not a nice URL ;)

schoettl:
Ah, got it. It must be the full URL because the link would be dead if organice is hosted elsewhere.

munen:
@schoettl Looks and works great, happy to merge. Thank you for the additional changes! :pray: 

Also good to know that `key` is a reserved keyword! :smile:

Issue #134, opened 2019-12-10
PWA implementation is broken


dotcs:
**Describe the bug**
Organice's implemenation of PWA is broken. It is currently not installable on Android (and likely on iOS) and on Linux operating systems.

**To Reproduce**
1. Go to https://organice.200ok.ch/
2. Check your browser's status bar and see if the "Install organice" button appears.

**Expected behavior**
The install button should appear, but it is missing. It should look like this: 
![image](https://user-images.githubusercontent.com/3976183/70540310-4c103980-1b65-11ea-970b-6b25be6ab69b.png)

**Desktop (please complete the following information):**
 - OS: `Linux mali 4.19.84-1-MANJARO #1 SMP PREEMPT Wed Nov 13 00:07:37 UTC 2019 x86_64 GNU/Linux`
 - Browser: Firefox, Chrome (latest versions)

**Smartphone (please complete the following information):**
 - Device: Pixel 2
 - OS: Android
 - Browser: Chrome, Brave

**Additional context**
I have found the root cause for this. It seems that PWA's require to have two icons with a) 192x192 and b) 512x512 pixel sizes in order to a) installable and b) provide a splash screen while loading. Both icons are currently not available.

To check open `Chromium -> Developer Tools -> Audit` and run an audit on Progressive Web Apps. This should look like this:

![image](https://user-images.githubusercontent.com/3976183/70540733-10c23a80-1b66-11ea-9f16-079b3e71cd84.png)

The audit then lists the findings as follows:

![organice-pwa-issue](https://user-images.githubusercontent.com/3976183/70540005-c3919900-1b64-11ea-83a1-fd3742b79a7b.png)

I am working on a solution for this bug.

munen:
Great bug report!

Looks like Android and Firefox are quite finicky with respect to PWAs. We just finished a ticket where the PWA functionality didn't work on those: https://github.com/200ok-ch/organice/issues/89

Awesome that you described how you debugged the problem! I was not aware that Chromium has this functionality. As an iOS user, for the last issue I was relying on feedback from Android users.

The [logo](https://github.com/200ok-ch/organice/blob/master/public/organice.png) is available in 1863 x 1861, so providing a logo in at least 512px and 192px should be feasible(;

Thank you for reporting and taking care of the issue! :pray: :bowing_man:

dotcs:
@munen Thanks for the kind words. I appreciate it.

These audit features are absolutely great! Glad that I could show you that those exist. :+1: 

[...]

Issue #52, opened 2019-09-23
Have organice show up as a share target on Android


branch14:
...preferably while also using a Capture Template.

Here is a blog post that explains Android share targets in WPAs in some detail:

https://chodounsky.net/2019/03/24/progressive-web-application-as-a-share-option-in-android/

munen:
@branch14 That sounds like a great thing! I did it for iOS (using the shortcuts app). Since I don't have an Android phone and cannot test, I'd be the wrong person to do this. Would you be interested in taking up this issue?

branch14:
Sure... given that I find time to do so. :)

munen:
Sweet!

So I’ll assign it to you and add a low prio tag. It’s mostly up to you when/if this comes(;

dotcs:
@munen I have tried to add support for webshare. Please see #138 for the current implementation.

In principle webshare does work. Sharing the GitHub repository to organice does currently end up in the following page:

![image](https://user-images.githubusercontent.com/3976183/70982529-6b541d00-20b7-11ea-99cb-40a6ac9236a8.png)

I wanted to ask how I can proceed here. I do have a proposal and would kindly ask you to think this through:

I think we could store the information from the webshare dialog in memory and give the user an UI element to add the content to a org document. The logic could be create a new sibling with the content from the webshare memory. We could then use the `renderShare()` function to just copy the content to the memory. It would serve as a middleware to copy the content to memory. After that we just redirect the user to the normal workflow, which would for example open the last file in which the user was active. If the webshare memory is not empty we should show an UI element that adds a sibling to the currently selected item (or if none is selected place a new item at the bottom).

If available I would use `title` as the headline, `text` and `link` as the description. If title is not available I would use `text` or `link` as the title or we could fallback to a default title such as "Shared element" or so.

Since you know organice best I would kindly ask for help. Can you give me some hints how you would do it and where to continue from here? Thanks in advance!

munen:
Hi @dotcs 

Thank you for picking up this issue! @branch14 is very happy^^

As for the UX on how to proceed, I see two valid options:

1. Since it duplicates functionality of bookmarklets, do it analogous to how bookmarklets work. (Of course the UX of native share is much better than the manual integration of bookmarklets!)
2. Open a capture dialog analogous to a capture template

Here's doc on how bookmarklets work atm:

- Prose and code example: https://github.com/200ok-ch/organice#bookmarklets
[...]

PR #463, opened 2020-09-06
fix: Get rid of `start_url` in manifest.json

munen:
This prevents users to bookmark specific files or URLs.

Downside: It looks like the borg set it as required field, though: https://web.dev/add-manifest/ Testing from an Android or Google Chrome user is required here.

On iOS, using organice as PWA works without this attribute.

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -18,7 +18,6 @@
       "type": "image/x-icon"
     }
   ],
-  "start_url": "./index.html",
   "display": "standalone",
   "theme_color": "#000000",
   "background_color": "#ffffff"


munen:
@schoettl Can you test if this change still allows on Android to use organice as PWA, please?

I can't do it, because I do not have an Android and because Google decided that it doesn't want to support their proprietary lighthouse tooling on Debian https://github.com/200ok-ch/organice/issues/134.

No time pressure on the review. I manually changed the manifest.json temporarily for me to get a proper PWA bookmark for the moment.

munen:
If Google won't allow PWAs without the `start_url`, then I'm afraid the only way to go forward properly will be to:

- Implement a backend to dynamically generate the manifest.json. This looks like a super bad choice for just a whim of Google. It'll break one of the big selling propositions of organice to be just a front-end app and it'll break all existing installations.
- Implement a feature in the settings to set a 'default' file that should be loaded instead of showing the root URL. That'll be easier than the back-end, but still feels like weird UX.

aspiers:
[**@munen**](https://github.com/munen) commented on [September 6, 2020 1:16 PM](https://github.com/200ok-ch/organice/pull/463#issuecomment-687774160 "2020-09-06T12:16:09Z - Replied by Github Reply Comments"):
> If Google won't allow PWAs without the `start_url`,

My memory is that it does, but I could easily be wrong.  Whenever I find time to get back to organice hacking, I will try to remember to test this if @schoettl or someone else doesn't beat me to it.

> then I'm afraid the only way to go forward properly will be to:
> 
> *   Implement a backend to dynamically generate the manifest.json. This looks like a super bad choice for just a whim of Google. It'll break one of the big selling propositions of organice to be just a front-end app and it'll break all existing installations.

Why would a backend be needed in order to dynamically generate it?  I agree - this would seem like a very odd way forward to me.

> *   Implement a feature in the settings to set a 'default' file that should be loaded instead of showing the root URL. That'll be easier than the back-end, but still feels like weird UX.

That doesn't sound weird to me, it sounds great ;-)  There could be a default value corresponding to what it is now, with an option to change it.  However it does raise [the same questions I previously raised regarding how / when a dynamically generated manifest would get updated](https://github.com/200ok-ch/organice/issues/52#issuecomment-657092025).  But as you can see from that link, now we have at least two good reasons to figure this out ;-)

aspiers:
> Google decided that it doesn't want to support their proprietary lighthouse tooling on Debian [GoogleChrome/lighthouse#7246](https://github.com/GoogleChrome/lighthouse/issues/7246).

This is a bit of a tangent, but I read that issue and [the corresponding Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=922431) and made the following observations:

- Lighthouse is open source ([Apache 2.0](https://github.com/GoogleChrome/lighthouse/blob/master/LICENSE)) not proprietary.
- The cause of the problem seems to be two-fold:
      a) It uses some minified Javascript files, the source for which has not yet been located by any of the involved parties (but I suspect this is simply a case of a lack of understanding and/or a disconnect in communication), and
      b) it seems that [the Lighthouse devs aren't sufficiently motivated to help the Debian people solve this problem](https://github.com/GoogleChrome/lighthouse/issues/7246#issuecomment-547463699).  (Of course, they are under no obligation to do so, so this isn't a criticism, although it would have been great if they were.)

Nevertheless, I would have thought that downloading Chrome for Debian [directly from Google](https://www.google.com/intl/en_uk/chrome/) should easily overcome any such issue with the version of Chromium natively packaged by Debian.

munen:
> Why would a backend be needed in order to dynamically generate it?

How else could we do it? Even if we added a route to the top, we can use React only to render within the `root` component:

image

Here's the diff of this quick demo:

image

I'm not aware that it's possible to render a static file within the browser. In the end, that file would have to be rendered by JavaScript running within the browser which would have to reside somewhere in the DOM.

munen:
> That doesn't sound weird to me, it sounds great ;-)

Well, at least there's some positive thoughts on what I consider to be a backward step for iOS*lol* Thanks for the positive thougt!

aspiers:
I think we might have just been misunderstanding each other :laughing:  At a glance, the diff you just posted looks like a nice solution, so I'm a little confused now.  What's the problem with doing it that way?

aspiers:
But even if there is a problem, maybe one of the approaches I linked to [above](https://github.com/200ok-ch/organice/pull/463#issuecomment-687786919) could work, e.g.

- https://medium.com/@rajathans/dynamic-pwa-manifest-using-javascript-and-rest-api-a23d132de7d6

aspiers:
Ah no, that one *does* require a backend REST API.  But maybe this one works:

https://medium.com/@alshakero/how-to-setup-your-web-app-manifest-dynamically-using-javascript-f7fbee899a61

Here is the trick - to create an object URL:

```js
const stringManifest = JSON.stringify(myDynamicManifest);
const blob = new Blob([stringManifest], {type: 'application/json'});
const manifestURL = URL.createObjectURL(blob);
document.querySelector('#my-manifest-placeholder').setAttribute('href', manifestURL);
```

munen:
> I think we might have just been misunderstanding each other 😆 At a glance, the diff you just posted looks like a nice solution, so I'm a little confused now. What's the problem with doing it that way?

The problem is that the manifest.json file is a static file. Static files cannot be generated in the fron-tend. As soon as you have a front-end, you no longer have a static file (hen and egg problem). The implementation that I've shown shows exactly that. We're not getting a static manifest.json file, but a rendition within the DOM (HTML). Stuff within HTML != static file.

munen:
> Ah no, that one does require a backend REST API. 

Yes, with a back-end, you can generate static files. As was my initial premise(;

> But maybe this one works:

> Here is the trick - to create an object URL:

This one looks interesting. Tbh, I didn't know about `createObjectURL`. Otoh, this might be the reason: https://caniuse.com/?search=createObjectURL

aspiers:
[**@munen**](https://github.com/munen) commented on [September 6, 2020 7:37 PM](https://github.com/200ok-ch/organice/pull/463#issuecomment-687865543 "2020-09-06T18:37:45Z - Replied by Github Reply Comments"):
> This one looks interesting. Tbh, I didn't know about `createObjectURL`. Otoh, this might be the reason: [caniuse.com/?search=createObjectURL](https://caniuse.com/?search=createObjectURL)

No, I think that's specifically about whether `createObjectURL` accepts `MediaStream` objects (since that was deprecated), not about support for `createObjectURL` in general.  See https://www.fxsitecompat.dev/en-CA/docs/2018/url-createobjecturl-no-longer-accepts-mediastream-as-argument/ which says:

> According to the current specs, only a `Blob` or `MediaSource` object can be accepted.

IIUC a `Blob` is what we would need, in which case https://caniuse.com/bloburls is presumably the relevant page.  But this is all new to me so ICBW as usual ;-)

> It's worth a try, though. On MDN, the compatibility table looks quite different: [developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL)

Indeed.  Note the second row there is "No longer accepts MediaStream object" which corresponds to the link you shared above.

aspiers:
[**@munen**](https://github.com/munen) commented on [September 6, 2020 7:36 PM](https://github.com/200ok-ch/organice/pull/463#issuecomment-687865197 "2020-09-06T18:36:23Z - Replied by Github Reply Comments"):
> The problem is that the manifest.json file is a static file. Static files cannot be generated in the fron-tend. As soon as you have a front-end, you no longer have a static file (hen and egg problem). The implementation that I've shown shows exactly that. We're not getting a static manifest.json file, but a rendition within the DOM (HTML). Stuff within HTML != static file.

But why is that a problem?  Why do we need a static file?

munen:
> But why is that a problem? Why do we need a static file?

Because that’s the spec on how to describe a PWA (link in desc above).

munen:
@aspiers Sounds great :pray: 

I have implemented this and tested it successfully on an iPhone. This PR is obsolete now and superseded by #465

PR #467, opened 2020-09-09
[WIP] Change interface of external link: allow children

schoettl:
What a mess :( I forgot that this refactoring was already merged. I added some commits, rebased on master and screwed it up.


schoettl:
See #453

PR #477, opened 2020-09-20
Recognize US and Swiss phone numbers

munen:
Next to 'canonical' phone numbers like +498025123456789, also recognize:

US phone numbers:
  - 123-456-7890
  - (123) 456-7890
  - 123 456 7890
  - 123.456.7890
  - +91 (123) 456-7890

Swiss phone numbers:
  - 0783268674
  - 041783268675
  - 0041783268674
  - +41783268676
  - +41783268677

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,6 +4,24 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-09-20 Sun]
+** Added
+
+   - Recognize US and Swiss phone numbers and make them clickable
+   - Next to 'canonical' phone numbers like +498025123456789, also recognize:
+     - US phone numbers:
+       - 123-456-7890
+       - (123) 456-7890
+       - 123 456 7890
+       - 123.456.7890
+       - +91 (123) 456-7890
+     - Swiss phone numbers:
+      - 0783268674
+      - 041783268675
+      - 0041783268674
+      - +41783268676
+      - +41783268677
+
 * [2020-09-07 Mon]
 ** Added
    - Added the capability to chose the start screen when installing organice to the homescreen.
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -446,7 +446,7 @@ describe('Render all views', () => {
       });
 
       describe('Link recognition', () => {
-        test('recognizes phone numbers', () => {
+        test('recognizes canonical format +xxxxxxxxx phone numbers', () => {
           fireEvent.click(
             queryByText('A header with a URL, mail address and phone number as content')
           );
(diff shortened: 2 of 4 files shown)


schoettl:
I hope this doesn't highlight too many false-positives. I recently installed (and uninstalled) the 3cx browser add-on to make phone numbers click-able and it marked more strings than real phone numbers... e.g. invoice numbers. We will see ;)

munen:
> I hope this doesn't highlight too many false-positives. I recently installed (and uninstalled) the 3cx browser add-on to make phone numbers click-able and it marked more strings than real phone numbers... e.g. invoice numbers. We will see ;)

I guess that's unavoidable. Even on the OS from the richest company on earth, we're getting numbers highlighted as phone numbers that are not phone numbers.

I'd recon that they tested this hypothesis, but for me it's way easier to not click an invoice number and call it (because I know what it is), than to 'edit' a header, fiddle around and find a phone number like 0783268674 (a regular Swiss mobile number), copy it and then figure out how to paste that to the phone app. I've seen a user (today) use two phones, because they couldn't do it. Hence, then new feature.

munen:
Btw, our German test number is either not a valid number (so you’re argument is already in play before this PR) or iOS has a worse algorithm than organice^^

In any case, I’m quite happy to see that we’re recognizing mobile numbers where iOS is not(;

7D92CED5-CF45-453F-BBB1-27D88E952B82

schoettl:
Alright, GitHub doesn't recognize canonical phone numbers when followed be a comma ^^

If a user often use phone numbers it's certainly a good thing. I just fear that there are dozens of phone numbers highlighted where actually no phone numbers are. But we'll see.

And if it really annoys me too much, I'll make it a setting :)

munen:
It’s not the comma, though^^

FCFBA965-74B7-4C38-8855-CA0879D241A1

munen:
I'll replace the number we had with +49123456789 which conforms to a German landline according to [wikipedia](https://en.wikipedia.org/wiki/Telephone_numbers_in_Germany).

munen:
Yes, that one also get's recognized by iOS/Github.

schoettl:
> I'll replace the number we had with +49123456789 which conforms to a German landline according to [wikipedia](https://en.wikipedia.org/wiki/Telephone_numbers_in_Germany).

Yeah, maybe it was just too long. But the zero in +498025xxxx is correct.

munen:
> But the zero in +498025xxxx is correct.

As in "There is a zero in the prefix for Miesbach."? https://www.telefonvorwahlen.net/de/prefix/08025

schoettl:
> > But the zero in +498025xxxx is correct.
> 
> As in "There is a zero in the prefix for Miesbach."? https://www.telefonvorwahlen.net/de/prefix/08025

Right, a German Ortsvorwahl (= prefix) often contain zero(s) and often start with zero. Leading zeros are omitted in the canonical form though.

munen:
> Right, a German Ortsvorwahl (= prefix) often contain zero(s) and often start with zero. Leading zeros are omitted in the canonical form though.

Ack. +4980251234 is still recognized.

PR #478, opened 2020-09-20
fix: Tighten phone number recognition

munen:


munen:
@schoettl I've taken you're input to heart and have severely tightened the phone number recognition.

schoettl:
Thanks :)

munen:
Sure thing, good call on calling me out on matching too much(;

It's still possible to match non-phone-numbers, but the likelyhood has gone down drastically, at least.

PR #479, opened 2020-09-20
refactor: Markup parsing via RegExp becomes more readable

munen:


munen:
@schoettl I wanted to write a small docstring to the RegExps for phone numbers. For that, I've finally split up the big RegExp into chunks that can each get a docstring.

We discussed this option at some point, but discarded it. I think it was in favor of putting the energy into `org-parser`. That's valid, but I've found a rather easy and non-breaking way (thank goodness for the many tests that we have by now^^). Do you have objections or comments on going this way? 

Of course, this PR is just meant as a small improvement on the readability. The general goal to move to `org-parser` at some point stands. But even for that, I think it helps if we have chunks of RegExps that can be read/tested/documented singularly, so that we can easily see what kind of stuff organice parses already (apart from looking at the tests, which might or might not include everything).

schoettl:
That looks good! Simple but effective refactoring.

munen:
> That looks good! Simple but effective refactoring.

Sweet, thank you very much! :pray:

Issue #494, opened 2020-10-04
As a user I want to see summaries of time logged on tasks


tarnung:
[Pull request](https://github.com/200ok-ch/organice/pull/497)

The timetracking in LOGBOOK does not provide an easy overview and therefore gets unmanagable quickly.
org mode has:

>C-c C-x C-d (org-clock-display)
>
>Display time summaries for each subtree in the current buffer. 
>This puts overlays at the end of each headline, showing the total time recorded under that heading, including the time of any subheadings. 
>You can use visibility cycling to study the tree, but the overlays disappear when you change the buffer (see variable org-remove-highlights-with-change) or press C-c C-c.

It would be nice to have something similar in organice.

I built a first draft in [a fork](https://github.com/tarnung/organice/tree/feat/org-clock-display).

Questions to be answered:
- Do you want this feature in organice?
- How would the display be activated? 
I added a toggle button to the header bar which is ...debatable.
- How would the display be shown? 
I just added " LOGGED 3:15 " to the end of the header line since I don't have strong frontend skills or sensibilities.
- How could the display be deactivated? 
I have not yet implemented the "the overlays disappear when you change the buffer" part of the org mode feature description. So currently the display is not updated (and does not disappear) on change. It needs to be untoggeled and retoggeled to update it.

schoettl:
I haven't looked at your fork yet, but...

>     * Do you want this feature in organice?

I'd like that feature.

> 
>     * How would the display be activated?
>       I added a toggle button to the header bar which is ...debatable.

I think, currently we have no better place. We could hide it under a new menu icon in the top bar for less important actions.


>     * How would the display be shown?
>       I just added " LOGGED 3:15 " to the end of the header line since I don't have strong frontend skills or sensibilities.

I think, right-aligned in the header title line (like in emacs org-mode) would be a good place.

>     * How could the display be deactivated?
>       I have not yet implemented the "the overlays disappear when you change the buffer" part of the org mode feature description. So currently the display is not updated (and does not disappear) on change. It needs to be untoggeled and retoggeled to update it.

As long as it doesn't update automatically, it could also be hidden automatically after some seconds.

But it shouldn't be too hard to make it update automatically.

tarnung:
> I think, right-aligned in the header title line (like in emacs org-mode) would be a good place.

I placed it there. The caveat about my frontend skills still applies.

> As long as it doesn't update automatically, it could also be hidden automatically after some seconds.
> 
[...]

PR #497, opened 2020-10-07
Display time summaries for each subtree

tarnung:
[See Issue 494](https://github.com/200ok-ch/organice/issues/494)

I'm willing to discuss things further in case there are shady things going on in my code,

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,6 +4,17 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-10-10 Sat]
+** Changed
+
+   - Improve header stickiness
+     - The header bar was not always sticky for longer documents and the therefore some functionality was hard to reach (the user had to scroll to the top to reach it).
+     - Thank you [[https://github.com/tarnung][tarnung]] for your [[https://github.com/200ok-ch/organice/pull/499][PR]] 🙏
+* [2020-10-07 Wed]
+** Added
+
+   - Added Documentation on [[https://orgmode.org/manual/Repeated-tasks.html][repeaters]] and [[https://orgmode.org/manual/Deadlines-and-scheduling.html#Deadlines-and-scheduling][delays]] to =sample.org= and to mouseovers for the relevant buttons.
+
 * [2020-09-20 Sun]
 ** Added
 
--- a/changelog.org
+++ b/changelog.org
@@ -5,11 +5,16 @@ All user visible changes to organice will be documented in this file.
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
 * [2020-10-10 Sat]
+** Added
+   - Display time summaries for clock entries in each subtree
+     - Thank you [[https://github.com/tarnung][tarnung]] for your [[https://github.com/200ok-ch/organice/pull/497][PR]] 🙏
+
 ** Changed
 
    - Improve header stickiness
      - The header bar was not always sticky for longer documents and the therefore some functionality was hard to reach (the user had to scroll to the top to reach it).
      - Thank you [[https://github.com/tarnung][tarnung]] for your [[https://github.com/200ok-ch/organice/pull/499][PR]] 🙏
+
 * [2020-10-07 Wed]
 ** Added
 
--- a/sample.org
+++ b/sample.org
@@ -9,6 +9,9 @@ This header has some description text and a couple subheaders. Look through the 
 * Actions
 ** Editing headers
 When you select a header the "header action drawer" appears. The first two icons in this drawer edit the header and description respectively. Try editing this header.
+
+If you want the "header action drawer" to disappear or to deselect the current header, tap into the title (or empty space) in the headerbar.
+
 ** All icons in the header action bar
 
 1. Pen: Edit header title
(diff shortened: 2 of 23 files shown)


tarnung:
UX / style is only tested on desktop.

Also it looks like I broke a test. I'll look into that.

tarnung:
I'll look into fixing the tests when I got some feedback for the current implementation.

munen:
@tarnung From a first read through the code, I like the implementation and what it looks like! It looks actually quite a lot better than in Emacs too be honest - everything is aligned nicely and auto-updates(; I also confirm that it works well on iOS.

Generally speaking, this PR will be a great joy to merge! :pray:

munen:
I have fixed three minor issues (eslint warning and two tests) and have deployed to staging for additional tests.

The good news is that where it works, it works really well. The design also works in PWA mode on iOS, for example. Good job on the flexbox layout!

There is one thing that I don't understand immediately, though. organice shows different times in the `org-clock-display` than Emacs. Here's a comparison on the same file:

Emacs:

image

organice

signal-2020-10-10-180651

I understand why Emacs is not showing times for all subheaders, even though there are clock entries. Emacs only shows clock entries within the `org-clock-display-default-range` which defaults to one year.

But reading through the PR, I don't find such a restriction. So, I have two questions:

1. Why aren't all clock entries used for displaying the time?
2. Would it make sense to add reading `org-clock-display-default-range` in organice, as well? If so, we have the options to add it as a setting or as a file local variable. The latter, we haven't implemented in Emacs, so far, but the syntax looks quite parsable: `# -*- org-clock-display-default-range: thismonth; -*-`.

tarnung:
It looks like your CLOCK entries are not inside of a :LOGBOOK: so they are probably not parsed as organice logBookEntries.

munen:
I have found the reason why not all clock entries are summed up in my personal Org files. Some headers don't use a `LOGBOOK`. I have written reducer tests to show the two scenarios (with and without `LOGBOOK`).

munen:
*hihi*

Thanks for getting back to me so fast. We've been writing in parallel.

Yes, they are not parsed as `logBookEntries`, but they also don't have to be. Org mode allows for clock entries to be within or without a `LOGBOOK`.

tarnung:
About your second point: I have to unveil the horrible truth - I'm not an emacs user.
I don't know about all the intricacies of org mode and I did not know about `org-clock-display-default-range`.
If you value consistency between organice and org mode we could implement it. I probably would not use the feature and I would not expect there to be a default cutoff.

munen:
Here's the relevant documentation in Emacs:

```
org-clock-into-drawer is a variable defined in ‘org-clock.el’.
Its value is t

  You can customize this variable.


This variable was introduced, or its default value was changed, in
version 26.1 of Emacs.

Documentation:
Non-nil when clocking info should be wrapped into a drawer.

When non-nil, clocking info will be inserted into the same drawer
as log notes (see variable ‘org-log-into-drawer’), if it exists,
or "LOGBOOK" otherwise.  If necessary, the drawer will be
created.

When an integer, the drawer is created only when the number of
clocking entries in an item reaches or exceeds this value.

When a string, it becomes the name of the drawer, ignoring the
log notes drawer altogether.

Do not check directly this variable in a Lisp program.  Call
function ‘org-clock-into-drawer’ instead.
```

munen:
> About your second point: I have to unveil the horrible truth - I'm not an emacs user.

That is actually amazing to hear this. So you're actually using organice, because you like it as a tool and not because you're into Emacs and want to use Org on the go? If so, I'm stoked! :partying_face: 

> If you value consistency between organice and org mode we could implement it. I probably would not use the feature and I would not expect there to be a default cutoff.

Generally, we go with consistency to Emacs. Of course, only within reasonable boundaries. Firstly, copying Org mode verbatim is a huge undertaking, so we are pragmatic with that. Secondly, not every feature makes sense on a mobile device or desktop browser that's not embedded in Emacs.

munen:
In this particular instance (regarding the `org-clock-display-default-range`), I'd opt to add it as a setting. The rationale is that I suspect (and kind of know from experience) that quite a few people use time tracking in Emacs for work and controlling. Hence, they will have to set certain boundaries (day, week, month, year) to get reasonable information.

Of course, different workflows can totally allow for an unrestricted summary of clock entries.

Having said that, this addition is quite easy to make and I'm happy to do it if you don't have a use-case for it. I don't want to block your PR because of it.

munen:
With regards to omitting clock entries that are not within a LOGBOOK drawer, I have just realized that organice only views those properly. On editing the timestamps, the time range will not change.

I might be a relic to still have files in use which use clock entries without a LOGBOOK drawer. I just checked my Emacs config history and I've explicitly asked Emacs not to log to a drawer some years ago, because I had some automation elisp programs (for billing customers) that didn't expect them. It's also been a long time that I've reverted to what is the default in Emacs which is to use a LOGBOOK drawer.

In conclusion: organice already only operates properly on clock entries within LOGBOOk drawers - and that's the default for Emacs since a while. So let's leave it at that. Apologies for making a fuzz about it :pray:

tarnung:
I was thinking more in the line of expanding the search functionality by searching for timeranges on CLOCK, SCHEDULED or DEADLINE and displaying a total time clocked over the found items. I didn't know emacs people did it by setting a variable :)

Well, I witnessed your 'playing emacs like an instrument' session live and it didn't convince me to switch editors. But I've been intrigued by org mode before and your presentation surely helped to keep it on my mind. I tried orgzly for a while but it didn't stick. A few weeks ago I decided to look around again for available tools that also support desktop use and I was surprised to see that my former lecturers are working on one.

munen:
> I was thinking more in the line of expanding the search functionality by searching for timeranges on CLOCK, SCHEDULED or DEADLINE and displaying a total time clocked over the found items. I didn't know emacs people did it by setting a variable :)

That's actually quite a reasonable way to go!

The thing is, in Emacs, there's always many ways to do things. Restricting/filtering the buffer on various contraints, that's what the [Org mode agenda views](https://orgmode.org/manual/Agenda-Views.html) are for. Those are mighty powerful. It's kind of a combination of using organice Agenda + Search. In organice, we've split them up, because non-Emacs users are not very used to thinking of an Agenda of a filtered down list of things, but more easily of "what is happing today/this week?" - and well, search is pretty intuitive, too. We might change that at some point in the future.

The other thing is, that I don't know Emacs users that use `org-clock-display` at all, so I cannot speak to them using the variable. To me, the feature is too basic and I suspect that it is a very old feature.

I suspect that many Emacs users would rather use the [Clock table](https://www.gnu.org/software/emacs/manual/html_node/org/The-clock-table.html) functionality. With clock reports, you can do pretty amazing stuff. For example, here I'm using it for internal controlling, but at the same time I'm using the same data to calculate a different table for automated time reporting to the customer:

image

Ok, so in conclusion on this issue: If you're interested in "expanding the search functionality by searching for timeranges on CLOCK, SCHEDULED or DEADLINE and displaying a total time clocked over the found items", then I'll forgo the `org-clock-display-default-range`, because it's very specific and your proposed changes are much more generic! Good thinking on your part, too!(;

> A few weeks ago I decided to look around again for available tools that also support desktop use and I was surprised to see that my former lecturers are working on one.

Ha! That's why I'm familiar with your name*g*

Thanks for checking out organice and starting to contribute and putting those Web1-3 skills to good use :pray: 

We're gaining more and more desktop users, so it's good to know that there's capable programmers in that pool as well. Personally, I'm using organice mostly on the phone - and by now, it's really nice to use (if I say so myself... well, whatever that's worth^^). On the desktop, it's quite a bit more clunky. However, it definitively is our goal to have organice be a proper web application and work well equally well on the desktop and on mobile. If you've got good ideas on how to improve UX on the desktop, we're all ears regarding ideas and PRs(;

munen:
I've updated the changelog with attribution to your contribution: d027807

I'm going to merge this PR now with the presumption that at some point there's another PR from you coming that expands the search functionality as discussed above. If your plans change, that's OK, of course. If possible, let me know then and I'll at least add a setting for `org-clock-display-default-range`.

By the way, if you want to discuss some changes in the future, you are very welcome to join the community chat: #organice on IRC Freenode, or [#organice:matrix.org](https://matrix.to/#/!DfVpGxoYxpbfAhuimY:matrix.org?via=matrix.org&via=ungleich.ch) on Matrix.

Thank you, again, for your great contribution. And big congratulations on getting a non-trivial feature proper right on the very first implementation! :pray: :bow:

tarnung:
> putting those Web1-3 skills to good use

At least I can blame phil if my code is subpar.

> If you've got good ideas on how to improve UX on the desktop, we're all ears regarding ideas and PRs(;

I'm working on something. I hope you don't mind settings inflation ;)

> I'm going to merge this PR now with the presumption that at some point there's another PR from you coming that expands the search functionality as discussed above.

I'm up for it. I don't see a clear best way to add it into the UI yet. But we can have this discussion in another issue/pr.

munen:
> At least I can blame phil if my code is subpar.

And Phil would happily listen to the feedback and become even stronger :muscle: (;

> I'm working on something. I hope you don't mind settings inflation ;)

Not at all! Emacs Org mode has over [650 config flags](https://orgmode.org/worg/org-tutorials/org-customize.html). We've got a long way to go(;

> I'm up for it. I don't see a clear best way to add it into the UI yet. But we can have this discussion in another issue/pr.

Sweet! 

With regards to the UI, it's probably a good idead to add @schoettl to the discussion. He's the initial contributor to the search functionality and also a maintainer of organice.

I've created a new issue here: https://github.com/200ok-ch/organice/issues/504 Let's continue the discussion there.

schoettl:
Hi again, great to see so much discussion and this new feature!

How do I activate the new time summary display? What have I missed? I don't see a button or a setting to show it.

munen:
> great to see so much discussion and this new feature!

👍 

> How do I activate the new time summary display? What have I missed? I don't see a button or a setting to show it.

In the settings, there's a new section "Display time summaries":

Screenshot 2020-10-11 at 12 18 38

PR #499, opened 2020-10-08
improve header stickiness

tarnung:
On Desktop the headerbar isn't sticky, since the scrollcontainer is the html-body. The whole #root-div itself scrolls inside of it, headerbar included. If the org-file is long enough I get an inner scrollbar for the #root-div and an outer scrollbar for the html-body.
On mobile I get an inconsistent experience. When I scroll to the top and back down, sometimes the headerbar stays, sometimes it scrolls away. Both scrollbars are present there too, but I don't know what determines which one is used.

This change fixes the issue on desktop browsers. I have not tried it on mobile yet but it works on browser-simulated mobile dimensions.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,6 +4,13 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-10-10 Sat]
+** Changed
+
+   - Improve header stickiness
+     - The header bar was not always sticky for longer documents and the therefore some functionality was hard to reach (the user had to scroll to the top to reach it).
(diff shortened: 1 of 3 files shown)


tarnung:
Known issues: layout breaks when clicking on a header behind floating action buttons

munen:
@tarnung Great changes! I've been wanting to tackle this, as well. Before, the functionality behind the header bar could easily be 'out of reach'.

Good catch, and nice implementation. I confirm that it works well on iOS, too.

Thank you and welcome to the contributors list :pray: :bow:

munen:
@tarnung I have updated the changelog with your contribution: https://github.com/200ok-ch/organice/pull/499/commits/8865e78203f26b68abbcc83efadfe8874e612b5b

PR #501, opened 2020-10-10
Make Docker containers configurable

maljub01:
This change makes it possible to configure the application when run using the official Docker image by setting the relevant environment variables or mounting a `.env` file.

It does this by delaying building assets until the container is started.

Without this change, it is not possible to use the official image with a Dropbox or Google Drive backend.

However, this change has two negative side-effects:
1) The image size jumps from 139MB to 819MB, because of the need to include the node dependencies in order to build at runtime.
2) The startup time is slightly slower.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,21 +8,19 @@ RUN sed -i 's/http\:\/\/dl-cdn.alpinelinux.org/http\:\/\/mirror.clarkson.edu/g' 
 
 RUN apk add --no-cache bash yarn
 
-COPY . /opt/organice
-WORKDIR /opt/organice
-
-RUN yarn install \
-    && yarn global add serve \
-    && yarn build \
-    && yarn cache clean \
-    && rm -rf node_modules
-
 # No root privileges are required. Create and switch to non-root user.
 # https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
 RUN addgroup -S organice \
     && adduser -S organice -G organice
+
+COPY --chown=organice . /opt/organice
+WORKDIR /opt/organice
+
+RUN yarn install \
+    && yarn global add serve
+
 USER organice
 
 ENV NODE_ENV=production
 EXPOSE 5000
-ENTRYPOINT ["serve", "-s", "build"]
+ENTRYPOINT ["/bin/bash", "-c", "yarn build && serve -s build"]
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -6,3 +6,5 @@ services:
       context: .
     ports:
       - "5000:5000"
+    volumes:
+      - ./.env.sample:/opt/organice/.env
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -4,3 +4,5 @@ services:
     image: 'twohundredok/organice:latest'
     ports:
       - '5000:5000'
+    volumes:
+      - ./.env:/opt/organice/.env


munen:
Hi @maljub01 

Thank you for looking into this!

I've tried it, but it doesn't work for me:

image

I've tried removing the container with `docker rm -f CONTAINER_ID` and re-create it with `docker-compose -f docker-compose-dev.yaml up`.

Having said that, even if it worked, I'm not sure if we should go down that path. Re-compiling everything on every instantiation of a container doesn't feel idiomatic for Docker. Personally, I haven't seen that before. Have you? I can be convinced I'm wrong, of course^^

Personally, I'd prefer to go down a different route. `dot_env` understands multiple .env files (https://dev.to/getd/how-to-manage-secrets-and-configs-using-dotenv-in-node-js-and-docker-2214), for example. However, that probably also doesn't help, because we compile it in. I only quickly checked.

If there's no other way to go without re-compilation, I'd personally prefer to have more documentation on the manner. It would feel more correct to me. Unless I'm mistaken and there is precedence to see that this would be idiomatic.

maljub01:
Hi @munen,

Sorry for the delay in getting back to you. I got a bit busy the past few days. You can see my responses inline.

> Hi @maljub01
> 
> Thank you for looking into this!
> 
> I've tried it, but it doesn't work for me:
> 
> I've tried removing the container with `docker rm -f CONTAINER_ID` and re-create it with `docker-compose -f docker-compose-dev.yaml up`.
>

The reason it didn't work is that I haven't edited the docker-compose files to mount the `.env` files. I just pushed a commit that does that, so it should now work for you, just make sure that you rebuild the image first by running `docker-compose -f docker-compose-dev.yaml up --build`.

I made `docker-compose-dev.yaml` mount `.env.sample` while `docker-compose.yaml` mounts `.env`. So depending on which docker-compose file you use you'll get a different file.

On that note, I don't think you should have two separate docker-compose files. You can keep the `build` settings in `docker-compose.yaml` since it won't rebuild the image unless you ask it to do that.

> 
> Having said that, even if it worked, I'm not sure if we should go down that path. Re-compiling everything on every instantiation of a container doesn't feel idiomatic for Docker. Personally, I haven't seen that before. Have you? I can be convinced I'm wrong, of course^^
> 

That's fair. I really didn't like having to recompile everything either. As for whether it's idiomatic, that really depends on how you view it. If you view this as:

1) Recompiling: Then this is definitely not idiomatic. A Docker image should not compile artifacts at runtime. Typically, the application would be compiled at build time, and the build dependencies discarded one way or another.

2) Setting up / Customizing the application: Then this would be quite idiomatic in my opinion. If you take a look at some of the official Docker images with applications whose configuration generates additional artifacts, you'll see it's quite common. Some quick examples I found are: [postgres](https://github.com/docker-library/postgres/tree/master/10), [mysql](https://github.com/docker-library/mysql/tree/master/8.0), [wordpress](https://github.com/docker-library/wordpress/tree/master/php7.4/apache) and [redmine](https://github.com/docker-library/redmine/tree/master/4.1). If you take a look at their code, you'll see they all have their `ENTRYPOINT` running a `docker-entrypoint.sh` script which sets things up based on the environment variables supplied by the user.

I had the 2nd view in mind when I made those changes, but since this is a node app, it seems we can't have one without the other.

Unfortunately, I can't think of other node-based applications off the top of my mind that I can compare with. I'll try to make some time over the next few days to find some.

> Personally, I'd prefer to go down a different route. `dot_env` understands multiple .env files (https://dev.to/getd/how-to-manage-secrets-and-configs-using-dotenv-in-node-js-and-docker-2214), for example. However, that probably also doesn't help, because we compile it in. I only quickly checked.
> 

I don't think that would work, since we'd still need to recompile the end-user's env file. Perhaps it might be possible instead to have the user supply a JSON file to avoid recompilation, and that file can be requested/loaded dynamically when needed?

> If there's no other way to go without re-compilation, I'd personally prefer to have more documentation on the manner. It would feel more correct to me. Unless I'm mistaken and there is precedence to see that this would be idiomatic.

Agreed.

branch14:
Dear @maljub1, thank you for putting thoughts and time into this.

Could you check if the changes introduced in #835 will work for your use case. If so I vote for closing this PR in favor of #835.

maljub01:
@branch14 that's great. Thank you!

Issue #246, opened 2020-02-09
As a user, when I'm in focus mode and start a search, I want the search to be narrowed to the focused headers


munen:

munen:
Closed by #517.

PR #517, opened 2020-10-18
only search subheaders if a header is focused

tarnung:
Issue #246

When a header is focused, should there be an indication in the search modal that search is limited?
I think it's ok as a standard behaviour without notification but not everyone might agree.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,6 +4,13 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-10-19 Mon]
+** Added
+   - When a header is focused, and the user uses the 'search' or 'task
+     list' feature, then the searched header list is automatically
+     narrowed to only subheaders of the originally focused header.
+     - Thank you [[https://github.com/tarnung][tarnung]] for your [[https://github.com/200ok-ch/organice/pull/517][PR]] 🙏
+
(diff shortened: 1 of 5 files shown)


munen:
@tarnung Yet another great feature contribution :pray: 

> When a header is focused, should there be an indication in the search modal that search is limited?
> I think it's ok as a standard behaviour without notification but not everyone might agree.

I also think it's the appropriate behavior. In Emacs, when you narrow a buffer (that's not even Org specific functionality), then what you see is what search and edit operates on whilst the remainder of the buffer is 'out of sight, out of mind. As for an indication, imo none is needed. Sane defaults with good docs should suffice.

I've added an integration test, updated the documentation and added your contribution to the changelog.

Great work :pray: :bow:

schoettl:
Nice feature! For me "header is focused" sounds like the header is selected (i.e. toolbar and details are visible). But it actually means header is in "focus mode" or "sub-tree mode". So IMO the documentation is not perfectly accurate. But sorry that I'm not providing a better alternative / PR right now..

munen:
@schoettl Very good point, thank you for bringing it up! We could adopt the Emacs nomenclature of 'narrow' and 'widen': https://www.gnu.org/software/emacs/manual/html_node/emacs/Narrowing.html

Thb, I'd rephrase to we 'should' adopt. There's no reason to diverge on the naming from Emacs if we're duplicating an exact feature - unless there is a reason, of course.

PR #519, opened 2020-10-20
refactor: 'focusing' becomes 'narrowing'

munen:
This is according to the naming in Emacs: https://www.gnu.org/software/emacs/manual/html_node/emacs/Narrowing.html


munen:
@schoettl Can you give this a glance and check if I hit all the right spots, but didn't over-rename. I've read twice through the diff, organice still seems to work and the tests are also green.

That is, when you have time, of course. This change is not time critical.

schoettl (approved):
Looks good to me! Narrow/widen are new words for non-orgmode users but you explained it nicely. And IMO it's better (especially in the code) to have unique names for the two concepts "narrow" and "focus/select header".

But I cannot test currently because of #533

munen:
> Looks good to me! Narrow/widen are new words for non-orgmode users but you explained it nicely. And IMO it's better (especially in the code) to have unique names for the two concepts "narrow" and "focus/select header".

Sweet, thank you for the review!

munen:
Impressive that we got a merge conflict in such a short time in a reducer that hasn't been worked on for a while(;

Issue #504, opened 2020-10-10
Expand search by supporting for timeranges on CLOCK, SCHEDULED or DEADLINE


munen:
This will also enable the [Clock display functionality](https://github.com/200ok-ch/organice/issues/494) to be filtered to specific time ranges.

If this feature will not be implemented, a less generic option is to implement Emacs Org mode `org-clock-display-default-range` in the settings.

schoettl:
Interesting! To comply to the current search syntax, the new (how did I called it?) search terms would be

```
scheduled:2020-10-11    deadline:2020-10-11     clock:2020-10-11
```

We could allow abbreviations for those, e.g. `sched` and `dead[l]`.

Regarding to the date specification:

I have the feeling that @munen and I agree that mu4e query date ranges is the best ;) – right?

`man mu-query` -> `/^DATE RANG`

Are there other suggestions (e.g. [hledger date query syntax](https://hledger.org/hledger.html#period-expressions))?

schoettl:
On that way we can also implement the full-text search with `description:word` or `desc:'word or text'`!

munen:
> Interesting! To comply to the current search syntax, the new (how did I called it?)

I think you called it "headline filter". "search syntax" works for me, too. The latter is probably easier to find for a new contributor, but the former is technically equally correct.

> search terms would be
> `scheduled:2020-10-11    deadline:2020-10-11     clock:2020-10-11`

I was thinking the same :+1: 

> I have the feeling that @munen and I agree that mu4e query date ranges is the best ;) – right?

Full agreement :100: 

Here's the doc for mu4e queries: https://www.djcbsoftware.nl/code/mu/mu4e/Queries.html

The relevant terms would be:

- YYYY-MM-DD
- \d[dwmy] (i.e. `2d` for 2 days)
- today
- now

The separator between two dates would be `..`.

The second date is optional, so these would be valid terms:

- `scheduled: today`
- `scheduled: 1w..`

> On that way we can also implement the full-text search with description:word or desc:'word or text'!

That would certainly be hugely appreciated by the community :+1:

tarnung:
> The relevant terms would be:
> 
>     * YYYY-MM-DD
> 
>     * \d[dwmy] (i.e. `2d` for 2 days)
> 
>     * today
> 
>     * now

I suggest to also allow
* YYYY
* YYYY-MM

> The separator between two dates would be `..`.
> 
> The second date is optional, so these would be valid terms:
> 
[...]

PR #520, opened 2020-10-20
[WIP] display clock times in search

tarnung:
I refactored my previous changes to TitleLine to allow a string to be passed in as an 'addition'-prop. That string is then displayed right aligned.
This allows to show the recursive clocked times in the regular header list (when the corresponding setting is activated) and the per-item sums in search (when a 'clock:' search term is present).
Additionally there is a sum of clocked times over all search results displayed in the top right of the search modal.

The discussion whether this is the right way to display clock times in search started over at #504 and is not yet resolved. This is my proposal.

### DIFF
--- a/src/components/OrgFile/components/AgendaModal/components/AgendaDay/stylesheet.css
+++ b/src/components/OrgFile/components/AgendaModal/components/AgendaDay/stylesheet.css
@@ -43,6 +43,7 @@
 }
 
 .agenda-day__header__header-container {
+  width: 100%;
   margin-top: 18px;
 }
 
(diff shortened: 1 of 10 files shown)


tarnung:
The duration shown is the sum of all clocking entries that touch the searched timerange.
E.g. if one end of a clocking entry is inside the timerange the enty adds its full duration to the sum even if the other end of the clocked duration is outside the searched timespan.

munen:
I've tried it locally and so far, it works nicely! :pray:

tarnung:
replaced by #524

Issue #513, opened 2020-10-17
Documentation for developers in Windows environment


schoettl:
Some unit tests with strings like
```
`multi
line`
```
fail because Git on Windows check out CRLF, not just LF. Another reason might be that there IDE/editor saves CRLF.

_Originally posted by @tarnung @munen @schoettl in https://github.com/200ok-ch/organice/pull/505#issuecomment-710776100_

Here are some ideas to fix this.

A good place for documentation would be https://github.com/200ok-ch/organice#development

munen:
Best would be if a solution can be found and documented by a Windows developer.

schoettl:
Copied from other #505:

> If git config core.autocrlf false fixes the issues, that would definitively be preferable to changing all occurences of \n [to EOL from 'os'] (;

> ... plus the editor/IDE must save LF, not CRLF.

> Maybe that can be addressed with * text=auto (https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html#_end_of_line_conversion).

munen:
Closed by https://github.com/200ok-ch/organice/pull/563

PR #524, opened 2020-10-21
Feat/recursive clock times in search

tarnung:
This is a rival PR to #520, relating to the discussion in #504.

In this version the `clock:` search term includes headers that have time logged on their children.

### DIFF
--- a/src/lib/clocking.js
+++ b/src/lib/clocking.js
@@ -46,3 +46,21 @@ export const updateHeadersTotalTimeLogged = (headers) => {
   });
   return headersWithtotalTimeLoggedRecursive;
 };
+
+export const updateHeadersTotalFilteredTimeLogged = (filters, headers) => {
+  if (!headers) {
+    return headers;
+  }
+  const headersWithtotalTimeLogged = headers.map((header) =>
+    header.set('totalFilteredTimeLogged', totalFilteredTimeLogged(filters, header))
+  );
+  const headersWithtotalTimeLoggedRecursive = headersWithtotalTimeLogged.map((header, index) => {
+    const subheaders = subheadersOfHeaderWithIndex(headersWithtotalTimeLogged, index);
+    const totalTimeLoggedRecursive = subheaders.reduce(
+      (acc, val) => acc + val.get('totalFilteredTimeLogged'),
+      header.get('totalFilteredTimeLogged')
+    );
+    return header.set('totalFilteredTimeLoggedRecursive', totalTimeLoggedRecursive);
+  });
+  return headersWithtotalTimeLoggedRecursive;
+};
--- a/src/lib/clocking.js
+++ b/src/lib/clocking.js
@@ -1,7 +1,7 @@
 import { subheadersOfHeaderWithIndex } from './org_utils';
 import { dateForTimestamp } from './timestamps';
 
-export const totalTimeLogged = (header) => {
+const totalTimeLogged = (header) => {
   const logBookEntries = header.get('logBookEntries', []);
 
   const times = logBookEntries.map((entry) =>
@@ -9,6 +9,21 @@ export const totalTimeLogged = (header) => {
       ? dateForTimestamp(entry.get('end')) - dateForTimestamp(entry.get('start'))
       : 0
   );
+
+  const addedTimes = times.reduce((acc, val) => acc + val, 0);
+  return addedTimes;
+};
+
+export const totalFilteredTimeLogged = (filters, header) => {
+  const clocks = header
+    .get('logBookEntries', [])
+    .map((l) => [l.get('start'), l.get('end')])
+    .filter(
+      ([start, end]) => start !== undefined && start !== null && end !== undefined && end !== null
+    )
+    .filter((ts) => ts.some((t) => filters.every((f) => f(t))));
+
+  const times = clocks.map(([start, end]) => dateForTimestamp(end) - dateForTimestamp(start));
 
   const addedTimes = times.reduce((acc, val) => acc + val, 0);
   return addedTimes;
--- a/src/lib/clocking.js
+++ b/src/lib/clocking.js
@@ -29,7 +29,7 @@ export const totalFilteredTimeLogged = (filters, header) => {
   return addedTimes;
 };
 
-export const updateHeadersTotalTimeLogged = (headers) => {
+export const updateHeadersTotalTimeLoggedRecursive = (headers) => {
   if (!headers) {
     return headers;
   }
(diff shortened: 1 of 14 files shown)


tarnung:
I would still like to have a sum in this version.
At the moment it is the same as in #520, namely the total of all relevant clocking entries on the shown headers.
This might be confusing, since it says 0:00 if only a header with time logged on its children but not on itself is shown.

src/lib/clocking.js
```diff
@@ -14,7 +14,22 @@ const totalTimeLogged = (header) => {
   return addedTimes;
 };
 
-export const updateHeadersTotalTimeLogged = (headers) => {
+export const totalFilteredTimeLogged = (filters, header) => {
+  const clocks = header
+    .get('logBookEntries', [])
+    .map((l) => [l.get('start'), l.get('end')])
+    .filter(
+      ([start, end]) => start !== undefined && start !== null && end !== undefined && end !== null
```
munen, line 22: You're always making sure that a var is neither `undefined` or `null`. In JS, this is not required imo. For example:

image

Lots less typing, and easier to read(;

Checking for `undefined` or `null` only makes sense if you want to check that they actually are one of those values, but not to see if there's a value behind the var.

I'm often seeing this kind of code from people who are more used to statically typed languages. However, I wouldn't actually assume that's the case for you, because all your other JS is very much functional, concise and overall just a great style. So, I'm curious (to potentially learn that I'm wrong^^): Why do you check ensure a var is neither `undefined` nor `null` like this?

munen:
Quite a bit more coding required to do it like this. Personally, I have to say that I like it much better, so thank you for the extra effort! :pray: 

> I would still like to have a sum in this version.

Yes, that would actually be a great feature. It's kind of the smallest common denominator between `org-clock-display` and `org-clock-report` and very benefitial to see it quickly. It's good thinking to add it, too!

> At the moment it is the same as in #520, namely the total of all relevant clocking entries on the shown headers directly (meaning nonrecursively).
> This might be confusing, since it says 0:00 if only a header with time logged on its children but not on itself is shown.

Yes, that could be confusing. I've nagged you long enough and you've actually went the extra mile and re-wrote the feature like I propsed. I'm fine with the sum either way. If you want to leave it like it is right now, for me it's good enough if we add documentation (and a `title` attribute) that explains the behavior.

src/lib/clocking.js
tarnung: > Why do you check ensure a var is neither undefined nor null like this?

I guess it's a reflex since zero is also false in js and i ran into that problem while adding durations. But in this example start and end should be objects so the chance that they ever are zero is hopefully very slim ;)
And yes, I work as a java programmer :P

tarnung:
I'm not quite sure how the snapshot testing is supposed to work. I seem to be unable to add the right kind of empty lines.

tarnung:
> ...and you've stated your preference on the topic of the sum.

I added a tooltip. Besides that I can't think of a way to make it more intuitive right now. We can always change it later if a better idea comes up.

munen:
> I'm not quite sure how the snapshot testing is supposed to work. I seem to be unable to add the right kind of empty lines.

The snapshot testing is a `jest` feature: https://jestjs.io/docs/en/snapshot-testing.html

You're don't have to write the snapshots by hand. You can generate them, check them in and then when doing a change, you'll get a nice diff when running `yarn test`. If you're happy with the diff, you can re-generate (or update) the snapshot which again is checked in.

It can be very helpful to catch regressions - or sometimes even to generate the snapshot to read through the generated markup and see if it's valid. But it's not really necessary nor realistic to write them by hand(;

image

munen:
> I'm not quite sure how the snapshot testing is supposed to work. I seem to be unable to add the right kind of empty lines.

Works for me :+1:

src/lib/clocking.js
munen: > I guess it's a reflex since zero is also false in js and i ran into that problem while adding durations. But in this example start and end should be objects so the chance that they ever are zero is hopefully very slim ;)

Ok, I see. Yes, there's other stuff in JS (and other dynamic languages) that's falsy. `0` is false due to it's C heritage. I'm not doing myself favors here talking to a Java programmer, but there's obvious *cough* consensus that `0` is false, isn't there?(;

image

munen: Having said so, as far as I know (and I'm quite sure), it's a convention in all of the above to use the `var` itself as a qualifier for 'truthiness', because it catches a lot of stuff and is quite readable. The specific semantics of what it catches are dependend on the language, of course, but if they were all the same, we could just use the one and wouldn't that be boring(;

tarnung:
> You're don't have to write the snapshots by hand. 

I suspected as much.

> You can generate them

I guess i should not only read the colored parts of the console output ;)

munen:
> I guess i should not only read the colored parts of the console output ;)

*hihi*, No worries, it's easy to miss, because there's quite a bit of output(;

Actually, I'm quite impressed how far you've dug into the codebase and have not gotten stuck so far! Said differently: You're kind of due to ask at least a tooling related question. So there you go(;

munen (approved):
Looks good to me :100:

munen:
Args, what have I done. I wanted to add attribution and some documentation for which I merged in `master`. Looks like `magit` did a rebase instead, so now we have some duplicated commits.

Please, nobody rebase nor make any other chnges to this branch. I'll clear my mistake up and force push a version without the duplication.

tarnung:
I'll keep my hands off for the time being.

munen:
> I'll keep my hands off for the time being.

Thx!

I do have a global setting `git config --global pull.rebase false`, so I don't know what happend. I must have accidentally hit something in Magit.

In any case, I cleaned it up - all good now. You're `update snapshot` commit is in, as is the attribution in the changelog and the bit of documentation. But no more duplication.

munen:
This PR is good to merge for me. Are you content with it, as well, @tarnung?

tarnung:
Good to go. Thanks for your feedback and for adding documentation.
Sorry to bother you so much, but could you update the snapshot again? I have to solve the line break issue #513  first to make it less bothersome for me.

munen:
> could you update the snapshot again? I have to solve the line break issue #513 first to make it less bothersome for me.

Sure thing!

You might just have caught a bug with the simplication. The `addition` is sometimes an empty string, so before the span with the min-width was rendered which effectively created a padding(;

munen:
> Good to go. 

Perfect, will merge after this comment!

> Thanks for your feedback and for adding documentation.
> Sorry to bother you so much...

No bother at all. I'm having quite a great time over here - lots of great additions coming in to organice whilst the code quality is already perfect and the discourse on UX/design is very fast on the one hand, but equally so appreciative. \\(^_^)/

As long as I can manage some time away from other duties, this is as far as I'm concerned, it's the perfect storm for a FOSS project and I could do this any time(;

Thank you for your continued effort and contributions, I'm sure many users will greatly appreciate them! :pray: :bow:

Issue #487, opened 2020-09-29
Dark mode?


Naheel-Azawy:
**Is your feature request related to a problem? Please describe.**
Well, yes you can call it a problem. The light theme is a serious deal breaker for me. It hurts my eyes in my dark room.

**Describe the solution you'd like**
An option to make the theme dark. Or perhaps add multiple themes.

**Describe alternatives you've considered**
Maybe tell be which css I should I start from? I'm a bit in rush for now to dig.

**Additional context**
Nothing much, I just really like this project. Thanks to all who worked on it!

munen:
Hi @Naheel-Azawy 

Thank you for checking out organice 🙏 

You are in luck! organice uses the popular [Solarized color scheme](https://ethanschoonover.com/solarized/). One of the properties of Solarized is that it implements light and dark modes using the same colors, so adding a Solarized dark mode should be pretty easy.

The current color scheme is mostly implemented in the CSS files in the [`src` directory](https://github.com/200ok-ch/organice/tree/master/src). The original PR to introduce it is #12.

Would you be interested in working on this?

Naheel-Azawy:
That sounds cool.

> Would you be interested in working on this?

Am I interested in that? Of course yes. But can I? Unfortunately, probably not..

So I guess let's keep this open till I or anyone else get some time to work on it

munen:
> Am I interested in that? Of course yes. But can I? Unfortunately, probably not..

I understand. In this case, we've got documentation on your options: https://organice.200ok.ch/documentation.html#how_we_work_with_issues

Naheel-Azawy:
Mmm, ok. Maybe some day I get back to this. But thank you anyway. Keep up the good work!

munen:
:+1:

munen:
Dark mode has been implemented by https://github.com/200ok-ch/organice/pull/527

Naheel-Azawy:
Super cool! Even though I'm not a big fan of the solarized bluish dark but that is great progress! Thank you @munen and @tarnung

munen:
@Naheel-Azawy Solarized is quite definitively one of the most popular choices out there. Having said that, if you're particular about your preferred color scheme, there's good news: With the new implementation, it's actually quite easy to add additional color schemes. You're very welcome to contribute one that is to your liking.

tarnung:
[...]

PR #527, opened 2020-10-22
Feat/dark mode

tarnung:
issue #487

known issues: 
- swiping does fancy things with interpolated style that is not changed to dark mode.
- cleaning up css to only use solarized colors has some small impact on bright mode.


munen:
Dark mode, super cool!

First impressions:

Some things are really popping nicely now, like switches:

image

Was it intentional to remove the delimiting vertical/horizontal lines around tab-buttons or is that part of the 'small impact' you've referred to in the description? Removing the delimiting vertical lines between options (and files/folders in the files view) could actually be argued to be cleaner. For the tab-buttons, I think it's helpful to see them as 'one entity'.

It's quite mystical how the shadow becomes a glow in dark mode :smile: 

image

src/base.css
```diff
@@ -42,14 +46,14 @@ h2 {
   background-color: var(--magenta);
   /* TODO: What's the right way to have text on an action button in */
```
munen, line 47: If we're going with `--base3`, then this TODO is done imo.

munen:
You've replaced quite a few colors with `--base3`:

image

That seems like a wide range of colors to be replaced. It could
totally be the best choice, at this point I'm just writing down stuff
that I notice whilst reading through the changes.

src/colors.css
```diff
@@ -7,6 +7,7 @@
   --base01:     #586e75;
   --base00:     #657b83;
   --base0:      #839496;
+  --base0-soft: rgba(131, 148, 150, 0.75);
```
munen, line 10: Just out of curiousness: `-soft` is not defined by solarized, it's an addition you came up with, right?

munen:
The opposite of `dark` is `light` in the naming of Solarized. I think it's even the canonical naming (since Emacs has the same distinction and CSS has it too).

tarnung:
> Was it intentional to remove the delimiting vertical/horizontal lines around tab-buttons or is that part of the 'small impact' you've referred to in the description? Removing the delimiting vertical lines between options (and files/folders in the files view) could actually be argued to be cleaner. For the tab-buttons, I think it's helpful to see them as 'one entity'.

&

> That seems like a wide range of colors to be replaced. It could
totally be the best choice, at this point I'm just writing down stuff
that I notice whilst reading through the changes.

No, All the changes in the css files are up for debate. I just matched them to the nearest color in the solarized palette (mostly base03/base3 for dark/bright and base0 for gray). That means I might have matched custom colors that were used to accentuate something to the color of their surrounding hence the vanishing lines. 

> Just out of curiousness: -soft is not defined by solarized, it's an addition you came up with, right?

base1-soft already existed. It's base1 with added transparency. I think it's used to highlight active headers. I mapped other colors with transparency that were used throughout the code to solarized colors with added transparency. base0-soft is used for shadows, green-soft is used for active cell highlighting in tables. base0-soft is currently not changing between modes.

> The opposite of dark is light in the naming of Solarized. I think it's even the canonical naming (since Emacs has the same distinction and CSS has it too).

Noted.

munen:
I've added a PR (which merges into this PR) to add an automatic toggle if the user switches the preferences on an OS or browser level: https://github.com/tarnung/organice/pull/1

It's only a goodie. It will only make a switch if the user changes the preferences while organice is running.

munen:
> No, All the changes in the css files are up for debate.

Noted. I'd continue like this:

1. Ensure there's no regression on light mode. That doesn't mean that there can be no change at all, but any change should be equally good or better than what we have.
2. Do what you did; a 'pragmatic' first version of dark mode.
3. Roll out.
4. If we encounter better settings along the way for either dark or light mode, we can quickly change/add them any time.

munen:
What do you think about moving this setting further up and using `TabButtons`?

image

Up there, we have other styling related settings in the same way. Also, the dark vs light mode debate seems prevalent enough in the community that it's a setting with a high change that the user has an opinion on how to set it whereas the other settings are more domain specific.

src/util/misc.js
```diff
@@ -13,3 +13,32 @@ export const formatTextWrap = (text, maxLineLength) => {
     }
   }, '');
 };
+
+export const darkModeColors = () => {
```
munen, line 17: Super nitty, but wouldn't `src/lib/color.js` be a better fit for color related functions? :nerd_face:

munen:
Ok, I've said all the nitty little details that I can see in this PR, now on to the more important stuff: Yet another great addition @tarnung :pray: :bow: This is something that a few people have asked for in the past. I haven't dabbled with it, yet, because I personally switched back from dark mode after many years to light mode, so I deemed it more fitting for someone to implement it who has his head currently wired to the aesthetics of dark mode. I'm stoked this time has come! :smile:

munen:
I'll wait with merging until these are addressed:

- [ ] https://github.com/tarnung/organice/pull/1
- [ ] styling of tab-buttons
- [ ] using a tab-button and moving the setting to the top
- [ ] everyone feels safe that light mode has no regressions
- [ ] Maybe @schoettl is a dark mode user and has an opinion on the proposed settings, so let's give him some time to see this nice gem and voice an opinion.

src/base.css
munen: Thx!

src/util/misc.js
munen: Thx!

munen:
Nice and super quick changes, I've resolved the related discussions and updated the todos accordingly :+1:

tarnung:
@munen please confirm that your os-level-switch still works after changing the setting to tab buttons.

munen:
@tarnung 

> please confirm that your os-level-switch still works after changing the setting to tab buttons.

Confirmed :+1:

munen:
Maybe Agenda and Search need some changes for dark mode?

image


image

munen:
NB: I have deployed the current version to https://staging.organice.200ok.ch/

tarnung:
yes, some more adjustments needed.

munen:
I visually like the change that timestamps (and other actionable links) in light mode are black now.  

image

image

image

So, visually I like it. But 'black' doesn't feel like a good indicator for 'actionable'. What do you think?

Another thing I just realized is that it doesn't make sense to color `:visited` links differently in organice. It's not a browser per se. I'd prefer it to always have the same color, I think. Of course, that's in the legacy CSS scheme already. What do you think?

tarnung:
>  'black' doesn't feel like a good indicator for 'actionable'.

The browser provided blue did not work for dark mode. I thought it should change between modes, so i chose the thing that I thought fits best: the outer ranges of the neutral color spectrum provided by solarized.

Since I single handedly decided to treat magenta and violet as two colors that get switched for dark mode, we could use this. I wouldn't like the looks of it that much, but it would very much scream 'actionable' since it's the action button colour.

tarnung:
> Another thing I just realized is that it doesn't make sense to color :visited links differently in organice.

I agree.

tarnung:
Changing TODO & DONE to solarized red & green might not be ideal since there is green text for some indentation levels. I think it's ok, but I would like some more opinions on that.

munen:
> The browser provided blue did not work for dark mode. I thought it should change between modes, so i chose the thing that I thought fits best: the outer ranges of the neutral color spectrum provided by solarized.

I understand.

> Since I single handedly decided to treat magenta and violet as two colors that get switched for dark mode, we could use this. I wouldn't like the looks of it that much, but it would very much scream 'actionable' since it's the action button colour.

Hm.. I actually like the dark mode look, but the light mode version would take some getting used to. I could, maybe. The more I look at it.. On the other hand, it's actually also weird, because we're only using this color for the action buttons.

I get the dilemma now. And initially I said that I like the look of the new 'black' color. It's good we talked about it, though. And thank you for looking at and providing an alternative! Design and color schemes are not trivial(; If it's alright with you, I'd say let's go with your initial approach.

munen:
I like the changes from commit bb38c68, agenda and search are looking good now! I'll update the tasks above.

munen:
> Changing TODO & DONE to solarized red & green might not be ideal since there is green text for some indentation levels. I think it's ok, but I would like some more opinions on that.

The Emacs solarized theme actually uses the same "Foreground" color for the second level as for "org-done", so I'd say we're fine. 

image

image

---

NB: It does use #2aa198 (`--cyan`) instead of red for "TODO", though. It could be that we're using red so far due to my setup (I use a #dc752f).

Personally, I'm perfectly fine fine with your options for the time being! 

image

munen:
>> Another thing I just realized is that it doesn't make sense to color :visited links differently in organice.

> I agree.

Nice. Then we're already good, because by explicitly declaring a style for `a`, you've taken care of it(;

munen:
I went through all screens in both light and dark mode and have to say, that they all look fabulous in both modes :champagne: 

Good job!

munen (approved):
Looks good to me :100:

munen:
I've fixed two minor issues (snapshot test and persistence of the setting) and have added attribution.

This PR is in great shape, let's push it upstream! :100: :confetti_ball: 

@schoettl If you're a dark mode user and want to chime in on the settings, please do. For the moment, this looks too good to keep as an unmerged PR(;

PR #530, opened 2020-10-23
doc: Google Drive is not in production mode for organice.200ok.ch

munen:
Document on the sign in page what that means for the users and what her options are.

On sign in, there's the option to click on 'important news' for Google Drive. Then, this is shown: 

image


munen:
@AlexMihov There's more users asking about Google Drive via various channels (including private email). I've added a bit of a help in the hopes that there will be fewer mails in the future. But a fix would be nice, too(;

PR #531, opened 2020-10-24
set all colors when switching color scheme

tarnung:
In preparation for supporting additional color schemes the color functions should set all colors, not just the ones that currently change between light and dark mode.

In addition I added a rough documentation to base.css.


munen:
Good thinking and very nice preparation in case someone wants to bring in more themes! :pray:

PR #534, opened 2020-10-25
Feat/interpreted style for color schemes

tarnung:
Style interpolation for the swipe action assumed light mode background color as a starting point. I added a function that reads out the defined background color.

It's not the most elegant fix. Maybe there is a pure css solution to the whole style interpolation but I have no expertise in this area.

### DIFF
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -211,6 +211,14 @@ describe('Render all views', () => {
 
     describe('Renders everything starting from an Org file', () => {
       test('renders an Org file', () => {
+        // INFO: This snapshot is semantically correct, but it does
+        // not have color theme information. We're implementing the
+        // color themes with an API that mutates the DOM in place (see
+        // `color.js::loadTheme`. We cannot use this API in jest
+        // tests, because calling the API does not yield the same
+        // side-effect as in a browser. Hence, some colors in this
+        // snapshot are off, but that's ok. We do colorScheme testing
+        // by eye and not with automated tests.
         expect(getAllByText(/\*/)).toHaveLength(7);
         expect(container).toMatchSnapshot();
       });
--- a/src/components/OrgFile/__snapshots__/OrgFile.integration.test.js.snap
+++ b/src/components/OrgFile/__snapshots__/OrgFile.integration.test.js.snap
@@ -331,7 +331,7 @@ exports[`Render all views Org Functionality Renders everything starting from an 
         >
           <div
             class="left-swipe-action-container"
-            style="width: 0px; background-color: rgb(211, 211, 211);"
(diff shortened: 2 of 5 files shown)


munen:
Thank you for this PR, @tarnung!

It's a good change and I want to merge it, but this PR shows one issue we have introduced before: At least since switching from 'darkMode' yes/no to 'colorScheme' 'light'/'dark', we had no default colorScheme. This is the reason for the tests to fail, too:

image

Since there is no default colorScheme, the function you wrote defaults to black (which is reasonable!). 

I propose that we should fix the issue of not having a default colorScheme in this PR or in parallel. Otherwise, we'll have to 'break' the test in order to have green tests. 

This is what it looks like for the user at the moment if she has not selected one, yet:

image

The 'solarized light' theme is visibly active, but no color scheme is selected in the settings. Likely, this will get more confusing as soon as there's more color schemes being introduced.

I have added a 'partial fix' in 57061cb. It's only partial, because it's the place where this information should be. Unfortunately it is not applied. I would need additional time for investigating this.

munen:
Hm.. I've merged `master` which has the fix for using defaults. But the snapshot still fails.

tarnung:
I don't know exactly what is going on, but i debugged it so far that I can say that during testing reading the css variable returns an empty string.
So it seems that colors.css is not loaded and the theme setup function is not run either.

munen:
Ok, I looked into it some more.

So, we have a architectural issue. Generally, in organice we're using React to create Views and don't use HTML5 APIs that mutate the DOM.

For the color theme implementation, we do:

```js
  const style = document.documentElement.style;
  Object.entries(themes[theme][colorScheme]).forEach(([k, v]) => style.setProperty(k, v));
```

That mutates the DOM in place.

We cannot do that in a test. Just calling the code from the test runner will not have the side-effect that would happen in a browser. 

I've spent some time thinking and came to the conclusion that the required changes would be pretty big, because we couldn't use that API. Instead we would have to build a styled component that wraps the current application. 

Since the current implementation does work well and this is the only test that's failing,

I have updated the snapshot in place and added documentation around the previously failing test.

tarnung:
Thank you for the effort.
One other possibility to specifically fix the test would be to write the background-color to redux and read it from there. Then there would be no need to read it from dom too.
I agree that setting the css variables is probably the least offensive use of direct dom manipulation in organice. So I think it's ok to update the snapshot.

munen:
> One other possibility to specifically fix the test would be to write the background-color to redux and read it from there.

True. Doing this for all colors in all cases would lead to the option "Instead we would have to build a styled component that wraps the current application.".

> I agree that setting the css variables is probably the least offensive use of direct dom manipulation in organice. So I think it's ok to update the snapshot.

We're on the same page here. Let's merge your nice changes before they get stale(;

PR #536, opened 2020-10-25
documentation dropbox-app

tarnung:
It seems that localhost must be defined as a redirect uri. It no longer works by default.


munen:
Thank you for updating the docs accordingly. 

I didn't know, because it's been a while since I last created a Dropbox app, so we've already got the first win that you did it yourself(;

PR #537, opened 2020-10-25
simplify theme loading

tarnung:
we should not spread the theme loading logic into components unnecessarily.

setting theme color for android is not yet tested, since i did not get port forwarding to work yet.


munen:
Nice refactor!

Regarding testing theme loading for Android: Let's do it in production(;

PR #538, opened 2020-10-26
fix: use a NixOS-compatible shebang line for bash in compile scripts

ivan:
This fixes some shebang lines for NixOS, which does not have a `/bin/bash`. The only thing NixOS has in `/bin` is `/bin/sh`, and the only thing it has in `/usr` is `/usr/bin/env`.

### DIFF
--- a/bin/compile_and_upload.sh
+++ b/bin/compile_and_upload.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
--- a/bin/compile_doc.sh
+++ b/bin/compile_doc.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if ! [ -e README.org ]; then
     echo >&2 "Error: Please run this script from top of the organice repository."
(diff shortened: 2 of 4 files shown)


munen:
Works for me, `/usr/bin/env` is available on Debian, so it's available on CI (and my dev environment). According to [StackOverflow](https://apple.stackexchange.com/questions/297533/does-macos-have-usr-bin-env), it's available on macOS.

@tarnung I don't know what environment you use to develop in Windows. Can you confirm that this won't break the scripts for you, please?

munen:
@ivan Nice to see a NixOS user dropping into this repository, in any case. Welcome and thank you for contributing :pray:

schoettl:
Works on Arch, too.

munen:
> Works on Arch, too.

Nice!

tarnung:
I get js errors when running the scripts. With and without this change. I guess that means this change does not impact them. That's a goahead from me :P

munen:
@tarnung 

> I get js errors when running the scripts. With and without this change.

I hope not for all four scripts - at least building the docs and the parser should work on Windows, too. The other two scripts we use for deployment via FTP, so those aren't strictly necessary for everyone.

What kind of errors do you get? If you want, we can try fixing those off this PR.

munen:
Very well, it seems the shebang works equally well as the previous one.

Welcome @ivan as a new contributor to organice! :pray: :bow:

Issue #539, opened 2020-10-26
How to decide on additional themes?


tarnung:
Now that we have the setup for it, it's rather easy to add additional themes.
I demonstrated that by researching some popular themes and adopting them for organice. (See  #487 for dark mode examples, see below for light mode examples.)
But how do we decide on themes that get included? Is there an upper limit? Should there be a balance between light and dark modes?

I'm ok with having many themes. But then the question becomes how they are chosen in the menu. Maybe a dropdown or a collapsing list?

What should I do with the themes I made so far? I could do individual draft-prs that could be promoted to live prs once someone shows interest in a certain theme. Or I could make one big pr with all the themes I consider ready in case we go with a 'more is better' approach.

gruvbox light
![gruvboxlight](https://user-images.githubusercontent.com/35081013/97202465-f14ef300-17b3-11eb-8d52-cef2b75e2c9e.png)

one light
![one-light](https://user-images.githubusercontent.com/35081013/97202487-fca21e80-17b3-11eb-886c-d8f76f15c2f0.png)

smyck light (there isn't supposed to be a light version, so consider this experimental)
![smyck-light](https://user-images.githubusercontent.com/35081013/97202516-0592f000-17b4-11eb-8b93-4d0b6c3192fe.png)

munen:
It's a good question on how to decide. Here's my take:

Emacs is a very conservative piece of software. It's not easy to add new stuff and new adopters often find it quite ugly by default. So, they either don't use Emacs or go to a 'starter kit' like Doom or Spacemacs. Having said that, even Emacs itself ships with 14 themes[1] by default! So, even the conservative Emacs community deems having a 'lot' of themes by default something worthwhile.

Personally, I'm in favor that organice adopts different themes - now that you've put the necessary infrastructure in place to easily and semantically add new ones. So, even if we change the design/UX, the themes should kind of auto-update. Before, that would have been my biggest concern, it would have been easy to have regressions on design/UX changes. But now that all CSS is in one place, I think it's managable.

[...]

PR #540, opened 2020-10-26
Feat/additional themes

tarnung:
#539

### DIFF
--- a/README.org
+++ b/README.org
@@ -251,6 +251,49 @@ organice implements this customization strategy:
 #+END_EXAMPLE
 
 
+** Themes
+
+organice implements a strategy pattern, so that adding new themes is
+quite easy. These themes come bundled with organice, each in =light= and
+=dark mode=:
+
+*** Solarized
+
+#+html: <p align="center">
+#+html: <img style="height: 30em;"src="https://github.com/200ok-ch/organice/wiki/images/themes/solarized_light.png"/>
+#+html: &nbsp;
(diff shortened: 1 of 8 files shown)


tarnung:
code light and smyck light are not usable right now. I have to tune highlighting and maybe some colors

tarnung:
now we're good.

munen:
An initial visual check of the themes shows that they all look amazing! :100: 

There is a little bug if there's no theme selected, yet:

The danger of checking against `undefined` continues(;

image

image

Of course, we can fix the issue inline in this PR. But my hunch in https://github.com/200ok-ch/organice/pull/534#issuecomment-717863785 that the lack of a default `colorScheme` will cause more problems later on is confirmed. So, it would be best if both (default and a more robust check) can be tackled.

munen:
I looked into it a bit more. It was a good hunch at https://github.com/200ok-ch/organice/pull/534#issuecomment-717863785, indeed. `loadTheme` is called with the actual arguments `"null"` (as a string), because there's no default.

tarnung:
I thought I had covered all the bases by checking against undefined AND "undefined" (as a string).
But actually having a default value would be better, I agree. I'll look into it later this week.

Something else that just came to mind:
I left the colors.css file intact to give the variables an initial value. This would be no longer necessary technically, since we set all the variables in code (once this reliably works). But I guess there's no harm in keeping it (apart from the duplication).

tarnung:
I rebased, removed the check for 'undefined' and added a default value for theme.

munen:
@tarnung Really good work! I've added contribution to the changelog and added a section to the documentation, because this is too pretty to leave it off the docs(;

image

PR #550, opened 2020-11-01
Feat/multiple files

tarnung:
What it doesn't do:
- It does not load any files for you. You need to open multiple files in succession to bring them into the application.
- It only synchronizes the currently viewed file
- Capture only works in the currently opened file.
- GUI integration in search/tasklist/agenda is not done. I just throw the data in there to prove that it arrives in the right place. (Data is delivered in a map of pathname to list of filtered headers)
- Switching between files currently works by changing 'path' in the org-reducer. This does not change the browser url.
- A lot of things i will recall soon.

What it does:
- Search/view agenda/view tasklist over multiple files. Jump to headers regardless which file they are in.
- Refile between files (you need to jump through search to the other file and trigger a sync. Opening the target file through the file browser currently causes syncronization issues since it parses the org file again - without the headers refiled there in the meantime).
- org reducer actions are mostly unchanged but scoped to one file. Only chosen actions operate over multiple files.

### DIFF
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -30,11 +30,11 @@ export const setLastSyncAt = (lastSyncAt) => ({
 
 export const stopDisplayingFile = () => {
   return (dispatch) => {
-    dispatch({ type: 'STOP_DISPLAYING_FILE' });
-    dispatch(ActionCreators.clearHistory());
     dispatch(widenHeader());
     dispatch(closePopup());
     dispatch(setLastSyncAt(null));
+    dispatch({ type: 'STOP_DISPLAYING_FILE' });
+    dispatch(ActionCreators.clearHistory());
   };
 };
 
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -197,7 +197,14 @@ export const selectHeader = (headerId) => (dispatch) => {
   }
 };
 
-export const selectHeaderAndOpenParents = (headerId) => (dispatch) => {
+const changePath = (path) => ({
+  type: 'CHANGE_PATH',
+  path,
+});
+
+export const selectHeaderAndOpenParents = (path, headerId) => (dispatch) => {
+  // TODO: change path does not change the browser path
+  dispatch(changePath(path));
   dispatch(selectHeader(headerId));
   dispatch({ type: 'OPEN_PARENTS_OF_HEADER', headerId });
 };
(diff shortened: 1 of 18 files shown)


munen:
@tarnung This is a great first jab!💯

I have glanced over the code (instead of giving it a thorough read-through for a review) as a sanity check. From what I can gather, this looks like a sane approach that can safely be continued upon.

Some thoughts on the points you added in the description:

> It does not load any files for you. You need to open multiple files in succession to bring them into the application.

You can certainly leave this out as long as you want.

If I wanted to tackle this, I'd add this similarly to the 'capture templates' functionality in the settings. A suggested UX:

1. User navigates to settings
2. There's an initially empty list of 'agenda files' and an action button 'plus'
3. On hitting 'plus', the user can enter a path to an Org file

It maybe looks a bit like overkill for just a path. But I assume that it will make sense to have more options on these files at some point.

> Search/view agenda/view tasklist over multiple files. Jump to headers regardless which file they are in.

Wow, search and tasklist over multiple files, that's a sweet bonus. That's like composing Emacs Org mode agenda + organice search + [Emacs occur](https://masteringemacs.org/article/searching-buffers-occur-mode) - mighty powerful.

Whilst I like this a lot, I would also add the option to restrict search and tasklist to only one file on the assumption that people have different files for different teams. This is likely something where people will have different preferences, so it's probably best to have it as an option "Search/Tasklist over all agenda files" vs "Search/Tasklist over currently open file".

Of course, I'm only thinking out loud here. You're trying to get a grasp on how to do the big refactor and I don't want to add too many restrictions. I'm not saying it now, because this feature needs to be in there in the first draft, but it's probably helpful to know while thinking about architecture. From what I read in the code, it could work out quite nicely, because search will either just work on `headers` or `Array.from(headers.entries())`.

> Refile between files (you need to jump through search to the other file and trigger a sync. Opening the target file through the file browser currently causes syncronization issues since it parses the org file again - without the headers refiled there in the meantime).

Yes, opening through the file browser is currently for bookmarks, starting file for the SPA on Android/iOS and for opening links between files.

Depending on how you want to tackle this issue, maybe there's a freebie in here: Currently organice has only [partial offline support](https://organice.200ok.ch/documentation.html#offline_support). If we'd save the opened Org files in `localStorage`, then we would:

- Have less network traffic on re-opens (and be a lot faster)
- Have overall better offline support
- As an implementation detail maybe push the file contents at a later point; after a dispatch on refile

Just thinking out loud here. There's likely different ways to build this. Just skip the thought if it doesn't help and you've got a straight forward idea. I'm not trying to push more features in here than necessary, just thinking loudly in case there's synergies(;

> org reducer actions are mostly unchanged but scoped to one file. Only chosen actions operate over multiple files.

Generally speaking, I think this is a sane approach to have actions by default work on a single file, because most actions are designed to work on a single file.

---

Let me know when you want it merged to `develop`. Imo, this can even happen before the tests are adapted - you're still getting a feeling on how to build this epic out, no need to fix legacy tests or write new ones (unless it helps your dev process).

tarnung:
Thanks for your input @munen. 
I would like to get this into develop so I can start thinking about next steps.

munen:
Then, let’s merge it!

PR #552, opened 2020-11-01
Module "settings persister" does not persist falsy values

munen:


munen:
@tarnung This doesn't implement working default settings, yet, but it gets rid of the rather senseless defaults "undefined" and "null" (as actual strings, because those are not fun to test, against). This should help regarding the issue we discussed in #534 and #540.

PR #553, opened 2020-11-01
fix: Actually use and persist default settings

munen:
Based on #552


munen:
@tarnung This should help regarding the issue we discussed in #534 and #540. Once merged, I propse to merge `master` into your PRs.

PR #554, opened 2020-11-01
fix: Remove previously stored bogus values from localStorage

munen:
Based on #553 

For me, this resulted once in showing this console:

image

Looking at localStorage now, all remaining entries look reasonable.


munen:
@tarnung This should help regarding the issue we discussed in #534 and #540. Once merged, I propse to merge `master` into your PRs.

PR #556, opened 2020-11-01
Feat/improve header stickyness more

tarnung:
My first contribution ( #499) did not hold up. There were several situations were the layout broke. Whenever something not currently on screen was pulled into view, the header bar was thrown offscreen. For example when choosing an agenda item.

This should hopefully resolve it for good. (I thought that once before though.)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -161,7 +161,7 @@ export default class App extends PureComponent {
         <Router>
           <Provider store={this.store}>
             <div className="App">
-              <HeaderBar/>
+              <HeaderBar />
               <Entry />
             </div>
           </Provider>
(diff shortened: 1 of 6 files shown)


tarnung:
Hmm...
I just realized this change has the following visual effect on the desktop version when using a short file:
Screenshot_2020-11-07 organice

mobil version seems not to be affected.

tarnung:
The issue above is now adressed.

munen:
The changes look good to me. I did some testing and confirm your findings, too.

There's one additional change that I saw:

image

image

The header is now always as big as previously when selecting the "Font Size" "Large".

PR #557, opened 2020-11-02
disable user select

tarnung:
This one surely needs some discussion.
One of the main points that contributes to a bit of a clunky feel for me when using organice is accidentally selecting text - especially but not only on desktop browsers.

It might come in handy from time to time to copy things straight out of the browser without having to enter edit mode and copy things piecemeal, but - at least for me - the negative effect on regular use outweigh this rare benefit.

If there is a bunch of users that like the ability to copy text out of the main header list, we could add this behind a setting by switching a css variable. If I'm the only one that is bothered by accidentally selecting text, then I will humbly close this pr unmerged ;)

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,7 +4,15 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-11-08 Sun]
+
+** Changed
+
 * [2020-11-06 Fri]
+  - Safeguard against selecting text by accident.
+    - Before this change, it was possible to select text when doing a 'swipe'.
+    - Now, selecting/copying text is only possible in 'edit mode', effectively safeguarding against accidentally selecting text.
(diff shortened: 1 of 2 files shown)


schoettl:
I'd be OK with disabling text selection mode. No need for a setting IMO because it's still easy to select/copy text by switching to edit mode.

On Firefox, the swipe behavior of headers is very buggy. In some cases it only selects text, in other it selects a bit of text and then swipes, or it just swipes. It depends on the mouse position and the angular of the cursor swipe movement.

munen:
I agree, this is good default behavior - let's merge it in!

Unrelated to above statement, just FYI: On the desktop, I didn't run into it too often, because Swipe on the Desktop browser is buggy as schoettl correctly pointed out, but it's also not a paradigm that I know from any other Desktop app. Therefore, I usually make use of the shortcuts feature to navigate and toggle states. Of course, this is not a valid workaround for two reasons:

1. Shortcuts are only for power users.
2. The shortcuts are buggy - they don't trigger every time (haven't taken the time to investigate enough, yet).

In any case, the change is great and very easily implemented. Thank you @tarnung :pray: 

I'll add contribution to the changelog and merge.

PR #559, opened 2020-11-06
Experimental ui overhaul without inline editing

tarnung:
I wanted to state my opinion on the whole ux issue discussed prominently but not exclusively in #188.
This is a draft pr and it should be considered as part of a discussion.

Since I'm not a ui person, I struggle to weigh pros and cons of different approaches. 
I can articulate what I want from the UI and I might be able to weigh conflicting goals against each other, but I can't really predict which exact gui elements will feel right in the end.
So I just played around and built what I thought i would want without yet having used it extensively.

My Motivation:
- I do not like the inline action bar. I want the main mode to be focussed on browsing the file (collapsing/expanding - having an unhindered view of the document) and the action bar is a distraction to that.
- I want the editing to be more consistent. Inline editing is an exception to the way most edit actions are handeled (in a drawer modal) and it is kind of clunky. On my phone, when i tap edit title or edit description many times the textarea is partially obscured by the onscreen keyboard and I have to scroll it into view first. Also textarea and done button fill up all the screen estate that is left besides the keyboard, so there is no advantage to having it inline anyway.

I did not bring any new elements to the table. I just used the tools that were already in place and used them differently.
Here is what I did:
- no more inline action buttons. Tapping one of the floating action buttons opens up a menu (like the arrow actions button does currently) The existing floating action buttons are rearranged to allow for that. (Agenda is under search like search and tasklist already were.)
- no more inline editing. Editing title, description and tables is done in drawers.
- to mitigate having to open the menu several times when changing multiple things about one header, there is an action bar at the bottom of modals to switch between them.
- i moved the delete header action into the floating edit actions and replaced it with visibility cycling (this has no nesessary connection to the other two points, except reinforcing the focus on browsing the document over editing).
- it's all behind a settings switch in case you want to try it out.

This approach comes with at least one major downside:
- The actions are much less at your fingertips. To reach most actions, an additional tap is needed. This probably is a deal killer for some users. 
- Like the arrow actions, the edit actions are 'global' not visually/spacially tied to one header. You can edit a header that is offscreen, if it is still selected.

Oppinions wildly differ on this topic and there probably is no change that isn't percieved as a detriment by somebody. I'm sure I have not found a silver bullet. Since there is no 'one size fits all' solution, I built this behind a feature flag to explore how feasible it would be to support multiple UI approaches. I've seen that there was discussion before on having a dedicated desktop mode with a different ui than mobile, so this idea is not new either.
Of course I would like to have this in a place where I could test drive it more. I'll probably have to look into running my own dev instance - or at least to get port forwarding working to try it on mobile while locally running it.
It's built out to a pretty usable level but there surely are a few details I didn't consider. I have not yet looked into how it interacts with keyboard shortcuts for example.

Alternatives I tried:
- I have also built a version with the existing action bar permanently visible in a fixed position under the header bar at the top of the screen.
- I thought about having a switch between a browse only mode and editing mode. But I have not built anything in that direction.

Now i can go back to think about multifile support :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,25 @@
 #
 version: 2
 jobs:
-  deploy:
+  deploy-stage:
+    docker:
+      - image: circleci/node:12.13-buster
+    steps:
+      - checkout
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - yarn-packages-{{ checksum "yarn.lock" }}
+      - run:
+          name: Install dependencies
+          command: sudo apt-get update -yq && sudo apt install -yq lftp pandoc gvfs emacs
+      - run:
+          name: Deploy to https://staging.organice.200ok.ch
+          command: 'FTP_USER=${FTP_USER_STAGE} FTP_PASSWD=${FTP_PASSWD_STAGE} ./bin/compile_and_upload.sh'
+      - run:
+          name: Deploy documentation to https://staging.organice.200ok.ch/documentation.html
+          command: 'FTP_USER=${FTP_USER_STAGE} FTP_PASSWD=${FTP_PASSWD_STAGE} ./bin/compile_doc_and_upload.sh'
+  deploy-prod:
     docker:
       - image: circleci/node:12.13-buster
     steps:
@@ -26,10 +44,6 @@ jobs:
   build:
     docker:
       - image: circleci/node:12.13-buster
-      # Specify service dependencies here if necessary
-      # CircleCI maintains a library of pre-built images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
-      # - image: circleci/mongo:3.4.4
 
     working_directory: ~/repo
 
@@ -75,10 +89,6 @@ jobs:
   build-docs:
     docker:
       - image: circleci/node:12.13-buster
-      # Specify service dependencies here if necessary
-      # CircleCI maintains a library of pre-built images
-      # documented at https://circleci.com/docs/2.0/circleci-images/
-      # - image: circleci/mongo:3.4.4
 
     working_directory: ~/repo
 
@@ -111,13 +121,20 @@ workflows:
     jobs:
       - build
       - build-docs
-      - deploy:
+      - deploy-prod:
           requires:
             - build
             - build-docs
           filters:
             branches:
               only: master
+      - deploy-stage:
+          requires:
+            - build
+            - build-docs
+          filters:
+            branches:
+              only: develop
       - push-image:
           requires:
             - build # because of tests
--- a/README.org
+++ b/README.org
@@ -251,6 +251,52 @@ organice implements this customization strategy:
 #+END_EXAMPLE
 
 
+** Themes
+   :PROPERTIES:
+   :CUSTOM_ID: themes
+   :END:
+
+organice implements a strategy pattern, so that adding new themes is
+quite easy. These themes come bundled with organice, each in =light= and
+=dark mode=:
+
+*** Solarized
+
+#+html: <p align="center">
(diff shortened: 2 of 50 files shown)


tarnung:
A short visual overview:

Main view with reordered floating action buttons.
main-view

Add header is the first button under capture.
add-header

Agenda is the first button under search.
agenda

The edit button opens a menu including what previously was in the inline header action bar minus 'add new header' plus 'delete header'. This could be slimmed down by only having one button for the seven main edit modals (see next image).
edit-menu

It's possible to switch between the seven main modals via an action bar at the bottom.
modal

To totally get rid of inline editing, the table editor also needs to be in a modal.
table-editor

tarnung:
This branch is currently running on my [raspberry pi](tarnung.internet-box.ch) in case anyone wants to try it. It also has #556 merged into it.

munen:
First, let me say: Wow, that's **huge** commitment to show a UX proposal!

Secondly, I'd like to say that I like very much where this is going! 💯

I'll quote myself from
https://github.com/200ok-ch/organice/issues/188#issuecomment-570043405:

> I'm not convinced that moving the current ActionBar any other place will solve the issue consistently. I'd argue that the issue is that organice has no clear UX guideline atm. There's a custom top bar, a custom action bar on items, custom notifications ("unpushed changes"), custom drawers and then there's material design in the bottom.
> 
> How about we embrace a well established UX paradigm like material design all the way? We could have a proper material design top bar, burger menu, bottom bar, menus, date pickers, snack bars and so on. There will be plenty space then - and users will find themselves more at home, because it'll be a known UX paradigm.
> 
> The only downside I see: It's a major refactor(; The upside: It's 100% possible to do this one step at a time.

You just went ahead with this and replaced the inline action bar with material design floating action buttons - whilst fixing another major UX issue by harmonizing editing into drawers. With the latter, I also strongly agree. Additionally you did great work by giving the user the ability to switch between the modals directly from the drawer, that makes editing much easier.

Generally speaking, I like the direction very much. And I'm quite impressed by your grit in quickly changing some three-thousand LOC just to show a mockup :rocket:

There's one tiny thing: It's _a lot_ of floating action buttons behind the pencil floating action button. I don't have a better suggestion at this point, it just feels like a lot. 

But that doesn't mean the direction isn't the right way to go. At some point, maybe we'll have another idea on how to divide and conquer those floating action buttons.

Bonus: I've shown your approach to my wife. She's a daily user of organice, as well, but she really is a user - as a psychologist she's not into programming, Emacs, text editors, oldschool UIs or any of that. That makes her my first go-to person to do a litmus test for UX changes. She immediately liked the direction, understood where the actions went and gave her approval. That's really not always the case when she's used to something, so it does say quite a lot.

All in all: Amazing work, this is the right direction. I give a big 👍

---

> This branch is currently running on my raspberry pi in case anyone wants to try it.

Thank you for offering. It looks like you've used up your 3 dev accounts:

image

I can do tests locally without deployment, so I'm good.

tarnung:
> It's a lot of floating action buttons behind the pencil floating action button.

I struggeled with that too. The only alternative I came up with is having one generic edit button that takes you to 'edit title' instead of the seven buttons we have now. But this would add another step to navigate to anything but 'edit title'.

now: 
select header > select 'edit description' in the header action bar
then: 
select header > press pencil floating action > select generic edit button > select 'edit description' in  the modal

Maybe I could try this:
Divide the edit action up and put the sync action in one of the groups to gain the space necessary:
actions opening an editing modal:
- title, description, tags, properties, schedule, deadline, note
actions doing something else:
- delete, narrow, clock, refile, share, sync
The spinning sync indicator could still be present on the main button of the group the sync action is in. It does not make complete sense, since syncing isn't something you do to one header and the grouping of the actions isn't clear cut either. Just an idea that came to mind right now.

Issue #514, opened 2020-10-17
Task list not scrollable on Android


tarnung:
It seems there was a similar issue with Agenda as seen in #65 and #67.

While I can scroll the content of the task list in my desktop browser, I can't on my Android phone. I can see a scroll bar in the task list body, but I doesn't react.

munen:
Can someone with an Android tackle this, please?

I confirm that there were the issues #65 and #67, but I have fixed those on Desktop and iOS. I can't for Android, because I do not own such a device.

munen:
Closed by https://github.com/200ok-ch/organice/pull/561

PR #561, opened 2020-11-08
Feat/scrollable modals

tarnung:
#514
`overflow: none` is not a thing so I don't do anything instead.
I used the bowser function `getPlatformType` instead of the custom query to determine whether organice is running in a mobile browser. Before isMobileBrowser alyways was undefined for me, on windows and android.

I verified that this returns the right result on windows and android.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,17 +4,20 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
-* [2020-11-08 Sun]
+* [2020-11-14 Sat]
+** Fixed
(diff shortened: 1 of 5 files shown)


munen:
@tarnung Sweet, happy to hear that task lists can be scrolled on Android, again.

And, of course, you're right that `overflow: none` is not a thing. That's an oversight on my part.

I've added the comment back and explained the rationale for it a bit more in the commit: https://github.com/200ok-ch/organice/pull/561/commits/699850ba3e1dc2a5e8f443ff910bc06726aa9256

I've also added attribution.

Happy to merge, good work as always! :pray:

munen:
Closes #514

Issue #220, opened 2020-01-10
As a user, when visiting the Agenda view, I want the selected tab to be persisted, so that it can be pre-selected when I visit the next time 


munen:
We have daily/weekly/monthly tabs. Let's assume a user mostly wants to work from the daily view, then just show her this view the next time.

If possible make this behavior generic, so that it can also be used for the refactored Search/Task-List/Refile modals: https://github.com/200ok-ch/organice/issues/211

munen:
Closed by https://github.com/200ok-ch/organice/pull/562

PR #562, opened 2020-11-08
persist agenda timeframe

tarnung:
#220
It's the staight forward solution, not the nice generic functionality @munen imagined.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -4,6 +4,12 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2020-11-21 Sat]
+
+** Added
+   - When going to the Agenda view, the selected tab is persisted - meaning it will be pre-selected when you go to the Agenda next time.
+     - Relevant PR: https://github.com/200ok-ch/organice/pull/562
+
 * [2020-11-20 Fri]
 ** Fixed
    - organice understands =:PROPERTIES:= drawers and smartly parses the values in case one of the values is a timestamp.
--- a/src/actions/base.js
+++ b/src/actions/base.js
@@ -198,3 +198,9 @@ export const activatePopup = (popupType, data) => ({
 export const closePopup = () => ({
   type: 'CLOSE_POPUP',
 });
+
+export const setAgendaTimeframe = (agendaTimeframe) => (dispatch) =>
+  dispatch({
+    type: 'SET_AGENDA_TIMEFRAME',
+    agendaTimeframe,
+  });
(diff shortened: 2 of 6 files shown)


munen (changes requested):
I wouldn't mind this kind of implementation which is less generic. However, at this point it has two open issues:

- `default` for settings is only implemented for `org`, not for `base`, iirc. So, opening the Agenda the very first time doesn't select any of the three tabs.
- more importantly, try what happens when you use the agenda in a private window without logging in(;

tarnung:
> default for settings is only implemented for org, not for base, iirc. So, opening the Agenda the very first time doesn't select any of the three tabs.

I stumbled onto that but then could not reproduce the behaviour. That also explains why the tests fail.

tarnung:
> try what happens when you use the agenda in a private window without logging in

It seems to work perfectly fine. Of course it does not persist anything and always starts out with the weekly agenda.

munen:
> I stumbled onto that but then could not reproduce the behaviour. That also explains why the tests fail.

Once set to localStorage or your settings file, the setting is there. But for the initial load, it's not.

> It seems to work perfectly fine. Of course it does not persist anything and always starts out with the weekly agenda.

So you don't have this behavior?

Video: https://www.dropbox.com/s/4esix81hd7533r0/agenda_crash.gif?dl=1

Resulting crash: image

tarnung:
Now I get the error too. Don't know what I did differently the last time around.
I fixed it by not writing the agenda timeframe into the config since syncing it to the backend is overkill anyway. I realized that `shouldStoreInConfig` was checked when restoring settings, but not when writing them. So all settings were persisted even though only flagged ones were read back in. Which up to now was no problem since all of them were flagged.

We could of course also explicitly check if a sync backend is available:
`if (state.syncBackend.get('client') && window.previousSettingsFileContents !== settingsFileContents) {
          debouncedPushConfigToSyncBackend(state.syncBackend.get('client'), settingsFileContents);
        }`

Either of these approaches fixes the problem you encountered. I chose the first because it allows for more flexibility when storing future settings. But we could do both if you prefer.

tarnung:
> Once set to localStorage or your settings file, the setting is there. But for the initial load, it's not.

That is not correct. If there is no value, the default is used. It's what I can observe when inspecting the redux state. My istincts tell me that I am right and the tests are wrong ;)

But then I once observed the effect of no content displaying in the agenda which I can't explain.

I can fix the tests by also having a default value in their state - which seems reasonable.

munen:
I like it! It works well, and regarding not storing the config, it could even be argued that it's a feature that users can have different default agenda timeframes per device. Personally, I'd likely choose 'day' on my phone, because I'm on the go and it's about the more pressing stuff whereas I might choose 'week' on the desktop, because then I'm on my work machine and there's more stuff that can get done.

The fix works for me and is very reasonable. This is good to go, I'll add it to the changelog and then let's get this gem merged :pray:

PR #563, opened 2020-11-08
default autocrlf & always lf for .org files

tarnung:
#513

the tests expect .org files to have lf line endings so always check out .org files with lf.


munen:
Looks good to me - sane defaults for any bigger project. Thanks for looking it up, @tarnung :+1:

PR #568, opened 2020-11-13
fix repeater for simple timestamps

tarnung:
Repeaters on simple timestamps are broken.
When the todo state changes, the last repeat property is created, a log entry is written and the planning item in the redux state is updated. The part in the title or description that represents the timestamp is not touched though, so while the item moves in the agenda, the change is not represented in text and is not persisted.

This fix connects the timestamps in title and description with their representation in planning items by keeping the id instead of generating a new one. It fixes the issue. I'm quite sure but not 100% certain that having planning items with the same id as parts of title or description won't cause problems. It makes sense, since the items with the same id represent the same thing. If it's preferred I could instead put a timestampId on the planning items and let them have their own unique id.

### DIFF
--- a/src/reducers/org.js
+++ b/src/reducers/org.js
@@ -15,7 +15,6 @@ import {
   extractAllOrgTags,
   extractAllOrgProperties,
   getTodoKeywordSetsAsFlattenedArray,
-  isRegularPlanningItem,
 } from '../lib/org_utils';
 
 import {
--- a/src/reducers/org.js
+++ b/src/reducers/org.js
@@ -1405,6 +1405,23 @@ function updatePlanningItemsWithRepeaters({
       ['headers', headerIndex, 'planningItems', planningItemIndex, 'timestamp'],
       adjustedTimestamp
     );
+
+    // INFO: Active timestamps are now manually updated in place.
+    // Rationale: The active timestamps in title and description are
+    // added to `planningItems` on parse. Since there can be an
+    // arbitrary amount of timestamps it makes sense not to have one
+    // `planningItem` representing the title or the description. We
+    // need to preserve the place of a timestamp in title/description
+    // and we want to have it in a list of `planningItems`. So they
+    // necessarily exist in more than one place. There might be a
+    // cleaner solution where we store the timestamp only in one place
+    // and use references to that place but I don't see any extra
+    // benefit for what would be no negligible refactoring effort.
+
+    // Scheduled / deadline timestamps on the other hand are part of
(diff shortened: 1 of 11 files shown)


munen:
I've added a reducer test case for it which fails on `master`, but works on this branch. I've also confirmed functionality manually.

What's left is to fix the `eslint` warning:

image

src/reducers/org.js
```diff
@@ -1371,10 +1371,39 @@ function updatePlanningItemsWithRepeaters({
   timestamp,
 }) {
   indexedPlanningItemsWithRepeaters.forEach(([planningItem, planningItemIndex]) => {
+    const adjustedTimestamp = applyRepeater(planningItem.get('timestamp'), timestamp);
     state = state.setIn(
       ['headers', headerIndex, 'planningItems', planningItemIndex, 'timestamp'],
-      applyRepeater(planningItem.get('timestamp'), timestamp)
+      adjustedTimestamp
     );
+    switch (planningItem.get('type')) {
```
munen, line 1379: I confirm this works and the new test does, too.

But, I ask myself: Why isn't the header part of the `planningItems` in the first place? I might be mistaken, but it could be that the fix here is more explicit than it needs to be and that there's a bug further up the stack trace.

tarnung: > Why isn't the header part of the planningItems in the first place?

The active timestamps in title and description are added to `planningItems` on parse. Since there can be an arbitrary amount of timestamps it makes sense not to have one `planningItem` representing the title or the description. We need to preserve the place of a timestamp in title/description and we want to have it in a list of `planningItems`. So they necessarily exist in more than one place. There might be a cleaner solution where we store the timestamp only in one place and use references to that place but I don't see any extra benefit for what would be no negligible refactoring effort.

Scheduled / deadline timestamps are part of `rawDescription` but not of the parsed description. These timestamps only exist in one place so changing them there is visible and will be persisted.

tarnung:
I added support for deleting timestamps.
In the process I found a bug in `removePlanningItem`. When the timestamp editor is opened by clicking on a timestamp that does not belong to the currently selected header, `removePlanningItem` tries to remove the item from the wrong header.

src/reducers/org.js
munen: This is all very much correct and a perfect explanation. Thank you :pray: 

I'll add this as a comment to the code, because it's well written and quite informative.

munen:
Great job, again :+1: 

I'll add it to the changelog and will merge it in!

Issue #325, opened 2020-06-05
`file:` links are optimistically opened as folders (when sometimes they are not)


munen:
https://github.com/200ok-ch/organice/pull/318 brought many improvements on the handling of `file:` links.

It adds the logic that if a link cannot be classied, it's optimistically opened as a folder. When that link doesn't go to a folder, the user will see an error like this:

![image](https://user-images.githubusercontent.com/505721/83855146-f71a3e00-a717-11ea-8579-de44ce997f13.png)

PR #570, opened 2020-11-14
EPIC: Multiple files

munen:
This epic includes multiple PRs that are only partially reviewed:

- https://github.com/200ok-ch/organice/pull/550
- https://github.com/200ok-ch/organice/pull/560
- https://github.com/200ok-ch/organice/pull/565
- https://github.com/200ok-ch/organice/pull/574
- https://github.com/200ok-ch/organice/pull/575

Open functional issues:

- [x] CI green
- [x] Optional: Landing test if possible
- [x] Optional: Browser Back button broke
- [x] Optional: liveSync default `on`
- [x] "Sync on application becoming visible" regression. Discussion:
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-738913041
- [x] Loading of `sample` and `changelog`. Discussion:
  - https://github.com/200ok-ch/organice/pull/560#issuecomment-723591882
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-738282502
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-738994164
  - Status: It seems to work for tarnung, but it does not work for munen on Staging. Where's the catch?
- [x] Undo sometimes leaves the Org file blank. Only recovery is 'reload'.
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735440875
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735696083
- [x] (Legacy) capture templates sometimes do not work. Discussion:
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735438317
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735446594
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735654214
- [x] “Sticky messages”
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735701889
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735729280
- [x] Something about capturing is not right: capturing does not refresh clock times & opening agenda after capturing crashes

Things to write tests for:
- [ ] (not) displaying things in Agenda, Search, Task-List with and without file settings
- [ ] jumping to a header in another file from Agenda, Search, Task-List
- [ ] capturing to another file with various capture template settings
- [ ] refiling to another file with various file settings

List of things to document after before/soon after merging:

- [ ] General functionality and awesomeness of this change
  - [ ]  Add to documentation that results are limited to 200 headers
- [ ] UX for developers: How to cope with slowing Redux (https://github.com/200ok-ch/organice/pull/570#issuecomment-735434357)

Follow Up Issues:

- [ ] Configure capture templates to be shown 'global' or in 'buffer'. Discussion:
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735438317
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735446594
  - https://github.com/200ok-ch/organice/pull/570#issuecomment-735654214

### DIFF
--- a/sample.org
+++ b/sample.org
@@ -418,6 +418,19 @@ file back up.
 Generally, when working with distributed Org files, we're recommending
 to put them under version control and to check for bugs and racing
 conditions between clients.
+
+* Multiple files
+organice supports using multiple files.
+
+default behaviour:
+- When you first open a file it is loaded from the backend. At this point the file is persisted to local storage. It will be loaded from there on future application starts. To ensure consistency, whenever a file is opened from the file browser, it will be synced with the backend.
+- Contents of all loaded files are displayed in the agenda.
+- All loaded files are refile targets.
+- Search only shows results from the currently viewed file.
+- Task list only shows tasks from the currently viewed file.
+
+You can adjust these defaults on a file per file basis by creating file settings in the settings menu.
+
 * organice operates completely client side
 You don't log in to organice directly because organice doesn't have a
 back end - it operates completely client side using Dropbox, Google
--- a/sample.org
+++ b/sample.org
@@ -422,7 +422,8 @@ conditions between clients.
 * Multiple files
 organice supports using multiple files.
 
-default behaviour:
+Default behaviour:
+
 - When you first open a file it is loaded from the backend. At this point the file is persisted to local storage. It will be loaded from there on future application starts. To ensure consistency whenever a file is opened it will be synced with the backend.
 - Agenda, Refile, Search and Task List by default show results from the currently viewed file.
 
--- a/sample.org
+++ b/sample.org
@@ -423,7 +423,7 @@ conditions between clients.
 organice supports using multiple files.
 
 default behaviour:
-- When you first open a file it is loaded from the backend. At this point the file is persisted to local storage. It will be loaded from there on future application starts. To ensure consistency, whenever a file is opened from the file browser, it will be synced with the backend.
+- When you first open a file it is loaded from the backend. At this point the file is persisted to local storage. It will be loaded from there on future application starts. To ensure consistency whenever a file is opened it will be synced with the backend.
 - Contents of all loaded files are displayed in the agenda.
 - All loaded files are refile targets.
 - Search only shows results from the currently viewed file.
--- a/sample.org
+++ b/sample.org
@@ -424,12 +424,9 @@ organice supports using multiple files.
 
 default behaviour:
 - When you first open a file it is loaded from the backend. At this point the file is persisted to local storage. It will be loaded from there on future application starts. To ensure consistency whenever a file is opened it will be synced with the backend.
-- Contents of all loaded files are displayed in the agenda.
-- All loaded files are refile targets.
-- Search only shows results from the currently viewed file.
-- Task list only shows tasks from the currently viewed file.
+- Agenda, Refile, Search and Task List by default show results from the currently viewed file.
 
-You can adjust these defaults on a file per file basis by creating file settings in the settings menu.
+You can adjust these defaults on a file per file basis by creating file settings in the [[/settings][settings menu]].
 
 * organice operates completely client side
 You don't log in to organice directly because organice doesn't have a
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -1,6 +1,8 @@
 import React from 'react';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
 
-import { Link, useLocation } from 'react-router-dom';
+import { useLocation } from 'react-router-dom';
 
 import './stylesheet.css';
 
@@ -10,10 +12,11 @@ import TimestampPart from './components/TimestampPart';
 import ExternalLink from '../../../UI/ExternalLink';
 
 import { orgFileExtensions } from '../../../../lib/org_utils';
+import * as orgActions from '../../../../actions/org';
 
 import classNames from 'classnames';
 
-export default ({ parts, subPartDataAndHandlers }) => {
+const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
   let className;
 
   let location = useLocation();
@@ -39,17 +42,13 @@ export default ({ parts, subPartDataAndHandlers }) => {
             // directory.
             target = target.replace(/^\/file\//, '/files/');
           }
-          return (
-            <Link key={id} to={target}>
-              {title}
-            </Link>
-          );
         }
       }
-
-      // Best effort in other cases; the file:// href may or may not work,
-      // but either way it should help show the user what's going on.
-      target = 'file://' + target;
+      return (
+        <span key={id} style={{ textDecoration: 'underline' }} onClick={() => org.setPath(target)}>
+          {title}
+        </span>
+      );
     }
 
     return (
@@ -71,7 +70,7 @@ export default ({ parts, subPartDataAndHandlers }) => {
       normalised = normalised.substr(3);
       dir = dir.match(/(.*)\//)[1];
     }
-    return dir + '/' + normalised;
+    return '/' + normalised;
   };
 
   return (
@@ -173,3 +172,11 @@ export default ({ parts, subPartDataAndHandlers }) => {
     </span>
   );
 };
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    org: bindActionCreators(orgActions, dispatch),
+  };
+};
+
+export default connect(null, mapDispatchToProps)(AttributedString);
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -40,14 +40,16 @@ const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
           if (!uri.match(orgFileExtensions)) {
             // Optimistically assume that the link is pointing to a
             // directory.
+            target = target.replace(/^\/file\//, '/files/');
             return (
-              <Link key={id} to={'/files' + target}>
+              <Link key={id} to={target}>
                 {title}
               </Link>
             );
           }
         }
       }
+      target = target.replace(/^\/file\//, '/');
       return (
         <span key={id} style={{ textDecoration: 'underline' }} onClick={() => org.setPath(target)}>
           {title}
@@ -74,7 +76,7 @@ const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
       normalised = normalised.substr(3);
       dir = dir.match(/(.*)\//)[1];
     }
-    return '/' + normalised;
+    return dir + '/' + normalised;
   };
 
   return (
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
-import { useLocation } from 'react-router-dom';
+import { Link, useLocation } from 'react-router-dom';
 
 import './stylesheet.css';
 
@@ -40,7 +40,11 @@ const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
           if (!uri.match(orgFileExtensions)) {
             // Optimistically assume that the link is pointing to a
             // directory.
-            target = target.replace(/^\/file\//, '/files/');
+            return (
+              <Link key={id} to={'/files' + target}>
+                {title}
+              </Link>
+            );
           }
         }
       }
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -51,7 +51,12 @@ const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
       }
       target = target.replace(/^\/file\//, '/');
       return (
-        <span key={id} style={{ textDecoration: 'underline' }} onClick={() => org.setPath(target)}>
+        <span
+          key={id}
+          style={{ textDecoration: 'underline' }}
+          data-target={target}
+          onClick={() => org.setPath(target)}
+        >
           {title}
         </span>
       );
--- a/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
+++ b/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
@@ -15,6 +15,7 @@ import {
 
 import { format, isPast } from 'date-fns';
 import classNames from 'classnames';
+import { determineIncludedFiles } from '../../../../../../reducers/org';
 
 function TaskListView(props) {
   function handleHeaderClick(path, headerId) {
@@ -133,11 +134,16 @@ function TaskListView(props) {
 
 const mapStateToProps = (state) => {
   const files = state.org.present.get('files');
+  const path = state.org.present.get('path');
+  const allFiles = state.org.present.get('files');
+  const fileSettings = state.org.present.get('fileSettings');
   return {
     // When no filtering has happened, yet (initial state), use all headers.
     headersForFiles:
       state.org.present.getIn(['search', 'filteredHeaders']) ||
-      files.map((file) => file.get('headers')),
+      determineIncludedFiles(allFiles, fileSettings, path, 'includeInTasklist', false).map((file) =>
+        file.get('headers')
+      ),
     todoKeywordSetsForFiles: files.map((file) => file.get('todoKeywordSets')),
   };
 };
--- a/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
+++ b/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
@@ -132,7 +132,7 @@ const mapStateToProps = (state) => {
     // When no filtering has happened, yet (initial state), use all headers.
     headers:
       state.org.present.getIn(['search', 'filteredHeaders']) ||
-      // TODO: this uses only the current file when no search term is entered. Use all of them.
+      // @tarnung TODO: this uses only the current file when no search term is entered. Use all of them.
       new Map().set(path, file.get('headers')),
   };
 };
--- a/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
+++ b/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
-import { Map } from 'immutable';
+import { List } from 'immutable';
 
 import './stylesheet.css';
 
@@ -21,76 +21,71 @@ function TaskListView(props) {
     return () => props.onHeaderClick(path, headerId);
   }
 
-  const { dateDisplayType, onToggleDateDisplayType, headers, todoKeywordSets } = props;
+  const {
+    dateDisplayType,
+    onToggleDateDisplayType,
+    headersForFiles,
+    todoKeywordSetsForFiles,
+  } = props;
 
-  // TODO this uses the todokeywordset of the current file for all files
-  // use each files own todokeywordset instead
-  const planningItemsAndHeaders = headers.map((headersForFile) =>
-    getPlanningItemsAndHeaders({
-      headers: headersForFile,
-      todoKeywordSets,
-    })
-  );
+  const planningItemsAndHeaders = getPlanningItemsAndHeaders({
+    headersForFiles,
+    todoKeywordSetsForFiles,
+  });
 
   return (
     <div className="agenda-day__container">
       <div className="agenda-day__headers-container">
-        {Array.from(planningItemsAndHeaders.entries(), ([path, planningItemsAndHeadersOfFile]) => (
-          <div>
-            <span>{path}</span>
-            {planningItemsAndHeadersOfFile.map(([planningItem, header]) => {
-              const planningItemDate = planningItem
-                ? dateForTimestamp(planningItem.get('timestamp'))
-                : null;
-              const hasTodoKeyword = !!header.getIn(['titleLine', 'todoKeyword']);
+        {planningItemsAndHeaders.map(([planningItem, header]) => {
+          const planningItemDate = planningItem
+            ? dateForTimestamp(planningItem.get('timestamp'))
+            : null;
+          const hasTodoKeyword = !!header.getIn(['titleLine', 'todoKeyword']);
 
-              const dateClassName = classNames('task-list__header-planning-date', {
-                'task-list__header-planning-date--overdue':
-                  hasTodoKeyword && planningItem && isPast(planningItemDate),
-              });
+          const dateClassName = classNames('task-list__header-planning-date', {
+            'task-list__header-planning-date--overdue':
+              hasTodoKeyword && planningItem && isPast(planningItemDate),
+          });
 
-              let planningInformation = <div />;
-              if (planningItemDate) {
-                planningInformation = (
-                  <div className="agenda-day__header__planning-item-container">
-                    <div className="task-list__header-planning-type">
-                      {getPlanningItemTypeText(planningItem)}
-                    </div>
-                    <div className={dateClassName} onClick={onToggleDateDisplayType}>
-                      {dateDisplayType === 'absolute'
-                        ? format(planningItemDate, 'MM/dd')
-                        : customFormatDistanceToNow(planningItemDate)}
-
-                      {!!planningItem.getIn(['timestamp', 'startHour']) && (
-                        <Fragment>
-                          <br />
-                          {format(planningItemDate, 'h:mma')}
-                        </Fragment>
-                      )}
-                    </div>
-                  </div>
-                );
-              }
-
-              return (
-                <div key={header.get('id')} className="agenda-day__header-container">
-                  <div className="agenda-day__header__header-container">
-                    <TitleLine
-                      header={header}
-                      color="var(--base03)"
-                      hasContent={false}
-                      isSelected={false}
-                      shouldDisableActions
-                      shouldDisableExplicitWidth
-                      onClick={handleHeaderClick(path, header.get('id'))}
-                    />
-                    {planningInformation}
-                  </div>
+          let planningInformation = <div />;
+          if (planningItemDate) {
+            planningInformation = (
+              <div className="agenda-day__header__planning-item-container">
+                <div className="task-list__header-planning-type">
+                  {getPlanningItemTypeText(planningItem)}
                 </div>
-              );
-            })}
-          </div>
-        ))}
+                <div className={dateClassName} onClick={onToggleDateDisplayType}>
+                  {dateDisplayType === 'absolute'
+                    ? format(planningItemDate, 'MM/dd')
+                    : customFormatDistanceToNow(planningItemDate)}
+
+                  {!!planningItem.getIn(['timestamp', 'startHour']) && (
+                    <Fragment>
+                      <br />
+                      {format(planningItemDate, 'h:mma')}
+                    </Fragment>
+                  )}
+                </div>
+              </div>
+            );
+          }
+          return (
+            <div key={header.get('id')} className="agenda-day__header-container">
+              <div className="agenda-day__header__header-container">
+                <TitleLine
+                  header={header}
+                  color="var(--base03)"
+                  hasContent={false}
+                  isSelected={false}
+                  shouldDisableActions
+                  shouldDisableExplicitWidth
+                  onClick={handleHeaderClick(header.get('path'), header.get('id'))}
+                />
+                {planningInformation}
+              </div>
+            </div>
+          );
+        })}
       </div>
     </div>
   );
@@ -100,8 +95,18 @@ function TaskListView(props) {
   // on every update of the headers state when not even looking at the
   // Agenda is certainly more inefficient. Hence, we're doing it on
   // every render.
-  function getPlanningItemsAndHeaders({ headers, todoKeywordSets }) {
-    const isTodoKeywordInDoneState = createIsTodoKeywordInDoneState(todoKeywordSets);
+  function getPlanningItemsAndHeaders({ headersForFiles, todoKeywordSetsForFiles }) {
+    const headers = List().concat(
+      ...headersForFiles
+        .mapEntries(([path, headersOfFile]) => [
+          path,
+          headersOfFile.map((header) => header.set('path', path)),
+        ])
+        .valueSeq()
+    );
+    const isTodoKeywordInDoneState = todoKeywordSetsForFiles.map((todoKeywordSets) =>
+      createIsTodoKeywordInDoneState(todoKeywordSets)
+    );
 
     return headers
       .filter((header) => header.getIn(['titleLine', 'todoKeyword']))
@@ -113,7 +118,9 @@ function TaskListView(props) {
         return [earliestPlanningItem, header];
       })
       .sortBy(([planningItem, header]) => {
-        const doneState = isTodoKeywordInDoneState(header.getIn(['titleLine', 'todoKeyword']));
+        const doneState = isTodoKeywordInDoneState.get(header.get('path'))(
+          header.getIn(['titleLine', 'todoKeyword'])
+        );
 
         const timeAsSortCriterion = planningItem
           ? getTimeFromPlanningItem(planningItem)
@@ -125,15 +132,13 @@ function TaskListView(props) {
 }
 
 const mapStateToProps = (state) => {
-  const path = state.org.present.get('path');
-  const file = state.org.present.getIn(['files', path]);
+  const files = state.org.present.get('files');
   return {
-    todoKeywordSets: file.get('todoKeywordSets'),
     // When no filtering has happened, yet (initial state), use all headers.
-    headers:
+    headersForFiles:
       state.org.present.getIn(['search', 'filteredHeaders']) ||
-      // @tarnung TODO: this uses only the current file when no search term is entered. Use all of them.
-      new Map().set(path, file.get('headers')),
+      files.map((file) => file.get('headers')),
+    todoKeywordSetsForFiles: files.map((file) => file.get('todoKeywordSets')),
   };
 };
 
--- a/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
+++ b/src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
@@ -1,6 +1,9 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useEffect } from 'react';
+import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
-import { List } from 'immutable';
+import { List, Map } from 'immutable';
+
+import * as orgActions from '../../../../../../actions/org';
 
 import './stylesheet.css';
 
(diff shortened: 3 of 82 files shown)


munen:
@tarnung FYI, I've opened this PR to keep track of all the PRs that we've partially discussed and that are merged to `develop` so far.

At some point, we will have to think about properly reviewing, testing and merging your awesome changes to `master`. Since the PRs are (necessarily) quite big, I would offer that we do this in a pairing session. We could either do it remotely, or if you're up to it, you could [come over for a day or a weekend](https://zen-temple.net/lambda-zen-temple/introduction/). With the latter, we can also have a chat over coffee or tea or take a hike in the mountains of Glarus together. We don't have to make plans already, I just want to give options in case you're interested in taking them.

tarnung:
@munen Does the autodeploy only work if all tests are green?

I started to fix the tests back up and I'll eventually get there, but I wonder if develop / staging should allow for experimental deploys.

munen:
@tarnung Yes, it does require the tests to be green at the moment. I agree that at this time, it would be helpful if the deployment wouldn’t require it. 

This is untested, but checking the build rules from my phone, it should be enough to comment this line: https://github.com/200ok-ch/organice/blob/76b46ae74f5dd676ec87fe954fbe61c7af46d224/.circleci/config.yml#L133

munen:
It would have worked. But some rules are enforced on building, so for this specific situation, it’s still not deployed. 

9C32FA7D-54E9-445E-84B3-4F9F18C9C0D6

tarnung:
> It would have worked. But some rules are enforced on building, so for this specific situation, it’s still not deployed.

Thanks for checking it out. Tests are green now so we are good for the moment. 
There aren't any new tests yet though.
I created a stupid situation with my recently pushed changes #561. Since the value of `isMobileBrowser` is environment dependent, the snapshot test is red on my machine while it passes on ci.

tarnung:
We discussed the default behaviours for sync in #565.
Does the addition of persisting files to local storage impact this discussion?

If all sync options are off, currently files are loaded from localstorage and they are only ever synced once a manual syncing action is triggered. Up to now, files were always loaded from the backend when they came into view, so you could be sure you're up to date in that moment. The file settings offer the option to have specific files sync on application start, but that is not the same as syncing every time you open a file and it's off by default. (When live sync is on, I trigger a sync when a file is opened.)

Should there be a way to purge a file from local storage? Currently if you once load a file, it will just hang around for ever and it will be loaded on each application start.

Local storage is accessed on a domain basis i think. That could mean if multiple people were to use the same browser for organice, they would be able to access each others files. Maybe this is a reason to only persist files to local storage if that is chosen as an option. (It's a default behaviour at the moment.)

There is also the meta-discussion of how many things we want to offer as settings and which decisions we want to make for the user.


A first draft for the sample file:

> * Multiple Files
>organice supports using multiple files.
>
>default behaviour:
>- A file is loaded form the back end when you first open it. At this point it is persisted to local storage. It will be loaded from there on future application starts. Apart from manually triggering a sync with the backend, there are several options to customize syncing behaviour.
>- Contents of all loaded files are displayed in the agenda.
>- All loaded files are refile targets.
>- Search only shows results from the currently viewed file.
>- Task list only shows tasks from the currently viewed file.
>
>You can adjust these defaults on a file per file basis by creating file settings in the settings menu.

munen:
> Thanks for checking it out. Tests are green now so we are good for the moment.

👍 

> If all sync options are off, currently files are loaded from localstorage and they are only ever synced once a manual syncing action is triggered. Up to now, files were always loaded from the backend when they came into view, so you could be sure you're up to date in that moment. The file settings offer the option to have specific files sync on application start, but that is not the same as syncing every time you open a file and it's off by default. (When live sync is on, I trigger a sync when a file is opened.)

I think, whenever you visit a file, that file should also be loaded from the back-end to ensure consistency. If it is available in localStorage, than the initial load will be from there. Then, the file can be loaded from the back-end. If there's no Internet connection, then that silently fails (or is omitted due to a `navigator.onLine` check). If the file can be loaded from the back-end, the contents in Redux and localStorage are replaced.

> Should there be a way to purge a file from local storage? Currently if you once load a file, it will just hang around for ever and it will be loaded on each application start.

Local storage is accessed on a domain basis i think. That could mean if multiple people were to use the same browser for organice, they would be able to access each others files. Maybe this is a reason to only persist files to local storage if that is chosen as an option. (It's a default behaviour at the moment.)

There is also the meta-discussion of how many things we want to offer as settings and which decisions we want to make for the user.

munen:
> Thanks for checking it out. Tests are green now so we are good for the moment.

👍 

> If all sync options are off, currently files are loaded from localstorage and they are only ever synced once a manual syncing action is triggered. Up to now, files were always loaded from the backend when they came into view, so you could be sure you're up to date in that moment. The file settings offer the option to have specific files sync on application start, but that is not the same as syncing every time you open a file and it's off by default. (When live sync is on, I trigger a sync when a file is opened.)

I think, whenever you visit a file, that file should also be loaded from the back-end to ensure consistency. If it is available in localStorage, than the initial load will be from there. Then, the file can be loaded from the back-end. If there's no Internet connection, then that silently fails (or is omitted due to a `navigator.onLine` check). If the file can be loaded from the back-end, the contents in Redux and localStorage are replaced.

> Should there be a way to purge a file from local storage? Currently if you once load a file, it will just hang around for ever and it will be loaded on each application start.

That's a good question. `localStorage` is limited to ~5mb per domain, that's quite a few Org files, but not an infinite amount. There are certainly a lot of options of handling this:

1. Ignore it for the moment, because 5mb is a lot of Org files.
2. Give the user an interface where she sees all cached files and an option to delete them.
3. Handle it silently by making space when needed. We could actually save all the files until we reach a quota error and then drop the 'oldest' Org file from storage: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API

> Local storage is accessed on a domain basis i think. That could mean if multiple people were to use the same browser for organice, they would be able to access each others files. Maybe this is a reason to only persist files to local storage if that is chosen as an option. (It's a default behaviour at the moment.)

`localStorage` is on a domain basis, that's right. However, if multiple users were using organice on the same browser, they would have to log out first. Logging out currently 'only' invalidates the auth token (because that's the only thing to invalidate atm). In the future, we should clear all of `localStorage`.

> There is also the meta-discussion of how many things we want to offer as settings and which decisions we want to make for the user.

Sync is terribly complicated as it is (parts outlined in https://github.com/200ok-ch/organice/pull/565#issuecomment-727194117). I'd be for not making it more complicated for us or for the user. Having a sane set of defaults is usually good UX and easier to debug/test.

> A first draft for the sample file:

I like it! With the addendum that manually opening a file in the files browser should also reload the file from the back-end as described above.

tarnung:
I changed it so files are now always either downloaded or synced when they are opened.

Signing out now triggers a `localStorage.clear()`. This also provides a way to get rid of obsolete files that still exist in local storage.

tarnung:
I tried one more small refactoring of how static files are loaded but that didn't work out.

I have no more issues that I see as necessary additions to basic multifile support. Therefore the pr is ready for review.
I'm testing it in regular use on staging, but I probably don't use all features.
Things I know I have not looked into:
- links between files
- capture links

munen:
Sweet! I’ll start testing from Staging, then, as well.

munen:
You mentioned that you didn’t look into it, but that’s the first thing I’m stumbling upon: File links don’t work. I stumbled upon it, because I have links to all secondary files from my main file. 

Opening them manually works. I’ll continue testing like that.

munen:
I quickly got stuck with an overlay “nothing to sync” after adding a couple of files to the file settings.  It’s not going away, even if I open an Org file, make a change (see “I pushed changes”) and sync it. 

390DCC0E-1078-4E16-BD01-552491869A88

tarnung:
@munen be careful for now. I think there is a bug giving files loaded from localstorage a new lastSyncAt date which means they may overwrite your changes made on another machine. I'll look ionto it asap.

munen:
Thanks for the heads up!

tarnung:
I think I caught the possibly data corrupting bug yesterday evening.
In the process I discovered that restoring opennessState had problems. It's fixed now.

About file links: When the url conflicts with the path set in redux, redux wins. That allows to change between files by dispatching an action. I suspected that it breaks file links, but on desktop file links open in a new tab so I didn't know whether it breaks any current behaviour.
I'll think about possible solutions.

I never got stuck with a sync message, but they aren't really nice. I just have all syncing files use the one overlay we already have, so if multiple files are synced, you might see different messages in quick succession, each overwriting the last one (or more likely you will only see some of the notifications).

munen:
Nice, I'll keep testing, then!

Regarding file links: They don't open in a new tab for me on Desktop (FF 80, Debian), either. Neither on `master`, nor `develop`. On `master`, I have the same behavior on Desktop and mobile (it works), on `develop` I also have the same behavior, but it does not work(;

Maybe we're talking about different kinds of links. I'm speaking about file links like `[[file:reference.org][Reference]]`.

 For reference, the current 'file links' implementation uses actual anchor tags. The implementation was done in https://github.com/200ok-ch/organice/pull/307 and https://github.com/200ok-ch/organice/pull/318. There's still an open issue with this implementation https://github.com/200ok-ch/organice/issues/325.

If Redux and the links are unhappy with each other, maybe the implementation for file links to use anchor tags can be swapped by using actions.

munen:
Hm.. this time I tried testing on the Desktop instead of on mobile. I quickly a non-disappearing 'nothing to sync', again. This time I remember what I did:

1. I opened file A, whent to settings and added it to the fileset.
2. I repeated step 1 with two more files for a total of three.
3. I openened the agenda and clicked on an item that was from a different file than the one that was currently open.

A 'reload' of the browser made the message go away.

munen:
Hm.. I wonder what happens if when this error message is disabled:

image

It looks to me like it is obsolete. I've tried:

- Disabling it locally.
- Throttled my browser to only @G speed
- Opened big Org file A and made a change to it
- Went to the file explorer, opened big Org file B and made a change to it
- Went to the file explorer, opeend big Org file C and made a change to it
- Went back to the file explorer

Through all that, I watched the network inspector to wait for the calls to Dropbox to finish. Then, in Emacs, I saw all files revert to the actual change.

Sorcery!

Maybe I'm just using a positive edge case. But you have the implementation in mind. Do you think the guard is still required? https://github.com/200ok-ch/organice/blob/master/src/components/Entry/index.js#L160-L163

Imo it was only there, because organice could handle only one open file.

munen:
Have I mentioned how crazy it feels to open files instantly in organice.

More sorcery! :godmode:

munen:
Could it be that due to this super human speed, we're not having some racing conditions on messages? I'm regularly getting this:

image

It's not triggering a bad bug, on clicking it, it's gone.

munen:
I think I found a bug in "file settings".

That's what I did:

- I opened about 7 different Org files and played around in them for a while (yes, I know, not very specific).
- For the first 4 of them, I one of the first things I did was add them to the file settings dialogue
- After opening the seventh file, I wanted to add it to file settings, as well
- When I did hit the 'plus' sign, not the seventh and currently open file was added, but the sixth file.

Here's a recording:

file_settings_bug

This was reproducible, I've:

- Removed the wrong file 'media.org' from the file settings.
- Closed the tab and started a new instance of organice.
- Opened 'media.org', again.
- Went to the file explorer and opened the file I actually wanted to add to the file settings.
- Went to the file settings menu, hit 'plus' and saw 'media.org', again.

munen:
Just a note and not directly related, just a side effect: When adding a couple of big (but real) Org files to search, then my maxed out max Lenovo notebook is starting to become slow, yet noisy.

Maybe, for search, we could limit the actually returned results to something reasonable (like 200). Reading through tens of thousands of headers in search is probably nothing that anybody would do. And if someone wants more, we could add it as a setting.

munen:
What do you think about adding the filename (similarly or exactly) like in Search/TaskList? The Emacs Org mode agenda prepends the filename to each item. When point is on the item, then the path will be shown in the echo area. 

image

munen:
I think I found a bug (or I understand the intention wrong). I have:

- File A in filesettings with option to refile to it.
- File B open.
- On a header in file B, when refiling, I'm getting options to refile to files B, C, D and E, but not for A.
- When I open file A, again, then go back to file B and retry, file B is also in the list. But there's still a couple targets too many: C, D, and E. They are neither currently open, nor did I add them to the filesettings.

image

tarnung:
> Maybe we're talking about different kinds of links. I'm speaking about file links like `[[file:reference.org][Reference]]`.

We are. I was talking about `[[./my-other-file.org][another org file]]`

> If Redux and the links are unhappy with each other, maybe the implementation for file links to use anchor tags can be swapped by using actions.

That's the direction I was thinking about.

tarnung:
> Do you think the guard is still required?
> 
> Imo it was only there, because organice could handle only one open file.

Nope. It's only still there because I usually have live sync enabled so I don't run into it that often. You can already bypass the guard by changing files via selecting an agenda item in another file. I'll take it out entirely.
The only place where the guard would still be useful is for checking if there are unpushed changes when leaving organice. But i don't think we can interfere with closing a browser tab.

tarnung:
> I think I found a bug in "file settings".

It's probably just me not explaining properly how it works.
If you create a file setting, it just takes the first path of a file that doesn't have a file setting already. This does not have to be the currently opened file. You can switch which file the setting is for by choosing another file in the path-dropdown.

It might be more intuitive to prefere the currently opened file as an initial choice when there is no file setting for it yet.

tarnung:
> Maybe, for search, we could limit the actually returned results to something reasonable (like 200). Reading through tens of thousands of headers in search is probably nothing that anybody would do. And if someone wants more, we could add it as a setting.

200 sounds ok :P

tarnung:
> I think I found a bug (or I understand the intention wrong).

It's the second. But that might also be useful to sniff out unexpected / unintuitive behaviour.

For all files that don't have explicit file settings, the default behaviour is used. That default currently is:
- show all files in agenda and refile
- always show currently opened file

I think things work correctly according to this logic. Of course we can change it if another behaviour would be more natural. It's just the defaults that seemed reasonable to me when I had to decide on something.

munen:
>> Do you think the guard is still required?
>> 
>> Imo it was only there, because organice could handle only one open file.

> Nope. It's only still there because I usually have live sync enabled so I don't run into it that often. You can already bypass the guard by changing files via selecting an agenda item in another file. I'll take it out entirely.

Nice!

In 'normal usage' I probably wouldn't run into it a lot, as well. But now I'm testing hat is possible with all the different files(;

> The only place where the guard would still be useful is for checking if there are unpushed changes when leaving organice. But i don't think we can interfere with closing a browser tab.

That would be good to prevent and we can. I think we currently already do, but for reference this is the event that's necessary: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

Here's example code to let the user confirm:

``` js
  window.onbeforeunload = function (e) {
  var message = "Your Org files are still synchronizing. Please wait with closing this window until it's finished. Otherwise you will lose data.",
  e = e || window.event;
  // For IE and Firefox
  if (e) {
  e.returnValue = message;
  }

  // For Safari
  return message;
};
```

munen:
>> I think I found a bug in "file settings".

> It's probably just me not explaining properly how it works. If you create a file setting, it just takes the first path of a file that doesn't have a file setting already. This does not have to be the currently opened file. You can switch which file the setting is for by choosing another file in the path-dropdown.

Aha! Well, then it works as intended(;

I now recall that you explained that, but tbh, I forgot. I was actually going to ask what the dropdown is for, because I often see only one file in there. But now I understand, the dropdown includes all files that have been opened at some point, but are not in the filesettings, yet.

> It might be more intuitive to prefere the currently opened file as an initial choice when there is no file setting for it yet.

Sounds like a good idea. 

Since it's multi-file is such a cool and powerful feature, we'll certainly also write some docs for it (I can also do that at some point). That'll help other people (and future me) from forgetting how it works(;

munen:
>> I think I found a bug (or I understand the intention wrong).

>It's the second. But that might also be useful to sniff out unexpected / unintuitive behaviour.

>For all files that don't have explicit file settings, the default behaviour is used. That default currently is:
>- show all files in agenda and refile
>- always show currently opened file
>
>I think things work correctly according to this logic. Of course we can change it if another behaviour would be more natural. It's just the defaults that seemed reasonable to me when I had to decide on something.

Ok, then I agree: Things work correctly according to this logic(; With some time thinking about it, I also understand why you did it like this. When starting out, it adds the least amount of friction for the user. It's a matter of 'should we white-list or should we black-list'. Generally speaking, both options are valid.

I was expecting the white-list approach, because that's how Emacs works. There's a variable `org-agenda-files` which holds the paths to all files that are in the agenda. And then there's a variable `org-refile-targets` which is a bit more complex as it can be very dynamic and also needs information which headers are acceptable (for example by depth, tag, keyword, regexp, etc).

I would argue that since Emacs uses the white-list approach, organice should do the same. Understandably, the argument can be made for the other approach, as well, so if you have a strong opinion and would really like it to behave the other way around, then I will not oppose it since you are building the feature.

src/components/OrgFile/components/AttributedString/index.js
```diff
@@ -71,7 +70,7 @@ export default ({ parts, subPartDataAndHandlers }) => {
       normalised = normalised.substr(3);
       dir = dir.match(/(.*)\//)[1];
     }
-    return dir + '/' + normalised;
```
munen, line 74: @tarnung I think this introduced a regression. If I'm in a folder `/org/a.org` which includes a link to `[[file:reference.org][Reference]]`, then the file will not be found:

It stays like this:

image
image

Was there a reason to remove the `dir`?

tarnung: Yes, there was. If i set the path in redux, there is no need for the initial `/file/`.
I'm sure I can solve it in a way that works for directories too.

tarnung:
> I would argue that since Emacs uses the white-list approach, organice should do the same.

I have no strong opinion about that, so we can do it how you wish.
So default behaviour should be to not include files anywhere.
Did I get your intentions right?
I always include the currently viewed file in all views (agenda, search, tasklist, refile). Is that a problem?

src/components/OrgFile/components/AttributedString/index.js
munen: :+1: 

I like that you're rewriting it to use React and Redux instead of file links, too. That seems more ideomatic to me.

Great work as always!

munen:
>> I would argue that since Emacs uses the white-list approach,
>> organice should do the same.

> I have no strong opinion about that, so we can do it how you wish.

👍

> So default behaviour should be to not include files anywhere. Did I get your intentions right? I always include the currently viewed file in all views (agenda, search, tasklist, refile). Is that a problem?

In Emacs, 'search' only works for the current file. Agenda and tasklist are pretty much the same as the agenda is very flexible (sidenote: That's something that we might look at, at some point. the organice query language is so powerful that it would work perfectly to build custom agendas). Agenda and refile work only on specified files, not even the currently open file.

So, in Emacs search is the only one that works on the currently open file. Everything else is explicitly configured. However, the agenda can also be opened and restricted to the current buffer. There's many options to open/configure the agenda.

Hence, I think it can be argued that it makes sense to include the currently open file to all views. It's not exactly the same as in Emacs, but it sounds like a reasonable default to me!

munen:
This time, I'm not assuming a bug, but a misunderstanding on my side. So just for clarification, how do I trigger "sync on startup"? I would have assumed that it would download all files in the fileset with that flag when I open a new tab, but that does not happen. Here's a screenshot showing that I have such files configured (on the right) and on the left it shows that a new window makes only one call to Dropbox to fetch the config:  

image

Even if it were a bug, it's not a blocker, because a 'manual sync' does work and download all the files that are in the known to localStorage.

munen:
Commit  a8b67fd  does work for me. Tested on the existing file links in my real Org files.

:+1:

munen:
@tarnung Does refile still work for you? It doesn't work for me at the moment on Staging. Here's Production and Staging side by side trying to do the same (bogus) refile within the same file:

refile-bug

tarnung:
> This time, I'm not assuming a bug

But I guess you found one. I currently load only the files that are not yet present in localStorage. That's not the intended behaviour.

tarnung:
@munen  refile explodes on me if i run it locally. I'm on it.

munen:
461cc23 works for me. All files load on startup! :+1:

munen:
First refile tests within the same and to another file also work :+1: 

Great job :godmode:

munen:
Not a blocker, but I found something that behaves differently for multi-file than for single-file: When I open a big Org file A with all headers closed, and then search for a header, the relevant sub-headers will open up, the searched for header will be selected and will be shown on the screen. This works for Production and Staging.

On Staging, when file A from above is in the 'search list', but I am in file B and do the same search, it still pretty much works, but with a twist: organice switches to file B, the relevant sub-headers open op, the searched for header will be selected, but there's a high likelyhood that I will have to scroll quite a bit up and down and search for the currently selected header.

munen:
I have just seen that capture templates do not work anymore:

06C3182C-1C5F-4A79-BED7-24A2CAF66129

I remember you once said you haven’t looked at that, so this is more a remind to self to test it when done(;

munen:
688d3ea works nicely for the test case where it didn't work before, great job :+1:

munen:
Not a blocker, but I tried debugging the capture template issue and saw that on `develop`, it takes my dev machine about a minute to load the redux store whilst FF is on 100% load. Logging out and back in again whilst ensuring that the fileSet doesn't contain too many files does bring back snappy dev behavior, though.

That's probably something we can add to the dev documentation.

image

munen:
I have found a bug for capture templates, but we could transform it into a feature.

When a capture template has no path at the moment, it will not work. I found the responsible code:

image

But, even when 'fixing' this, it'll crash afterwards.

But going down further that rabbit hole got me thinking. In Emacs, all capture templates have a file associated. The benefit of that is that in Emacs, when you call the capture templates, you can always chose any of them, not just the ones for the current file. That can be super benefitial - for example when you're in your mails and want to add something to the shopping list shared with your wife, you can just do it(;

We could enforce the same in the settings and then display all the capture templates anywhere. That would be pretty swell, imo. What do you think?

munen:
I found an issue with 'undo'. When I have a file, then refile a header to the same file I'm currently on, that works well, now.

When I hit 'undo', then organice doesn't show my Org file anymore, but shows an empty screen. This is reproducibly on Staging and locally on `development` for different headers in different files. There's no error in the console.

Here's how it looks for me:

refile-undo-bug

tarnung:
> When a capture template has no path at the moment, it will not work.

This did work before. It has been in a strange state though were sometimes you have two empty options in the dropdown and only one of them would work. I could remove the empty option, choosing the current file as initial state in the dropdown.

I guess it's feasible to enforce setting a path for every capture template. I can't think of many use cases where you'd want to use the same template in multiple files. 
I think the option to decide which templates are visible in a file is still valuable. I have some capture templates I only use when I'm in my work file for example. Since screen estate for action buttons is limited, I'd prefer to not see them in my other files. 
The current logic where a template is visible in all files per default while allowing you to switch to a whitelist approach works well in my opinion.

munen:
>> When a capture template has no path at the moment, it will not work.

>This did work before. It has been in a strange state though were sometimes you have two empty options in the dropdown and only one of them would work.

That's it! I have two empty options. When choosing the other one, then it works(;

> I could remove the empty option, choosing the current file as initial state in the dropdown.

For setting up a new capture template, that's an option.

Probably I'm running into the issue, because I tried using an existing capture template.

> I guess it's feasible to enforce setting a path for every capture template. I can't think of many use cases where you'd want to use the same template in multiple files.

This seems to contradict what I said before, but I do actually use the same capture template in organice for multiple files. I have several GTD style files (private, 200ok, wife) that all have "Inbox > Tasks" for example.

Having said so, that's certainly a special use-case and not necessarily something that other people do (always hard to assume, though). However, in Emacs capture templates are bound to paths, so in Emacs I can't use the same capture template for the different files, as well(;

> I think the option to decide which templates are visible in a file is still valuable. I have some capture templates I only use when I'm in my work file for example. Since screen estate for action buttons is limited, I'd prefer to not see them in my other files. The current logic where a template is visible in all files per default while allowing you to switch to a whitelist approach works well in my opinion.

That's a good idea. We could add that as a flag 'show globally vs. just in buffer'. I wouldn't have thought about that. In organice, I only have 4 capture templates at the moment. That would work globally. In Emacs, I have 10, though. That would be a little cramped to have them all visible all the time(;

Of course, that's a separate issue from getting this PR ready.

Only getting existing capture templates working again is relevant for this PR.

munen:
The 'undo issue' I had with refiling (https://github.com/200ok-ch/organice/pull/570#issuecomment-735440875) does not seem to be related to refiling. I also have issues undoing other things on Staging where as it works on Production. Here's a demo of the same change (Toggle 'done' on a header) in the same file on Staging vs. Prod:

*Staging*

undo-bug-2

*Prod*

undo-works-on-prod

munen:
To keep track of all relevant open things from this discussion, I have updated the description of this issue with:

- Open functional issues
- List of things to document after before/soon after merging
- Follow Up Issues

I'll try to keep the list updated.

But at this point, I have a good feeling that the current two 'open functional issues' are the last ones. 

NB: I'm currently doing part of a GTD review across many files and projects using organice whilst checking all the diffs this generates, too. Usually, I can only do parts of such a review in organice one file at a time. So I usually do them in Emacs. So far, so good. This PR really is a major milestone for using organice for even more use cases! And the speed is just amazing. :godmode: :rocket:

munen:
I'm still getting the 'nothing to sync' messages on a regular basis. They don't break anything, I can click on them and they go away. But they keep coming. Here's one I got after clicking on an agenda item and jumping to a new file for example:

munen:
It seems to be more general. I also have other “sticky messages” that appear sometimes. For example this one happened after clicking a link to a different Org file:

46329D65-6E1F-4EFC-B26C-F5E31D530F0C

tarnung:
> To keep track of all relevant open things from this discussion, I have updated the description of this issue with:
> 
>     * Open functional issues
> 
>     * List of things to document after before/soon after merging
> 
>     * Follow Up Issues

Thanks! It sure helps to have a list of what's left to do, now that this feature is out of the exploratory phase.

- I have not yet looked into the sync messages
- I have not looked into undo yet
- Seems like the most useful thing to do about capture templates is to leave them the way they are now but ensure they work consistently.

munen:
> Thanks! It sure helps to have a list of what's left to do, now that this feature is out of the exploratory phase.

It's already quite useful and fun to use^^

> - I have not yet looked into the sync messages
> - I have not looked into undo yet

No pressure. I'll focus on testing for the moment, because so far it still bore fruit, but you work on this whenever you want. And I could also help find/fix bugs. We can schedule a screen sharing pairing session anytime if you want to.

> - Seems like the most useful thing to do about capture templates is to leave them the way they are now but ensure they work consistently.

Agreed. This PR is already a huge undertaking from you. The focus is 100% on getting it merged. For that we will:

- Test more (I take care of it).
- Fix regressions (optional pairing session any time).
- Code Review (pairing session).

After that, we'll merge and have a party(;

I've added it to the 'follow up issues'. When we merge this PR, I'll start adding those (well, so far it's one issue^^) as proper GH issues.

tarnung:
Legacy capture templates were not handled correctly, since they had an `undefined` file property instead of the expected `''` default value.

`undo` was able to return to a state before a path was set while not changing the browser path, leading to a blank view. Now when a file is first shown the history is cleared, preventing you to return to that unusable state. This also prevents accidentally undoing changes in a file you are no longer looking at.

tarnung:
I left the sync messages as is for now. Hopefully they don't stick around anymore when they should not. They don't in the cases I could easily reproduce.

There is still room for improvement since in cases where multiple files are synced they still share one message box they all update. But I don't have motivation to build a real clean implementation of multi file sync messages.

munen:
I forgot to add the loading of 'sample' and 'changelog' to the description. We've discussed it before (https://github.com/200ok-ch/organice/pull/560#issuecomment-723591882). I'll add it to the open issues.

tarnung:
...and I'll cross it off, unless you see any problems with it still. I think I fixed that.

munen:
Does it work for you on Staging? It doesn't for me on FF 80, Debian:

image

I made a hard reload to ensure I'm running the latest code.

munen:
I confirm that "(Legacy) capture templates sometimes do not work" works for me, too, now!

munen:
While using Staging for real life use cases and therefore properly testing, I have found a regression compared to Production: "Sync on application becoming visible" does not work. It's the same for me on Desktop and mobile (iOS). Here's a screencast showing that the flag is 'on' for both systems, but it only works on Production:

bug-sync-on-becoming-visible

munen:
It's tricky. When I log out and in, again, I can open the changelog and sample.

But then I add some 5 files to the fileset with different settings, look at the agenda and click on either changelog or sample, and it crashes, again.

tarnung:
> But then I add some 5 files to the fileset with different settings, look at the agenda and click on either changelog or sample, and it crashes, again.

I confess I was not that thorough with my testing. I hope it does that consistently enough to reproduce it. I'll give it a shot this weekend.

munen:
> I confess I was not that thorough with my testing. I hope it does that consistently enough to reproduce it. I'll give it a shot this weekend.

No worries, it's a **huge** PR (both in LOC changed, but more so in usability improvement) at this point. I'm still very much impressed with how fast you got going on this project and what big changes you're introducing. It's a blast using it on Staging every day(;

I hope you're not getting fed up with me posting some regressions, though. Those don't take away any of the greatness of the epic you're introducing - it's next to impossible to do those big changes without any regressions for anyone^^

tarnung:
> I hope you're not getting fed up with me posting some regressions, though. Those don't take away any of the greatness of the epic you're introducing - it's next to impossible to do those big changes without any regressions for anyone^^

I'm not fed up at all, I eventually want to use this feature with a certain level of confidence that things work as they should. So keep 'em coming.

Displaying sample and changelog crashing was a performance issue. I reload static files when they come into view to reset changes to sample. That reloading was done in the render function of Entry on condition that the file is not loaded yet. It worked fine while loading the static file was fast enough to happen between renders. Once there's enough going on, loading was slowed down so much that the next render triggered another load, slowing things down further, ending in an infinite chain of load actions.
I now trigger the load when the button is pressed, so it will only happen once.

sync on becoming visible currently only syncs dirty files. This is not the expected behaviour. But then I don't know what is.
- Should it sync the currently viewed file?
- Should it sync all the loadOnStartup flaged files (plus the currently viewed)?
- Should it sync all files loaded in redux?
I tend to favour option 2. What do you think, @munen ?

munen:
Happy to hear you're still going strong on this feature. I'm also stoked to be working with this feature set on production, soon. Apart from the small quirks, it's been really solid on Staging and a huge improvement in how to use organice.

Cool bonus feature that you reload the sample! That was actually a bit of pain to use before - any changes to the 'doc/tutorial/sample' needed a restart to check them out. Great effort!

> sync on becoming visible currently only syncs dirty files. This is not the expected behaviour. But then I don't know what is.

It's a good question. So, before it synced just the currently loaded file, independend of it being dirty. The rationale is that you might have been away for a while (think mobile app in the background for the day), you or someone else made changes to the file, you open up the app, and organice ensures that you're working with a fresh state.

With that rationale and the new capabilities, I think you're spot with reloading the `loadOnStartup` files :+1:  From a usability perspective, it's the same as loading the app initially. There's just the technical difference that organice itself is already cached.

munen:
`sample` works for me on Staging, now.

The changelog throws a new error, though:

image

tarnung:
I fixed access to the changelog and sample from places other than the header bar.

I recently changed accessing a file in `mapStateToProps` in `Orgfile` to produce an empty default value instead of null when there is no file present. This might have been a poor decision since it has the following consequences:
- Until a file is loaded the could not parse message is shown
- There is no way of telling if there was a parsing error

The offending changes are in the `fix loading static files` commit. I might have to undo them.

munen:
The `changelog` does load from the settings screen now :+1: 

Loading the `sample` from settings does crash now:

image

Loading the `sample` from the header bar works, but it doesn't change the route:

image

tarnung:
> Loading the `sample` from settings does crash now:

Same cause as changelog. Just didn't catch it yet.

> Loading the `sample` from the header bar works, but it doesn't change the route:

It never did.

munen:
Well, this looks very good. 'sync on becoming visible' as well as loading 'sample' and 'changelog' work for me now, too.

This likely just needs a final push to get ready - getting CI green, again, and having a code review. How about we do these things in a pairing session (over Zoom)? Today, my vacation is supposed to start and should go until the 3rd of January. So, apart from the following days, I'm quite flexible:  Today, 24, 28, 29, 31, 01. If any of the days in between work for you, let me know with a time proposal and then let's get this awesome PR ready and merged :fireworks:

tarnung:
@munen I'm usually a bit low energy during the short days this season, so I didn't work on anything much the last few days.
I used staging for a while now and I agree that it's probably in a place where we can wrap it up and merge it. There's always the option to clean up more stuff later (like the sync notifications).
Let's try a pairing session. I propose the 27th in the afternoon. How about 14:00?

munen:
@tarnung No worries whatsoever. First, this is a FOSS project so it's perfectly reasonable to only work on it when you're actually having fun or want to accomplish something that you're going to use yourself. Secondly, you've been productive like crazy lately(; Please, enjoy a well deserved time of cozy recovery!

Apologies for not having answered so far. I've been sick and this is actually the first of the many stacked up messages I'm replying to on the first time on the computer, again^^

27th at 14:00 sounds good. I'll send you a Zoom link via email.

Looking forward to the pairing session! All the best until then.

sample.org
```diff
@@ -418,6 +418,19 @@ file back up.
 Generally, when working with distributed Org files, we're recommending
 to put them under version control and to check for bugs and racing
 conditions between clients.
+
+* Multiple files
+organice supports using multiple files.
+
+default behaviour:
+- When you first open a file it is loaded from the backend. At this point the file is persisted to local storage. It will be loaded from there on future application starts. To ensure consistency whenever a file is opened it will be synced with the backend.
+- Contents of all loaded files are displayed in the agenda.
```
tarnung, line 427: update: default is only current file

sample.org
```diff
@@ -418,6 +418,19 @@ file back up.
...
+- All loaded files are refile targets.
```
tarnung, line 428: check if still the case

tarnung (changes requested):

src/components/OrgFile/components/TaskListModal/components/TaskListView/index.js
```diff
@@ -114,14 +139,20 @@ function TaskListView(props) {
...
 
-const mapStateToProps = (state) => ({
-  todoKeywordSets: state.org.present.get('todoKeywordSets'),
-  // When no filtering has happened, yet (initial state), use all headers.
-  headers:
-    state.org.present.getIn(['search', 'filteredHeaders']) || state.org.present.get('headers'),
+const mapStateToProps = (state) => {
+  const files = state.org.present.get('files');
+  return {
+    // When no filtering has happened, yet (initial state), use all headers.
```
tarnung, line 145: Add to documentation that results are limited to 200 headers

src/reducers/org.js
```diff
@@ -450,44 +444,57 @@ const moveSubtreeRight = (state, action) => {
...
-    headers,
+    sourceHeaders,
     sourceHeaderId
   );
-  let targetHeaderIndex = indexOfHeaderWithId(headers, targetHeaderId);
+  let targetHeaderIndex = indexOfHeaderWithId(targetHeaders, targetHeaderId);
 
   // Do not attempt to move a header to itself
-  if (sourceHeaderIndex === targetHeaderIndex) return state;
+  // TODO: this should be handeled by search not showing the selected header on refile
```
tarnung, line 482: delete comment

PR #573, opened 2020-11-16
Fix DOMAIN variable example

edgar-vincent:
DOMAIN should only contain the domain name, without a protocol prefix.


munen:
Looks good to me^^

@edgar-vincent Welcome to the organice contributors list :pray: Your contribution was already in, but with this commit, you're also showing up as a contributor :rocket:

edgar-vincent:
Thank you munen :)

PR #578, opened 2020-11-20
fix: Parser does not swallow property values

munen:
organice understands `:PROPERTIES:` drawers and smartly parses the values in case one of the values is a timestamp. However, parsing _all the values_ and saving the parsed result in any case will lead to wrong results. Most values of properties are just plain text and non-interactive things in Org mode.

Hence, before this fix, we'd have problems like:

image

Because organice believed this to be text with `_` which yields underlined text.

### DIFF
--- a/test_helpers/fixtures/all_the_features.org
+++ b/test_helpers/fixtures/all_the_features.org
@@ -20,7 +20,10 @@
 
 # Complex TODO with priority, active timestamp, link to an email and a tag
 **** TODO [#B] <2019-05-02 Thu> Some Person [[mu4e:msgid:CAN6MX5o-i9koRnBXGOFQ-wLUhp82mGCwFSxo0b2i0+TROYNZkg@mail.gmail.com][Re: HÖCHENSCHWAND 05.Februar 2019]] :200ok:
-
+*** Header with =CUSTOM_ID= property
+    :PROPERTIES:
+    :CUSTOM_ID: link_to_me
+    :END:
 *** Planning items
 
 **** TODO I am scheduled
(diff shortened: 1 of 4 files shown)


test_helpers/fixtures/all_the_features.org
```diff
@@ -102,7 +105,7 @@ qui dolorem eum fugiat quo voluptas nulla pariatur?"
    CLOCK: [2019-11-11 Mon 13:12]--[2019-11-12 Tue 13:12] => 24:00
    CLOCK: [2019-10-13 Sun 12:15]--[2019-10-13 Sun 14:12] =>  1:57
    :END:
-** A messy headline with trailing whitespace                    
+** A messy headline with trailing whitespace
```
schoettl, line 108: Ups, did you remove the trailing whitespace here?

munen: Nope, Emacs did(;

Thanks for the quick review. I'll revert this of course.

Haven't read through the diff, yet. Only fixed the issue at hand. I would have hoped to also catch it, but you were blazingly fast. Thanks a bunch :pray:

schoettl: Haha, I'm faster than all tests :D

munen: Really impressive! :rocket: :godmode:

munen:
Btw, to all the readers: Sorry I'm doing this PR before reviewing/commenting on others.

I'm still working like mad on our new product, which is live now, but the customer has daily new change requests that need to be deployed asap^^ And this bug actually tripped me up in our companies project management Org file, so I needed to fix it quickly.

PR #590, opened 2020-11-29
Set up editor indentation correctly

aspiers:
This ensures that many editors can automatically configure themselves to the right indentation style to be consistent with this codebase, as per https://editorconfig.org/

----

#


munen:
Sure, adding editorconfig is a nice idea. I'll happily pull your changes in 🙏

Issue #594, opened 2020-11-30
Change the width of the main container


CharlesGueunet:
On Firefox especially, the main container is ridiculously thin, see my screenshot:

![image](https://user-images.githubusercontent.com/5412579/100670578-5f607a00-335f-11eb-82dc-1132135d6a85.png)

Even if the main app is larger on Chrome, it still takes less than the fourth of the full screen which is not optimal

I think it could be nice to have a way to increase the width of the main app, or maybe just a full-width mode that would expend it horizontally as much as possible.

For now, I use this app in a Chrome instance that takes only a small vertical space on my screen and serves as a sidebar of my main Firefox. It works but I would prefer having Firefox only.

CharlesGueunet:
With a really simple patch, I have the following result

![image](https://user-images.githubusercontent.com/5412579/100729766-79cf3d80-33c9-11eb-96ae-06578d79affb.png)

here is the patch:

```patch
diff --git a/src/base.css b/src/base.css
index 77a91ea..1678191 100644
--- a/src/base.css
+++ b/src/base.css
@@ -14,7 +14,7 @@
   .action-drawer-container,
   .drawer-inner-container,
   .loading-indicator {
-    max-width: 35em;
+    /* max-width: 35em; */
     margin-left: auto;
     margin-right: auto;
   }
```

munen:
Hi @CharlesGueunet 

Thank you for checking out organice!

I understand what you're saying, but your change is a little bit oversimplified, because it basically just reverts a semantic change that we did last year:

![image](https://user-images.githubusercontent.com/505721/100730312-30332280-33ca-11eb-9553-0a7e42df5c2e.png)

If you want flexibility in the width, I propose to add this as a setting where the user can choose a width. Looking forward to your PR! Let me know if you need help.

munen:
As for the ridicioulus width in Firefox, could you check out how that happens? Because it's supposed to look like this on Firefox:

Never should the body of the Org file be more narrow then the floating action buttons.

![image](https://user-images.githubusercontent.com/505721/100730607-8dc76f00-33ca-11eb-94bb-594f8e42cc36.png)

CharlesGueunet:
Ny screenshot comes from a Firefox 83.0 on Artix Linux (Arch Linux).
The problem may come from my screen as it has a native resolution of 3840x2160 but I use it with a resolution of 1920x1080.

[...]

PR #598, opened 2020-12-12
WIP: Add the wide option

CharlesGueunet:
This PR is a WORK IN PROGRESS, I still need to implement the dynamic css rule.

This PR address #594 , it adds a new option to allow wide editor. This make this project more usable on desktop (as it seems it is one of its original goal ;) )

### DIFF
--- a/src/base.css
+++ b/src/base.css
@@ -1,7 +1,7 @@
 @import 'colors.css';
 
 /* Tablets and bigger are constrained to a max width and are centered */
-@media (min-width: 767px) {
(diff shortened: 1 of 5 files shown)


src/base.css
```diff
@@ -10,6 +10,7 @@
     box-shadow: 0 2em 2em 1.5em var(--base1-soft);
   }
 
+  /*
```
CharlesGueunet, line 13: this is temporary, for my personal use while the final fix is not available.

PR #604, opened 2020-12-29
fix: Listen to browser buttons

munen:
As a user, when going 'back' or 'forward' in the history, I want the shown Org file and the URL to match.


munen:
Closing, because a rebased version of this has already been merged in #605.

PR #633, opened 2021-03-02
adding the instructions for webdav server in go

MisterY:
There's a fairly simple WebDAV server in go that may be easier to setup. And it's multi-platform!

### DIFF
--- a/README.org
+++ b/README.org
@@ -377,9 +377,13 @@ feature that does not need synchronization, you can skip this step.
 
 **** WebDAV
 
-organice has support for WebDAV and ships with a Docker container with
-a WebDAV server based on Apache. You can make use of that and use this
-WebDAV back-end for local development.
+organice has support for WebDAV and you can make use of that for local development.
+We offer instructions for two options for setting up WebDAV servers but you
+may use any other standard solution. 
+
+- Option 1: A Docker Container with Apache
+  organice ships with a Docker container with a WebDAV server based on Apache.
+- Option 2: A WebDAV server in go (Docker image also available)
 
 Having said that, if you're a Dropbox or Google Drive user, then it's
 convenient to have working setups for either of them if you want to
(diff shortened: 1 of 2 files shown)


README.org
```diff
@@ -377,17 +377,22 @@ feature that does not need synchronization, you can skip this step.
...
 **** WebDAV
 
-organice has support for WebDAV and ships with a Docker container with
-a WebDAV server based on Apache. You can make use of that and use this
-WebDAV back-end for local development.
+organice has support for WebDAV and you can make use of that for local development.
+We offer instructions for two options for setting up WebDAV servers but you
+may use any other standard solution. 
+
+- Option 1: A Docker Container with Apache
```
schoettl, line 384: These two lines are collapsed to one. Can you make two sentences out of it or just use:

1. A Docker container w Apache: …
2. A WebDAV server in go...

README.org
```diff
@@ -377,17 +377,22 @@ feature that does not need synchronization, you can skip this step.
...
+  organice ships with a Docker container with a WebDAV server based on Apache.
+- Option 2: A WebDAV server in go (Docker image also available)
 
 Having said that, if you're a Dropbox or Google Drive user, then it's
```
schoettl, line 388: Oh, if you're just editing this file, you could ~strike~ the "or Google Drive" since support has been dropped.

schoettl (changes requested):
Looks good to me. Thanks for contributing! I just added two comments.

README.org
MisterY: :O

Thanks for the feedback!

munen:
NB: Regarding support of Google Drive: Support has not been dropped as the integration itself still generally works. The support is limited, though. Hence, there's this disclaimer when logging in:

image

At any point in time, one of the maintainers might want to invest some more time and either completely drop Google Drive or fix the integration. Until that time, I'd say organice has partial support for Google Drive.

MisterY:
ok

Issue #277, opened 2020-04-09
org parser ignores timestamps with max repeat period


cpbotha:
**Describe the bug**

From the [Orgmode habits manual](https://orgmode.org/manual/Tracking-your-habits.html), Orgmode supports timestamps like this:

```
SCHEDULED: <2009-10-17 Sat .+2d/4d>
```

which means that it should repeat every two days but at most every four days.

There's nothing organice can do to really apply these timestamps (you repeat them the same as always, using the `.+2d`, the `/4d` is an additional piece of information used by for example orgmode habits).

However, currently that `/4d` causes the whole timestamp to be missed and ignored completely by the rest of the organice logic.

See in this screenshot that the timestamp is not clickable:

![image](https://user-images.githubusercontent.com/937871/78926492-72a39a00-7a9d-11ea-80fb-5e94f4cd2362.png)

When I remove the `/4d`, all is good:

![image](https://user-images.githubusercontent.com/937871/78926537-87802d80-7a9d-11ea-9125-b94317f5216d.png)

**To Reproduce**

Create a heading like this:

```
* TODO H: Running
  SCHEDULED: <2020-04-05 Sun .+2d/4d>
```

Toggle it to `DONE`.

**Expected behavior**

As with other scheduled items with repeats, the item should return to the `TODO` state with the scheduled date moved 2 days into the future from the done date, because `.+`.

What happens instead, is that organice toggles it to done, and that's it.

**Possible solution**

The regexes and other logic should be tweaked to recognize these timestamps, and then simply to disregard the `/4d` bits. Then everything should work as expected.

A few minutes later: Ok, if I add an additional `(\/\d+[hdwmy])?` after the first hdwmy-group, it fixes this specific issue.

A few more minutes later: It has to be a non-capturing group, so nothing else breaks. With `(?:\/\d+[hdwmy])?` inserted right after the first `([hdwmy])` all tests still pass. However, because that group is not stored in any of the parser data structures, it will get lost.

@DanielDe  and @munen I think I could perhaps fix `timeStampFromRegexpMatch()`, but `parseMarkupAndCookies()` along with its monster regex *really* needs a bit of documentation to help random contributors (like me) to figure out how they work.

munen:
Hi @cpbotha 

Thanks for opening this issue. A couple of things:

[...]

PR #674, opened 2021-05-10
Parse and preserve repeater deadlines #277

tomonacci:
This PR aims to make Organice recognize timestamps with "repeater deadlines" (not sure what the official name is), the `/4d` part of `<2009-10-17 Sat .+2d/4d>`, before org-parser is integrated into Organice. A nongoal is to make repeater deadlines editable in the pretty timestamp UI (which shouldn't be the end of the world for most cases because you can edit timestamps manually).

Explanation of the changes:

- The timestamp structure defined in src/lib/timestamp.js now includes repeater deadlines so that they are preserved across edits and TODO state toggles.
- In the parser, `timestampRegex` is modified to recognize repeater deadlines. This increases the number of captures in this regexp by four.
- To reflect the change in the number of capture groups, some indices are shifted by 4 or 8, depending on how many timestamps are included in the regexp in question.
- I also deleted the occurrences of timestamp regexps in `beginningRegexp` and put them directly in `markupAndCookieRegex` (but this time reusing `timestampRegex` instead of repeating the whole expression).

If tests are required to merge, suggestions re what kind of tests and where to put them would be appreciated. When I run `yarn test`, all but one test cases pass.

```
  ● Render all views › Org Functionality › Renders everything starting from an Org file › TaskList › orders tasks for an Org file

    expect(received).toMatchSnapshot()

    Snapshot name: `Render all views Org Functionality Renders everything starting from an Org file TaskList orders tasks for an Org file 1`

    - Snapshot
    + Received

    @@ -35,11 +35,10 @@
            value=""
          />
        </div>
        <div
          class="task-list__headers-container"
    -     style="overflow: auto;"
        >
          <div
            class="agenda-day__container"
          >
            <div

      402 |           fireEvent.click(getByTitle('Show task list'));
      403 |           const drawerElem = getByTestId('drawer');
    > 404 |           expect(drawerElem).toMatchSnapshot();
          |                              ^
      405 |         });
      406 |
      407 |         test('search in TaskList filters headers (by default only with todoKeywords)', () => {

      at Object.<anonymous> (src/components/OrgFile/OrgFile.integration.test.js:404:30)

 › 1 snapshot failed.
```

This test case fails even on master. I'm wondering if that's the case for others as well? [edit] Apparently at least it isn't the case for CircleCI...

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -5,6 +5,17 @@ All user visible changes to organice will be documented in this file.
 When there are updates to the changelog, you will be notified and see
 a 'gift' icon appear on the top right corner.
 
+* [2021-05-16 Sun]
+** Added
+   - Parse and preserve habit timestamp ranges
+     - A timestamp may have minimum and maximum ranges specified by
+       using the syntax =.+2d/3d=, which says that you want to do the
+       task at least every three days, but at most every two days.
+     - Upstream documentation:
+       https://orgmode.org/manual/Tracking-your-habits.html
+     - Relevant PR: https://github.com/200ok-ch/organice/pull/674
+     - Thank you [[https://github.com/tomonacci][tomonacci]] for the PR🙏
+
 * [2021-05-13 Thu]
 
 ** Changed
--- a/sample.org
+++ b/sample.org
@@ -184,6 +184,23 @@ Try tapping on the timestamps below to get a feel for the editor:
 [2018-09-17 Sun 10:00-11:30]
 
 <2018-09-17 Sun>--<2018-09-25 Tue>
+
+** Habit tracking
+
+Org has the ability to track the consistency of a special category of
+TODO, called "habits." From the [[https://orgmode.org/manual/Tracking-your-habits.html][upstream doc]]: A timestamp may have
+minimum and maximum ranges specified by using the syntax =.+2d/3d=,
+which says that you want to do the task at least every three days, but
+at most every two days.
+
+Example timestamp: <2009-10-17 Sat .+2d/4d>
+
+organice supports parsing and preserving the minimum/maximum range
+timestamps. Alas, only the minimum range is handled as a [[https://orgmode.org/manual/Repeated-tasks.html][repeated
+task]]. There's no UI for the maximum range, it can be edited as raw
+text. Also, there is no UI to show how well a habit has been
+exercised.
+
 * Automatic/Implicit links
 
 organice recognizes various types of hyperlinks automatically which Emacs Org mode would not necessarily do. That makes sense, because mobile devices, or browsers, enable a different feature set.
--- a/src/lib/parse_org.js
+++ b/src/lib/parse_org.js
@@ -7,7 +7,7 @@ import _ from 'lodash';
 // TODO: Extract all match groups of `beginningRegexp` (for example
 // like `emailRegexp`), so that they can be documented and are less
 // unwieldly.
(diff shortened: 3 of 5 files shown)


munen:
This PR looks great! Thank you for further cleaning up the `beginningRegexp` while building a new feature. Nice application of the pathfinder principle 🙏

As for the tests, organice unfortunately little to no coverage of timestamps. This feature precedes the rigorous testing practice that we have only added later.

The right place to add such tests would be around here: https://github.com/200ok-ch/organice/blob/57b364860993de7e3d3c47d3bdcb6f87c4b16ac4/src/lib/parse_org.unit.test.js#L74

org-parser, on the other hand, has quite a few timestamp tests, which could be used for inspiration: https://github.com/200ok-ch/org-parser/blob/master/test/org_parser/parser_test.cljc#L346

If you can write tests akin to those from org-parser, that would be nice. Since we have no regression tests for timestamps and changing the parser can have substantial consequences, this will have to be done before the PR can be merged. Of course, if you don't have the time, it's not solely on you(; If you prefer, I can take over the PR and write such tests. Let me know what works for you.

As for the red test on your machine, that's curious. On `master` on my machine and on CI, the tests are green. I think we had this before with a developer on Windows iirc. Are you by any chance using Windows or a Mac? If so, maybe we can use this as a chance to make sure the tests are green for all operating systems. Otherwise, it would still be prudent to find the discrepancy making the test red for you, of course.

In the meantime, thank you, again 🙏

tomonacci:
@munen Thank you for the review. I added some timestamp tests in the file you mentioned (although not exactly at the place you suggested). Hopefully it matches what you had in mind... I made two commits for the tests, one with repeater deadlines and one without. I rebased them so that you can checkout `HEAD~2` first and confirm that the tests without repeater deadlines work on master (effectively). While I was at it, I rebased the whole branch on top of the current master. The content of the commit you already reviewed (now the second of the three) hasn't changed.

As for the failing test case, I think it has something to do with me running this on a Mac as you have correctly guessed. It seems that `isMobileBrowser` https://github.com/200ok-ch/organice/blob/ce175d3b28db197f56ef698a20381522a1b42852/src/lib/browser_utils.js#L22-L25 returns true on my machine where it should be (?) returning false. Indeed, making this function `return false` makes all tests pass. I don't plan to dig further soon so hopefully we can resolve this outside of this PR.

munen:
Great job🙏

That's quite sufficient for regression testing timestamps. I'm very happy to pull this in. Before pulling it in, I'll add it to the changelog and documentation.

There's one small thing that I will change in the code: 

``` javascript
function testTimestampText(text, expectedFirstTimestamp, expectedSecondTimestamp?)
```

I suppose you marked the second parameter as optional with the `?`. As far as I know, this is not standard syntax. I tried looking it up and couldn't find anything. Probably you know it from TypeScript? Interestingly it doesn't throw an error in JS, too. Tbh, I saw it, because my linter marked it as red.

As for the failing test, we certainly can table it for the moment. It's interesting that `browser.getPlatformType()` will return different results on different operating systems. That sounds like a bug that we could address upstream.

munen:
@tomonacci I tried to document the functionality to the best of my ability. As far as I can tell, the change relates solely to tracking habits (https://orgmode.org/manual/Tracking-your-habits.html) while repeated tasks (https://orgmode.org/manual/Repeated-tasks.html) were already fully supported. If you'd like to suggest a change in the documentation, feel free to ping me or open a PR.

Thanks, again, it's a very nice addition :pray:

tomonacci:
@munen Thanks for the second review, the fixes and merging! Yes, that `?` is TypeScript-ism... I use TS at work and quite like it, any chance we can start using it in Organice? 😄 The docs you added LGTM in content; it seems that formatting is a bit off when you open it in Organice, however. It could probably be fixed by removing the hard-wrap and putting everything on one line.

munen:
@tomonacci 

> Yes, that `?` is TypeScript-ism... I use TS at work and quite like it, any chance we can start using it in Organice? smile 
*hihi*, nice suggestion(; Personally, I'd love to use Clojure for organice, too^^

It boils down to this: For organice, I'm trying to enable as many contributors as possible by only employing the 'best of breed'™ and 'most popular'™ programming language and libraries whilst trying to keep the number of libraries to a minimum.

So, I do understand that many people like TS and I'm happy for them. There's other good languages, too. JS is the one and only that everyone kind of can write, though. Maybe that'll change when browsers directly implement TS. Then, we can open this discussion to the floor(;


> The docs you added LGTM in content; it seems that formatting is a bit off when you open it in Organice, however. It could probably be fixed by removing the hard-wrap and putting everything on one line.
Yes, thanks for reminding me. organice does not reflow hard-wrapped text - A bit like hard-wrapped emails don't show up nicely in some GUI MUAs (unless `format=flowed` is set correctly). That's something that org-parser will improve on, as well.

For the time being, I have put a little fix in place for content in changelog.org and sample.org: https://github.com/200ok-ch/organice/pull/688

Issue #266, opened 2020-03-12
Ability to select the day a week starts with


fkohrt:
**Is your feature request related to a problem? Please describe.**
The agenda currently always starts the week on a Sunday.

**Describe the solution you'd like**
An option in the settings that let's the user choose on what day he wants the week to start.

schoettl:
That would be a nice feature. Can you work on this?

Otherwise, please see our contributing guidelines. Unless someone commits to work on the issue, we will close it because it's not a bug.

schoettl:
You can re-open the issue if you have time to work on it. But we want to keep the number of issues manageable.

munen:
:+1:

PR #676, opened 2021-05-12
Implement an equivalent of org-agenda-start-on-weekday #266

tomonacci:
WISOTT. One caveat is that I changed the default start of the week to Monday from Sunday to align with the default in Org to minimize surprise for people coming from Org (at the expense of surprising longtime Organice users).

Controls in the settings: 
Screen Shot 2021-05-12 at 3 18 05 AM

### DIFF
--- a/src/components/OrgFile/components/AgendaModal/index.js
+++ b/src/components/OrgFile/components/AgendaModal/index.js
@@ -18,6 +18,7 @@ import {
   addDays,
   addWeeks,
   addMonths,
+  getDay,
   subDays,
   subWeeks,
   subMonths,
@@ -38,10 +39,13 @@ function AgendaModal(props) {
     agendaTimeframe,
     agendaDefaultDeadlineDelayValue,
     agendaDefaultDeadlineDelayUnit,
+    agendaStartOnWeekday,
   } = props;
 
   const [selectedDate, setSelectedDate] = useState(new Date());
   const [dateDisplayType, setDateDisplayType] = useState('absolute');
+
+  const weekStartsOn = agendaStartOnWeekday === null ? getDay(selectedDate) : agendaStartOnWeekday;
 
   function handleTimeframeTypeChange(agendaTimeframe) {
     props.base.setAgendaTimeframe(agendaTimeframe);
@@ -93,7 +97,7 @@ function AgendaModal(props) {
       case 'Day':
         return format(selectedDate, 'MMMM do');
       case 'Week':
-        const weekStart = startOfWeek(selectedDate);
+        const weekStart = startOfWeek(selectedDate, { weekStartsOn });
         const weekEnd = addWeeks(weekStart, 1);
         return `${format(weekStart, 'MMM do')} - ${format(weekEnd, 'MMM do')} (W${format(
           weekStart,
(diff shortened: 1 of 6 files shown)


src/components/OrgFile/components/AgendaModal/index.js
```diff
@@ -172,12 +176,14 @@ const mapStateToProps = (state) => {
...
   const allFiles = state.org.present.get('files');
   const fileSettings = state.org.present.get('fileSettings');
+  const agendaStartOnWeekday = state.base.get('agendaStartOnWeekday');
   return {
     files: determineIncludedFiles(allFiles, fileSettings, path, 'includeInAgenda', false),
     todoKeywordSets: file.get('todoKeywordSets'),
     agendaTimeframe: state.base.get('agendaTimeframe'),
     agendaDefaultDeadlineDelayValue: state.base.get('agendaDefaultDeadlineDelayValue') || 5,
     agendaDefaultDeadlineDelayUnit: state.base.get('agendaDefaultDeadlineDelayUnit') || 'd',
+    agendaStartOnWeekday: agendaStartOnWeekday == null ? 1 : +agendaStartOnWeekday,
```
schoettl, line 186: Why not `===`? I thought it's best practice to use it instead of `==`. Why the `+`? To convert to integer?

How far do we trust the state store in terms of data integrity? The existing lines above this one don't check the types. So I think we better stick to that scheme. @munen what do you think?

src/components/Settings/index.js
```diff
@@ -326,12 +347,14 @@ const mapStateToProps = (state) => {
...
   // defaults which get loaded on an initial run of organice, look at
   // `util/settings_persister.js::persistableFields`.
+  const agendaStartOnWeekday = state.base.get('agendaStartOnWeekday');
   return {
     fontSize: state.base.get('fontSize') || 'Regular',
     bulletStyle: state.base.get('bulletStyle'),
     shouldTapTodoToAdvance: state.base.get('shouldTapTodoToAdvance'),
     agendaDefaultDeadlineDelayValue: state.base.get('agendaDefaultDeadlineDelayValue') || 5,
     agendaDefaultDeadlineDelayUnit: state.base.get('agendaDefaultDeadlineDelayUnit') || 'd',
+    agendaStartOnWeekday: agendaStartOnWeekday == null ? 1 : +agendaStartOnWeekday,
```
schoettl, line 357: Same question as above, here.

schoettl (commented):
Nice! I like the new default Monday, especially if it's the default in org mode.

It would be great if you could add tests to prove the settings control works.

munen:
Hi @tomonacci

Thank you for your contribution :pray: 

@schoettl Thank you for the code review :pray: 

I've looked through the code. It works well, but I'd make some changes. I'll pull PR into the `develop` branch and make a new PR from there against `master` with changes that I propose. Feel free to make comments there.

I haven't written a test, yet. If you can do so, as @schoettl suggested, that would be nice^^

src/components/OrgFile/components/AgendaModal/index.js
tomonacci: Right, triple equals is best practice in general but some consider double equals against null and undefined to be an exception since it’s a convenient way of eliminating both at once. Yes + is for number conversion (maybe this was too hacky)

I’m type checking here because falseness check interferes with 0... I’ll comment more on that in #677

tomonacci:
Thanks for the review @schoettl and @munen! Regarding tests, let me know if you have a specific existing test case for a setting control that you want me to mimic the format/style of, otherwise I’ll look around and try my best

PR #677, opened 2021-05-12
Refactor: #266 Org agenda start on weekday

munen:
## Changes to https://github.com/200ok-ch/organice/pull/676

- Bugfix: TabButton keys were not unique (twice an `S`), that makes React throw many errors
- Ideomatic organice default value
- Tooltips for Settings

## TODO

- [x] Get some feedback from @tomonacci
- [x] Write an integration test
- [x] Write a changelog entry
- [ ] Merge to `master`

### DIFF
--- a/src/components/OrgFile/components/AgendaModal/index.js
+++ b/src/components/OrgFile/components/AgendaModal/index.js
@@ -176,13 +176,14 @@ const mapStateToProps = (state) => {
   const file = state.org.present.getIn(['files', path]);
   const allFiles = state.org.present.get('files');
   const fileSettings = state.org.present.get('fileSettings');
+  const agendaStartOnWeekday = state.base.get('agendaStartOnWeekday');
   return {
     files: determineIncludedFiles(allFiles, fileSettings, path, 'includeInAgenda', false),
     todoKeywordSets: file.get('todoKeywordSets'),
     agendaTimeframe: state.base.get('agendaTimeframe'),
     agendaDefaultDeadlineDelayValue: state.base.get('agendaDefaultDeadlineDelayValue') || 5,
     agendaDefaultDeadlineDelayUnit: state.base.get('agendaDefaultDeadlineDelayUnit') || 'd',
-    agendaStartOnWeekday: state.base.get('agendaStartOnWeekday') || 1,
+    agendaStartOnWeekday: agendaStartOnWeekday == null ? 1 : +agendaStartOnWeekday,
   };
 };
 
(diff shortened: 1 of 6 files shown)


src/components/OrgFile/components/AgendaModal/index.js
```diff
@@ -176,14 +176,13 @@ const mapStateToProps = (state) => {
...
   const fileSettings = state.org.present.get('fileSettings');
-  const agendaStartOnWeekday = state.base.get('agendaStartOnWeekday');
   return {
     files: determineIncludedFiles(allFiles, fileSettings, path, 'includeInAgenda', false),
     todoKeywordSets: file.get('todoKeywordSets'),
     agendaTimeframe: state.base.get('agendaTimeframe'),
     agendaDefaultDeadlineDelayValue: state.base.get('agendaDefaultDeadlineDelayValue') || 5,
     agendaDefaultDeadlineDelayUnit: state.base.get('agendaDefaultDeadlineDelayUnit') || 'd',
-    agendaStartOnWeekday: agendaStartOnWeekday == null ? 1 : +agendaStartOnWeekday,
+    agendaStartOnWeekday: state.base.get('agendaStartOnWeekday') || 1,
```
tomonacci, line 185: Since number 0 is falsy, I believe this will lead to a behavior where tapping Sunday will set it to Monday

src/components/UI/TabButtons/index.js
```diff
@@ -30,7 +30,7 @@ export default ({ buttons, titles, values, selectedButton, useEqualWidthTabs, on
 
         return (
           <div
-            key={buttonName}
+            key={buttonName + index}
```
tomonacci, line 33: Nice catch, thanks; when I had the inspector open I was using the three-letter week and after switching to one-letter week I didn’t check the console again so that’s probably why I missed this

tomonacci (commented):
Thanks for the fixes!

src/components/OrgFile/components/AgendaModal/index.js
schoettl: I don't have a better idea than checking for null. Maybe the store has a method for this? Otherwise, I would check the spec, if store.get() ever returns undefined. If not, I'd go with `=== null` and otherwise rely on the store to return a integer (i.e. no `+` hack).

munen: You are correct, of course!

Apologies for the extra work. I'm currently on day 7 of the monthly 7-day Zen retreat - I'll take that and blame the lack of sleep for missing the falsiness of 0. Thank you for the extra effort and the review :pray:

src/components/UI/TabButtons/index.js
munen: No worries, it's easy to do - for you, there was a predefined component which already uses keys, so you even had good reason to be happy with a quick glance(;

munen:
I have added an integration test and updated the changelog with attribution. This is ready to be merged into `master`.

Thank you @tomonacci for your contribution :pray:

Issue #705, opened 2021-06-23
`yarn test` fails on fresh install on mac


jzohrab:
**Describe the bug**

Running all tests as at commit `f999aeba31c257cc52c`

```
 FAIL  src/components/OrgFile/OrgFile.integration.test.js (30.213s)
  ● Render all views › Org Functionality › Renders everything starting from an Org file › TaskList › orders tasks for an Org file
```

detail:

```

    expect(received).toMatchSnapshot()

    Snapshot name: `Render all views Org Functionality Renders everything starting from an Org file TaskList orders tasks for an Org file 1`

    - Snapshot
    + Received

    @@ -35,11 +35,10 @@
            value=""
          />
        </div>
        <div
          class="task-list__headers-container"
    -     style="overflow: auto;"
        >
          <div
            class="agenda-day__container"
          >
            <div

      402 |           fireEvent.click(getByTitle('Show task list'));
      403 |           const drawerElem = getByTestId('drawer');
    > 404 |           expect(drawerElem).toMatchSnapshot();
          |                              ^
      405 |         });
      406 | 
      407 |         test('search in TaskList filters headers (by default only with todoKeywords)', () => {

      at Object.toMatchSnapshot (src/components/OrgFile/OrgFile.integration.test.js:404:30)

 › 1 snapshot failed.
Snapshot Summary
 › 1 snapshot failed from 1 test suite. Inspect your code changes or press `u` to update them.

Test Suites: 1 failed, 10 passed, 11 total
Tests:       1 failed, 381 passed, 382 total
Snapshots:   1 failed, 2 passed, 3 total
Time:        25.271s
Ran all test suites.
```

I re-ran all tests, this failed again.

If I re-run with just failed tests, that test still fails.

**Desktop (please complete the following information):**
 - Mac Catalina 10.15.7

LMK if any more detail needed.

I wanted to submit this before spending any time trying to solve the problem, perhaps the team's already aware.  Thank you!

munen:
Hi @jzohrab 

Thank you for bringing this up and apologies for the late response!

It's unfortunate that the tests do fail on a Mac. We haven't been aware of this, yet. The contributors mainly use Linux and Windows, I'm not sure if a Mac user has contributed yet. Having said so, there shouldn't be anything in the tests or organice that's platform specific and all contributors and users are very welcome!

[...]

PR #707, opened 2021-06-28
fix issue 705: failing unit test.

jzohrab:
Fix for https://github.com/200ok-ch/organice/issues/705

All tests pass with this fix.


munen:
Commenting in the original issue at #705.

Issue #583, opened 2020-11-24
Unable to display org file without any headers


joelmccracken:
I have a file for which I receive a parse error:

```
Notes for self


ef.el contains code
.dir locals automaticaly loads this file


defun ef/print-files-with-new-versions
defun ef/print-all-files
```

Basically this is just a plain text file with the extension `.org`. I would expect it to be able to render this as a plain text file. Often my org mode files start out as a dump of text and I knock it into a structure when I have time.

joelmccracken:
Problem appears to be here

https://github.com/200ok-ch/organice/blob/master/src/components/OrgFile/index.js#L458

it displays an error message if there are no headers, not necessarily if there is a parse error. So actually parser is seemingly fine.

munen:
Hi @joelmccracken 

Thank you for checking out organice. What you describe is a known restriction in organice and documented in the README: 
https://github.com/200ok-ch/organice#current-restrictionsexpectations-of-organice

All actions in organice focus around headers, so it doesn't make sense to display a file without headers, because you couldn't do anything with it. You can check out the UX paradigm of organice with the sample file here: https://organice.200ok.ch/sample

Are you interested in proposing how to handle a file such as yours?

munen:
The only option that I can think of is force the content of a file without a header into an implicit header. That would probably work. 

On export, it would add a `*` to the beginning of your file. However, that would violate a principle of organice which is not to create a diff on unchanged content of a viewed/edited Org file. 

The exporter could then in turn be changed to handle this edge case, too, and remove the star with empty title and just dump the rawDescription. 

@joelmccracken Would you be interested in implementing this?

joelmccracken:
Yeah I can implement it. I'm trying to give org-mode another go as my primary means of self-organization, and I'm thinking that organice will be a big part of it. I was planning on hosting my own, so its either I fork and fix it for myself, or I just stick a header at the top of every org file I have in my repo. I'll poke around and see if I can figure it out.

munen:
I meant not to ask if you want to fork and fix it for yourself, but to fix it upstream. 

There’s also no need to fork for self hosting. Here’s documentation for deploying yourself, there’s plenty of options: https://organice.200ok.ch/documentation.html#deployment

munen:
In any case, good luck with getting organized with Org mode!(;

joelmccracken:
[...]

PR #708, opened 2021-06-28
Wip 583 handle empty file

jzohrab:
DRAFT PR ONLY to share work.

Tentative work to address empty org files, issue raised in https://github.com/200ok-ch/organice/issues/583

I added a dummy header (with text `* ORGANICE_DUMMY`) because when I added an empty line, some tests started to fail.  There are existing tests that check that org files with an empty header aren't changed ... but if I add and then remove an empty line, things break.

I haven't updated any view code to ignore the "ORGANICE_DUMMY" header line.  I'm actually not sure how the code _should_ handle the case where there aren't any headers.  Since a `.org` file is sort-of-assumed to have a header, maybe the existing failure is actually legit, because adding a dummy line like this code does, or changing structure, could somehow influence an existing org file in a bad way.  Not sure, and I don't know if it's useful thinking about this one too hard!

Cheers, jz

### DIFF
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -683,3 +683,82 @@ describe('Render all views', () => {
     });
   });
 });
+
+
+describe('Render file with no headings', () => {
+  const testOrgFile = readFixture('no_headings');
+
+  let store;
+
+  beforeEach(() => {
+    let capture = Map();
+    capture = capture.set('captureTemplates', []);
+    store = createStore(
+      rootReducer,
+      {
+        org: {
+          past: [],
+          present: Map({
+            files: Map(),
+            fileSettings: [],
+            search: Map({
+              searchFilter: '',
+              searchFilterExpr: [],
+            }),
+          }),
+          future: [],
+        },
+        syncBackend: Map({
+          isAuthenticated: true,
+        }),
+        capture,
+        base: new fromJS({
+          customKeybindings: {},
+          shouldTapTodoToAdvance: true,
+          isLoading: Set(),
+          agendaTimeframe: 'Week',
+        }),
+      },
+      applyMiddleware(thunk)
+    );
+    store.dispatch(parseFile(STATIC_FILE_PREFIX + 'fixtureTestFile.org', testOrgFile));
+    store.dispatch(setPath(STATIC_FILE_PREFIX + 'fixtureTestFile.org'));
+  });
+
+  describe('Org Functionality', () => {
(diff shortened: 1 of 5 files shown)


jzohrab:
Hm, all tests pass locally, and I couldn't see the circleci failures.  No problem just yet, until we decide if this work is any good.

munen:
Hi @jzohrab 

Nice that you've picked up this task!

> There are existing tests that check that org files with an empty header aren't changed ... but if I add and then remove an empty line, things break.

Yes, one of the quality goals for the parser is that when it parses and re-exports an Org file, it should not change the original file. Not seeing unrelated diffs is important for the productivity of the user. It sounds trivial, but lots of alternative products do not live up to this expectation. (See https://organice.200ok.ch/documentation.html#background-information)


> Hm, all tests pass locally, and I couldn't see the circleci
> failures. No problem just yet, until we decide if this work is any
> good.

Yes, let's ignore this issue in this PR. We have a separate issue (https://github.com/200ok-ch/organice/issues/705) for that.

> On the other hand, this is still potentially worthwhile ... at least
> in my case, I have an "inbox" file that I empty regularly, and I add
> TODO items to it as they come to me. Since it's usually empty,
> organice doesn't like it, but that's actually one of the main
> reasons I'd like to have organice :-) I could just as well add a
> dummy entry, that's not a problem, and maybe that's the easiest
> workaround for now.

There's no need to add a dummy entry for that. This use case is well supported by organice and I use it daily.

If you have a file:

``` org
* Inbox
** Tasks
```

Then this capture template will work well for you:

image


As for the implementation in this PR, it doesn't quite work for me. I've tested:

1. Create an org file with this content:

``` org
- Foo

- Bar
```

2. I've opened the file in organice.
3. First bug: I didn't see the content of my file.
4. I've added a header "TODO Foobar"
5. The content of the resulting file was:

``` org
- Foo

- Bar
* ORGANICE_DUMMY
* TODO Foobar
```

I would have expected:

``` org
- Foo

- Bar
* TODO Foobar
```

Having said so, I've added a UX proposal here that's a bit different than adding a dummy header: https://github.com/200ok-ch/organice/issues/583#issuecomment-877782251

jzohrab:
Hi @munen , thanks for the notes.  Yes this PR code was rather hacky :-) and I didn't even bother checking how it worked in the UX because it felt like I was going in the wrong direction.  (I did see the notes about idempotent open-and-export .org files which makes sense, and makes things tricky.)  I'll add some notes in #583, but won't update this PR just yet, as I think it's likely a dead end.

PR #717, opened 2021-07-15
Fix typo

schoettl:
Is a typo, right @munen ?


munen:
@schoettl *hihi*, good catch!

100% copy/paste typo(;

PR #721, opened 2021-07-21
Onedrive

apaes:
This is my "amateur" try at implementing OneDrive backend for organice. Some requirements to make this work:

1) Register an application with Microsoft (in my case I've done it through my free Azure account)
2) Add a Single Page Application client to your registered APP and specify some redirect URIs, in my case:
    2.1) http://localhost:3000/blank.html
    2.2) https://<my-own-domain>/blank.html
2) Decide who will be allowed to authenticate against the application (single-tenant, multi-tenant, multi-tenant + personal accounts, personal accounts only)
3) Copy the CLIENT-ID info of the registered application and use it in your .env file as REACT_APP_ONEDRIVE_CLIENT_ID
4) copy the TENANT ID (in case of single-tenant)  or one of the terms "organization", "common", "consumer" depending on the authentication scenario  in (2) and update REACT_APP_ONEDRIVE_TENANT_ID on your .env file

Other notes:

a) I've added dependency of MS authentication library  - @azure/ms-browser
b) I've added onedrive.js to /sync_backends to (implement some sign-in functionality, etc.) and onedrive_sync_backend_client.js
c) added public/blank.html for the OneDrive login redirect process
d) Microsoft supports 2 authentication methods (through window popup and page redirect) but due to lack of knowledge on how to implement additional URIs in organice I decided to rely solely on PopUp auth. This means popups will need to be enabled for this APP

Best,
Alex


yarn.lock
```diff
@@ -1,13513 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
schoettl, line 1: I don't think that this file should be deleted. It's not a temporary .lock file, it's rather an config file to freeze the installed packages and make the build reproducible.
https://classic.yarnpkg.com/en/docs/yarn-lock/

schoettl (changes requested):
I just had a short look on your changes. Maybe you can install prettier in your IDE to stick to Organice' indentation style. Otherwise, the automatic formatting can also be done in a separate commit later...

Would be really cool, if we had an additional cloud option!

munen:
Just a quick note from me: @apaes did send me an email before opening the PR. He clearly stated that he’s not a developer, but that he did manage to get OneDrive to work for him. He asked if that would be interesting for upstream organice. I congratulated him on the proper hacker spirit and that this is a great achievement which actually kind of makes him a developer(; I also said that I would look into the PR and help with feedback and testing.

Since the PR has been opened, there was too little time for me to properly look into it. Currently, I’m sitting in a train and am typing this on my phone. I’m off to proper vacation without a laptop for the first time in about 8 years. I did just read through the code and on a first glance it looks like a good base to get OneDrive working. I’ll definitively get back to this PR when I can.

In the meantime, feel free to collaborate with @schoettl who already made some good suggestions. He’s a maintainer of organice, as well.

🙏

apaes:
Hi,

I apologize for not having replied to schoettl email earlier. Have been
going through Incident Response exercises in my org followed by internal
audits so my diary has been quite full.

Will do my best to follow up with it this weekend.

Wish you great vacation and excelent rest!

Best,
Alex

munen:
No worries, take your time. This is a community FOSS project build in spare time. There’s no deadlines^^

Thank you for the kind wishes! The first day winds down, we’re just coming back from a nice restaurant by the sea. Today has been a greatly relaxing day, indeed. So far, I can recommend this idea of doing a vacation every once in a while(;

Issue #739, opened 2021-11-02
Cannot log in when selecting the GitLab back-end (sporadically) 


munen:
**Describe the bug**

@branch14, the co-founder of 200ok.ch, was super happy to hear about the GitLab back-end in organice, so he wanted to use it right away. At first, he could login using Firefox on the Desktop and could log in on his smartphone (Android).

After some time passed, he got an error on organice on the smartphone, so he tried logging back in. Now, logging in on the smartphone reproducibly doesn't work whereas it did work before - and he's still logged in on the Desktop browser.

**To Reproduce**
0. Use @branch14 Android device (yes, this is where the reproducability already stops unfortunately)
1. Go to https://organice.200ok.ch
2. Click on 'Sign in'
3. Select 'GitLab', enter his org repo

**Screenshots**

![Screenshot_20211102_185224_org mozilla fenix](https://user-images.githubusercontent.com/505721/139920693-99952f1f-f2e3-4321-b964-c23c5fd57fd9.jpg)

The visible error is generated here: https://github.com/200ok-ch/organice/blob/master/src/actions/sync_backend.js#L69

When inspecting the error by connecting the phone to his Desktop, we see that the issue is not actually 'retrieving the directory', but one step ealier while authenticating against GitLab.

![image](https://user-images.githubusercontent.com/505721/139919357-a1c52655-61fd-44d5-bf03-6b4b6f38eddb.png)

![image](https://user-images.githubusercontent.com/505721/139919390-741030a9-7c85-4de9-a32d-b1620ebe322a.png)

**Smartphone (please complete the following information):**
 - Device: Huawei P20 Pro
 - OS: I'm assuming some current Android, don't think/hope that's relevant.
 - Browser: Firefox

munen:
@chasecaleb Do you have an idea what this is about?

The only additional context that I can provide that @branch14 can can continue to use organice.200ok.ch on the Desktop and while he is describing the issue, I can continue to log out and in both on the Desktop and my iPhone, so it's likely not a general outage on the side of GitLab.

chasecaleb:
Ugh, sorry about this! So just to make sure I'm on the same page, there's two scenarios where that token call can happen:
1. While obtaining the initial token, i.e. clicking "sign in" from https://organice.200ok.ch/sign_in
2. While using the refresh token to obtain a new access token, i.e. at some random time while using the app (roughly >= 2 hours after sign in)

[...]

PR #740, opened 2021-11-03
Fix GitLab OAuth token refresh

chasecaleb:
Fix for #739. This turned out to be tricky to debug but the solution is trivial, fortunately. Here's what was happening, brace yourself:

1. Signed-in user with an expired OAuth token visits https://organice.200ok.ch/files . Tangential-but-irrelevant trivia: GitLab tokens expire after 2 hours and are not refreshed until expiration, so expiration could happen even while actively using the app.
2. On startup, Organice checks if the sync client is still signed in. **Important detail**: because this occurs in a promise that isn't waited on before doing anything else, additional API calls (e.g. to fetch the config file and fetch org file contents) also happen concurrently before this finishes. https://github.com/200ok-ch/organice/blob/891716bcf9c199356a024dfde35e8a9796bdbe42/src/App.js#L168-L174 
3. The GitLab sync client's `isSignedIn()` will refresh the OAuth access token if it is expired by calling `OAuth2AuthCodePKCE.getAccessToken()`. If that it fails, `isSignedIn()` returns false. Keep in mind that a rejected Promise translates to a thrown exception in an `async` function. https://github.com/200ok-ch/organice/blob/891716bcf9c199356a024dfde35e8a9796bdbe42/src/sync_backend_clients/gitlab_sync_backend_client.js#L118-L130
4. Like I pointed out in step 2, there are additional GitLab API calls potentially happening concurrently to `isSignedIn()`. All of the GitLab API calls use `OAuth2AuthCodePKCE.decorateFetchHTTPClient(..)`, which handles OAuth things **including** calling the previously mentioned `getAccessToken()`.
5. As a result, there are multiple concurrent calls to GitLab's token refresh endpoint: https://gitlab.com/oauth/token. All of these refresh calls use identical parameters, which OAuth (at least with 2.0 PKCE) does not allow. All but one of these calls fail with a 400 status and `invalid_grant` error. Side note: according to the OAuth 2.0 spec, `invalid_grant` is pretty much a catch all for "you did something wrong, good luck figuring it out" :facepalm:
6. When `isSignedIn()` returns false, the `App.js` startup logic triggers a sign out action (refer back to step 2 for the code snippet).
7. There's some other problematic behavior going on at this point, but I'm not exactly sure what. Somehow it's possible for `localStorage.authenticatedSyncService` to be cleared without necessarily everything else in `localStorage` being cleared. This is the behavior that we saw here: https://github.com/200ok-ch/organice/issues/739#issuecomment-958107531

**TLDR summary**:
The GitLab backend was fine, turns out the root cause is due to the OAuth library not handling concurrent requests very well if the access token is expired. Fortunately as you can see in this PR, the workaround is simple and isolated to a single location.

---
**Reproduction** from any device, not just mobile:
1. Open dev tools
2. Run this in console:
    ```javascript
          let authstate = JSON.parse(localStorage["oauth2authcodepkce-state"]);
          authstate.accessToken.expiry = new Date("1/1/2020").toString();
          localStorage["oauth2authcodepkce-state"] = JSON.stringify(authstate);
    ````
3. Switch to network requests view of dev tools
4. Ensure that network requests are persisted across page reloads (there's a setting for this somewhere in dev tools, specifics depend on your browser)
5. Reload page
6. Observe multiple calls to https://gitlab.com/oauth/token. One will succeed, the rest will fail with status code 400.

If you do the same with my fix in this PR, instead of multiple token calls in step 6 you will only see a single one.

---

**Tangent**: If the opportunity arises, I think it would be beneficial to improve how side effects are handled in this application, particularly when it comes to asynchronous things like API calls. The craziness that is the constructor of `App` in `src/App.js` is a good example of what I'm referring to.

For example, I've used [Redux Saga](https://redux-saga.js.org/) extensively in a good sized project at work and it worked great. However it can be confusing for people to learn because it's based around the [Command pattern](https://en.wikipedia.org/wiki/Command_pattern) using [JavaScript generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*), so I'm not necessarily suggesting it as a good fit for this project. There are other libraries for this use case too, e.g. [Redux Thunk](https://github.com/reduxjs/redux-thunk) is another popular option off the top of my head. I'm sure there are quite a few other ways to improve things too.

**However**,  I'm just pointing this out as food for thought someday, I am **not** suggesting any particular changes or solutions since it would be a significant amount of work to deal with. I haven't put much thought into specific solutions or a way to get there without rewriting the world... just pointing out what seems like a pain point from my limited experience with the project so far.

---

I also created an issue for the library as you can see from my comment: https://github.com/BitySA/oauth2-auth-code-pkce/issues/29 . Hopefully we can remove this workaround one day in the future; in the meantime, at least it's a straightforward and well-contained solution.

### DIFF
--- a/src/sync_backend_clients/gitlab_sync_backend_client.js
+++ b/src/sync_backend_clients/gitlab_sync_backend_client.js
@@ -5,8 +5,13 @@ import { getPersistedField } from '../util/settings_persister';
 
 import { fromJS, Map } from 'immutable';
 
-export const createGitlabOAuth = () =>
-  new OAuth2AuthCodePKCE({
+export const createGitlabOAuth = () => {
+  // Use promises as mutex to prevent concurrent token refresh attempts, which causes problems.
+  // More info: https://github.com/BitySA/oauth2-auth-code-pkce/issues/29
+  // TODO: remove this workaround if/when oauth2-auth-code-pkce fixes the issue.
+  let expiryPromise;
+  let invalidGrantPromise;
+  return new OAuth2AuthCodePKCE({
     authorizationUrl: 'https://gitlab.com/oauth/authorize',
     tokenUrl: 'https://gitlab.com/oauth/token',
     clientId: process.env.REACT_APP_GITLAB_CLIENT_ID,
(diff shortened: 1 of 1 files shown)


branch14:
This is great work and I highly appreciate the effort you put into documenting the issue and your workaround.

LGTM.

PR #745, opened 2021-11-06
WIP Refactor search filter

schoettl:
The filter for clock times should be in `isMatch` ...

### DIFF
--- a/src/reducers/org.js
+++ b/src/reducers/org.js
@@ -1167,9 +1167,7 @@ export const setSearchFilterInformation = (state, action) => {
     const headers = files.map((file) => file.get('headers'));
 
     // show clocked times & sum if there is a clock search term
-    const clockFilters = searchFilterExpr
-      .filter((f) => f.type === 'field')
-      .filter((f) => f.field.type === 'clock');
+    // TODO these depend on clockFilters which is now in isMatch() ...
     const filterFunctions = clockFilters.map(timeFilter);
     const showClockedTimes = clockFilters.length !== 0;
     state.setIn(['search', 'showClockedTimes'], showClockedTimes);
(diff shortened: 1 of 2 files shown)


src/reducers/org.js
```diff
@@ -1167,9 +1167,7 @@ export const setSearchFilterInformation = (state, action) => {
     const headers = files.map((file) => file.get('headers'));
 
     // show clocked times & sum if there is a clock search term
-    const clockFilters = searchFilterExpr
-      .filter((f) => f.type === 'field')
-      .filter((f) => f.field.type === 'clock');
+    // TODO these depend on clockFilters which is now in isMatch() ...
```
schoettl, line 1170: It will take me a time, too, to understand all those filters again. I moved the clock filter added by you, @tarnung, to `isMatch` – but I don't understand yet what your other lines do and how we can refactor that...

It's late ^^ I just wanted to make a note for now.

tarnung:
The main issue is that the "clock" search criterion depends on information that is spread out over multiple headers because "logged time" is a recursive property that gets inherited by and added up in parent headers.
It is therefore necessary to do the calculations of "logged time" in a place where we have access to all headers - which isMatch hasn't.
The result of these calculations are saved as "header.totalFilteredTimeLogged" and "header.totalFilteredTimeLoggedRecursive". The actual search function is just checking if "totalFilteredTimeLoggedRecursive" is not zero.

1. checking if not zero could be pushed to "isMatch"
2. calculation of "totalFilteredTimeLoggedRecursive " zero could be pushed into "isMatch" if the returned function gets all headers with "totalFilteredTimeLogged" precalculated as a second argument
3. calculation of "totalFilteredTimeLogged" can only be pushed into "isMatch" if we calculate it for the header handed to "isMatch" and all its subheaders.

Option one does not offer much gain.
Option two depends on if you are willing to allow "isMatch" or the function returned by it access to all searched headers.
Option three would probably be the cleanest but some calculations are done multiple times especially for deeply nested headers.

Issue #737, opened 2021-11-01
Show currently clocked heading


chasecaleb:
**Is your feature request related to a problem? Please describe.**
As a user who tracks their time with clocking, I would like a way to tell when I have a clocked in heading and a way to navigate to it as well.

The current clock functionality in organice is better than nothing since it allows me to clock in and out on a heading, but the lack of UI visibility is limiting. There are a couple main scenarios where this is an issue:
1. As a user, I have the memory of a goldifsh and forget whether or not I have a clock running. Similarly, even if I remember that I'm clocked in I may forget which heading is clocked.
2. As a user who sometimes clocks into a heading but then gets distracted by multi-tasking or needs to find related information in a different org file, I would like a button to take me back to the currently clocked heading.
3. *[Tangential, but likely needs to be tackled as part of this]* As a user with the memory of a goldfish, I would like a way to prevent clocking into multiple tasks at the same time.

**Describe the solution you'd like**
- For scenarios 1 & 2 above:
    - At minimum, an indicator that a task is currently clocked in and a button to navigate to it.
    - Optionally: also a way 
- For scenario 3: Emacs handles this by automatically clocking out of the previous heading when clocking into a new one. I think this would be fine for organice as well.

Additional requirements:
- This should work across multiple files. It doesn't necessarily have to be every single org file in the backend; either every organice agenda file or every file that has been loaded/persisted to storage would both work I think.
- *Nice-to-have*: It would be nice if this can detect existing clocks by scanning file contents, as opposed to keeping track of the current clock out-of-band. This would make it possible to clock in from one place (e.g. from Emacs on a computer) and then clock out from another (e.g. from organice on my phone). This would also be a more robust implementation with lower risk of bugs due to state getting out of sync.

**Additional context**
Emacs accomplishes this by displaying the current task in the mode line and providing function `org-clock-goto`.

--- 

I'm **planning on implementing this myself** if there's interest. I'd like some help coming up with and agreeing on a mock first though. I haven't worked on any mocks in Excalidraw myself yet because I wanted to at minimum bounce ideas off you first. 

Possible ways to do this:
1. Add a button within the "Show Search / Task List" magnifying glass icon at the bottom to go to the clocked heading. Perhaps the magnifying glass icon could also change colors to indicate when a clock is running.
[...]

PR #747, opened 2021-11-07
feat: suggestion for clock:now

tarnung:
A quick and dirty implementation for parts of the feature set described in #737

=> clock:now searches for active clocks

I thought that other things than clocks can create "logBookEntries". But maybe i misremember. I could not find any in my demo files.

This pull request does not address the issue #745. Since it does not add to the problem we can go along and refactor later once @schoettl and I what if anything should be done.

### DIFF
--- a/sample.org
+++ b/sample.org
@@ -374,6 +374,7 @@ Next to units, =START= and =END= can also be expressed in dates like. Dates can 
 Some examples on how to use time ranges:
 
 - =clock:..= searches for all items with time clocked. This includes headers that have time logged on their children.
+  - =clock:now= searches for currently clocked in headers.
 - =sched:w= searches for all scheduled items in the current week (same for other units).
 - =dead:..w= searches for all deadlines between now and the end of the week.
 - =date:today= searches for all planning items (scheduled and deadline items) as well as items with active timestamps.
--- a/src/colors.css
+++ b/src/colors.css
@@ -6,7 +6,7 @@
   usage guide:
   ------------
 
-  this is supposed to document current usage, 
+  this is supposed to document current usage,
   it's not complete, not necessarily prescriptive and
   some usages might need to be reconsidered to be more consistent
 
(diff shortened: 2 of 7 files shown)


tarnung:
I added a basic indicator for active clocks. It checks for them when parsing the org file and updates their number when clocking in/out.
It currently just turns a button red. This should probably be surplanted by a solution that is more theme aware.

tarnung:
The indicator can get out of sync when headers with active clocks are deleted. Maybe there's a more robust implementation, maybe this is good enough as it corrects itself on the next file parse.

munen:
Nice PR!

> I thought that other things than clocks can create "logBookEntries". But maybe i misremember. I could not find any in my demo files.

A `:LOGBOOK:` drawer can also contain a state change for a task. For example updating this task:

```
* TODO Repeating task
  SCHEDULED: <2019-11-27 Wed +1d>
````

will become:

```
* TODO Repeating task
  SCHEDULED: <2019-11-28 Thu +1d>
  :PROPERTIES:
  :LAST_REPEAT: [2021-11-09 Tue 14:39]
  :END:
  :LOGBOOK:
  - State "DONE"       from "TODO"       [2021-11-09 Tue 14:39]
  :END:
``` 

Example taken from https://github.com/200ok-ch/organice/blob/master/test_helpers/fixtures/various_todos.org#repeating-task

Relevant Org mode doc: https://orgmode.org/manual/Drawers.html

Having said that, your `hasActiveClock` implementation looks solid enough that it will only trigger on actual clocks(;

munen:
I have gone through the code and tested it. It works for me and I like the implementation.

I haven't merged, yet, for this reason: Since the magnifying glass already changes color, I'd say this PR also needs the icon that populates a search with `clock:now`. Otherwise the magnifying glass indicates something that is actionable, but the user doesn't know what. Alternatively, to get this PR quicker into `master`, we can remove the color change and add it at a later date.

tarnung:
this pr is replaced by #751 if the changes over there are accepted

chasecaleb:
Nice job on this, thanks for implementing @tarnung!

Issue #217, opened 2020-01-08
As a user, when creating/modifying a SCHEDULED and DEADLINE timestamp, I want to see more semantic information in the modal


aspiers:
(This is a minor UI/UX bug, but when combined with #218 it becomes a slightly more significant problem.)

If I click either of [the new `DEADLINE` or `SCHEDULED` icons](https://github.com/200ok-ch/organice/issues/188#issuecomment-571730711), I get an identical-looking modal dialog with no way to tell which type of timestamp I'm editing.  They should say "Edit scheduled timestamp" and "Edit deadline" or similar to differentiate the two.

This is particularly important given that currently the two icons are very similar and it's not immediately obvious which is which.  I've filed #218 to track that separate but related issue.

schoettl:
Good point, this was also on my list.

Maybe we can make the title of the modal a property and pass an appropriate value for it.

There must be a third option "Edit timestamp" for normal timestamps in title or description.

We could also differentiate between "Add deadline" and "Edit deadline", to make clear that there was no deadline before.

munen:
@aspiers @schoettl Does either of you want to work on this?

One note I want to make: Currently, it's always "edit timestamp", never "create timestamp". Rationale: As soon as you hit the action button, you'll have a corresponding timestamp for today. In the modal, you can change the date or delete the timestamp altogether. 

It would be possible to change this behavior, but that would not be trivial - even from a UX standpoint, because it would require more clicks to set a simple timestamp. 

Hence, I propose to keep this issue to differenciate between 'schedule' and timestamp' in the modal header.

aspiers:
[**@munen**](https://github.com/munen) commented on [January 9, 2020 10:41 AM](https://github.com/200ok-ch/organice/issues/217#issuecomment-572502930 "2020-01-09T10:41:36Z - Replied by Github Reply Comments"):
> [@aspiers](https://github.com/aspiers) [@schoettl](https://github.com/schoettl) Does either of you want to work on this?
> 
> One note I want to make: Currently, it's always "edit timestamp", never "create timestamp". Rationale: As soon as you hit the action button, you'll have a corresponding timestamp for today. In the modal, you can change the date or delete the timestamp altogether.
> 
> It would be possible to change this behavior, but that would not be trivial - even from a UX standpoint, because it would require more clicks to set a simple timestamp.

Makes sense.

> Hence, I propose to keep this issue to differenciate between 'schedule' and timestamp' in the modal header.

[...]

PR #748, opened 2021-11-08
feat: no inline editing again

tarnung:
In #559 I tried two things:
- get rid of the inline editor
- get rid of the inline action bar

While #559 was some tinkering on my side to see what's possible, I'd like this PR to be something that can actually be merged.

I took the first point - get rid of the inline editor - and brought the code up to date with 200ok/organice.
I think getting rid of the inline editor should be less controversial than an overall UI-overhaul. We already have several buttons in the inline action bar that open editing popups. It's not that big of a change to have "edit title" and "edit description" do the same.

In #559 i opted to have all my changes behind a feature toggle. I don't in this PR.

Some issues still need to be tackled:
- [x] General clean up of code no longer used (things considering editMode)
- [x] Shortcuts / Hotkeys
- [x] unnecessary syncing when editing title / description
- [x] the table editor crashes when the last cell is deleted
- [x] fix tests
- [x] scrolling long descriptions on iOS
- [x] fix / remove exit editor mode shortcut

Tell me if you notice anything else!


munen:
This looks amazing! I agree, doing it like this is absolutely not contoversial, but brings a lot of the benefits that we've discussed. I also agree that there should be no feature toggle for this. I'm super stoked that you've brought this PR up, again :bow: :pray: 

> Tell me if you notice anything else!

One thing I have seen so far is: When I'm editing a header and I toggle between 'edit header title' and 'edit header description', I see the 'unpushed changes' notice flashing which then will trigger a sync.

Interestingly, the change is 0 bytes. That can be seen nicely when using the new GitLab backend. The commits that are created look like this:

```
% git log --oneline | head -3
7f154a6 [organice] Update test.org
12c8506 [organice] Update test.org
2ce2b97 [organice] Update test.org
```
However:

```
% git diff --shortstat HEAD~3 HEAD
%
```

tarnung:
The "exit edit mode" shortcut does not work - it would have to close the modal now.
That said, it didn't work for me before. But now it's guaranteed not to work.

tarnung:
The last commit does what #217 proposed.

tarnung:
I still need to fix the tests and there is the one broken shortcut, but otherwise I think this is ready.

munen:
First impression: This is amazing - it's enjoyable to use organice like this!

+1 for solving #217 in this big PR, as well :rocket: 

I confirm that the shortcut to close the edit modal does not work on `master`. If you prefer, you could delete it for the time being.

I have deployed it to Staging for further testing (NB: Staging does not support the GitLab back-end yet).

P.S.: The only issue I've seen so far is that I still see synchronization triggered without making content changes:

- when switching to the `tags` modal
- when in edit title and then closing the modal
- when in edit description and then closing the modal

munen:
I did find one more bug: On iOS, when I try to scroll a description that's longer than the textarea, I can't.

Here's a video comparing current `master` (first part) with the PR branch (second part): https://www.dropbox.com/s/y6ue9731155423s/2021-11-12_cannot_scroll_description.mkv?dl=0

Tested on an iPhone Xs with iOS 15.0.2.

tarnung:
@munen scrolling long descriptions works on (my) android so its hard for me to debug that.

munen:
@tarnung Looks like you’ve made good progress, again!

I’m in the mountains this weekend and don’t have a notebook with me. If you want, you can force push to develop to deploy a new version to Staging and I can already test the changes on iOS. 

As for the scrolling issue on iOS, I can take a look at that when I’m back. Happy to hear it’s working on Android, already!

tarnung:
Shortcuts are kind of broken on master too. It's possible to open the title editor or move headers and change the selected header while a popup is opened.
Shortcuts are ignored when they originate in an 'input', 'select' or 'textarea' element. This can be changed by [configuring react-hotkeys](https://github.com/greena13/react-hotkeys#Configuration).
Syncing seems to removes focus from the HeaderList therefore disabling the shortcuts.

My current best solution
- switch to GlobalHotKeys so focus on the HeaderList is not required
- configure react-hotkeys to also listen to keys pressed when focus is on input elements
- guard all shortcuts to only do something when a header is selected and no popup is open (and not shouldDisableActions)
  - exception: "exit edit mode" is always available and closes any popup

It seems to work quite well. I sometimes land in a strange state where an action triggered by a key+modifier like "ctrl+enter" already triggers on pressing the modifier "ctrl" in this case. I don't know if this has anything to do with changes and I have not yet found a reliable way to trigger it.

tarnung:
this pr is replaced by #751 if the changes over there are accepted

munen:
After an hour of debugging, I have narrowed down the issue of iOS scrolling: The `<Motion>` component has listeners on `touchstart`, `touchmove`, `touchend` and `touchcancel`: https://github.com/200ok-ch/organice/blob/72b46bf80ade330f59d0747246f0c8b5547ab227/src/components/UI/Drawer/index.js#L64-L67

The `<textarea>` of `<DescriptionEditorModal>` is embedded within `<Motion>`. These touch event listeners interfere with the ability to scroll on iOS. When removing the listeners, scrolling within the textarea works just fine.

I see multiple ways going forward:
- Remove the touch/drag option to close a drawer (and instead add an 'x' to the top right or something similar
  - Just removing the touch handlers is an immediate fix. "Just" the UX of the drawer is a bit wonky, then.
- Figure out how to get scrolling to work for a textarea within a touchable floating container
  - This article seems to touch (nice evening pun..) on this: https://medium.com/turo-engineering/ios-mobile-scroll-in-web-react-1d92d910604b

munen:
Good news, I have fixed scrolling on iOS!

tarnung:
Nice! I guess the same should be done to other input fields that are moved to modals like "edit title" and "add note".

munen:
> Nice! I guess the same should be done to other input fields that are moved to modals like "edit title" and "add note".

Good point!

I just saw that "add note" actually only accepts one line of input - as soon is the user hits RET, the note is saved to the document whilst the modal stays open.

tarnung:
"add note" had the same functionality as "edit title" - If the cursor is at the end of input "return" instantly adds the note - unlike "edit title" the popup stays open since you might want to add another note while you won't want to edit another title...

I'll have to check out what you changed exactly.

munen:
> I'll have to check out what you changed exactly.

TL;DR: Necessary bug fixes (; I found more issues with the note functionality while looking to fix the iOS scrolling bug.

tarnung:
This PR was included in #753.

Issue #376, opened 2020-06-26
Checking or unchecking a checkbox discards text with markup


schoettl:
**Describe the bug**
Checking or unchecking a checkbox discards all text with markup under the same headline.

**To Reproduce**
1. Create a header with content like:
```
hier =code= hier

- [ ] check

hier *bold* uiae
```
2. Check the checkbox in organice
3. Notice that the mark-up text has gone:
```
hier  hier

- [ ] check

hier  uiae
```

**Expected behavior**
Organice should not delete any text when checking or unchecking checkboxes.

**Desktop (please complete the following information):**
 - Firefox
 - ?

**Smartphone (please complete the following information):**
 - iOS
 - ?

**Notes:**
The deletion is not immediately visible: you have to *edit* the text or *reload* the file to see the change.

munen:
I confirm. Tested on iOS. I did have to reload the file to see the loss of markup. Not that it makes things better, it makes it harder to see the bug for a user. 

Good catch!

schoettl:
More information and screenshots in #796

saikiransathpadi:
@munen  It happened to me even without a reload in the chrome browser. I would like to try this out, could you please assign it to me?

chasecaleb:
@saikiransathpadi Thank you so much!!! You'll make my day if you fix this. I'm not a maintainer so I can't assign this to you, but you're welcome to get started anyways and make a PR whenever you get it figured out.

munen:
@saikiransathpadi You're very welcome, thank you for stepping up :pray: 

I'll be happy to help with testing, too!

saikiransathpadi:
@munen  Hi, https://github.com/200ok-ch/organice/pull/928 I have made the respective changes. Hope this resolves the issue.

munen:
Closed by https://github.com/200ok-ch/organice/pull/928

cc @chasecaleb

PR #749, opened 2021-11-08
add setting to respect os light/dark preferences

tarnung:
As seen in #376


munen:
Pushing to Staging to test.

munen:
Nice change :bow: :pray: 

Happy to pull it in!

PR #746, opened 2021-11-07
fix repeater adjusting time of date

tarnung:
Should fix #683

Since I'm still not using emacs, someone might want to check if my assumptions based on [org mode docs](https://orgmode.org/manual/Repeated-tasks.html#Repeated-tasks) are correct.

using ".+":
- when repeaterUnit is not "h", hour and minutes are never touched.
- when repeaterUnit is "h", hour and minutes are set to n hour(s) from now (as it was already implemented before this fix)

PR #750, opened 2021-11-09
test: Add tests for repeating task with +1d and +1h repeaters

munen:
This adds working test for the recent fix from @tarnung (https://github.com/200ok-ch/organice/pull/746).

Unfortunately, adding the tests uncovered a problem with the test setup. Adding one more top headline into various_todos.org leads to an old test expectation to fail. I have already spent some time in trying to figure out what the problem is and haven't found it so far.


test_helpers/fixtures/various_todos.org
```diff
@@ -3,4 +3,13 @@
 * This is not a todo
 * TODO Task with active timestamp and repeater <2020-11-15 Sun +1d>
 * TODO Repeating task
-  SCHEDULED: <2019-11-27 Wed +1d>
+  SCHEDULED: <2019-11-28 Thu +1d>
+* TODO Repeating task with +1d repeater
```
munen, line 7: The reason why the expectation fails is this line. Adding another headline to `various_todos.org` after the `* TODO Repeating task` leads to the log drawer not to be build. 

Unfortunately, the reason is unclear to me.

tarnung:
Oh my.

This seems to be a parser error. It also exists on master.
Taking a note generates an entry in "logNotes" as it should. Parsing does not recognize the Note though and seems to handle it as regular Text. I can find the Note as a List in "description" and of course in "rawDescription".

To make the behaviour visible:
- Take a note
- Take another note
- See how they are neatly arranged next to each other
- Open and close "edit description" -> we have now lost the "logNotes"
- Take another note
- See how it is displayed visually separated from the other no longer recognized notes

Is it possible that something in your tests touches the description - thereby getting rid of the "logNotes"?

tarnung:
Seems like it's not a bug after all.
Notes are ambiguous. They could be part of a valid description so the parser has to make a decison.
The decision here is to only parse notes if the parser can be sure they are notes. The behaviour is even documented in parse_org.js/_parseLogNotes.

// Only parse log notes if a logbook exists. Otherwise, a list - log notes
// or just a normal list - will go into the description.

Issue #730, opened 2021-09-22
Allow syncing to local filesystem / new sync target


markokocic:
**Is your feature request related to a problem? Please describe.**
Right now it is not possible to configure organice to use local file system to sync / store org files. Only certain cloud providers are supported. For users that are already syncing their org files through different (server, desktop and mobile) devices (manually, using git, syncthing, rsync, network shares, ...) requiring another sync system setup to use organice is an overkill.

**Describe the solution you'd like**
I'd like to be able to configure organice to use a local folder (with subfolders) on the device I use to read from and store org files. It should also be able to automatically detect external changes and has conflict resolution a same way as for current synchronization targets.

**Describe alternatives you've considered**
As I want to avoid using cloud services to store my private data, and setting up, maintaining and securing a personal WebDAV server is too much of a hassle, I don't see any alternative.

**Additional context**
Adding support to sync to local filesystem would decouple syncing from the app, and allow users to use whatever external sync mechanism they like.

schoettl:
See also #93

munen:
@markokocic Which JS API do you propose to employ for accessing the local file system?

markokocic:
I was under impression that progressive web apps have local filesystem access possibility, but looks like I was wrong :(

That being said, [IPFS ](https://ipfs.io/) as a concept for data sharing looks interesting, not sure if there's interest in organice.

fullofcaffeine:
+1 would love to be able to use organic with syncthing.

munen:
@fullofcaffeine I haven’t used Syncthing before and have no experience with it. From a quick glance, I am not sure if it technically can be used from a browser based app. However, if it is: Synchronization in organice is built using the strategy pattern which is consumed here[1]. Implementing a new back-end implies implementing 8 functions[2]. For Dropbox, it’s about 120 LOC JavaScript. 

Would you be interested to integrate a new back-end? I’ll be happy to help with questions, code review and testing. 

https://github.com/200ok-ch/organice/blob/master/src/actions/sync_backend.js
https://github.com/200ok-ch/organice/blob/master/src/sync_backend_clients/dropbox_sync_backend_client.js#L134-L141

munen:
@markokocic @fullofcaffeine @non-jedi Do you plan to work on this?

Otherwise I'll have to close this for one month of inactivity due to https://organice.200ok.ch/documentation.html#how_we_work_with_issues

mrloop:
[...]

Issue #663, opened 2021-05-04
Keep an eye on `browser-fs-access`


munen:
https://github.com/GoogleChromeLabs/browser-fs-access

It's a File System Access API with legacy fallback in the browser. Currently, it only works well on Chrome/Chromium. But there, it's emulating desktop UX: 

- Create a file in the app
- Hit 'save'
- Get a GUI dialogue with the name
- Afterwards when hitting 'save' or using C-s, the document is saved to the same file

This could be nice to integrate to organice whence/if it's implemented for all major browsers.

schoettl:
Haha, browser-fs-access README says:

> This library is a ponyfill.

> Like polyfill but with pony pureness

ponyfill.com says:
> Pony pureness, really?
> While polyfills are naughty, ponyfills are pure, just like ponies.

:D

munen:
Love it!(;

munen:
Status update 3 years later: Still no broad browser support

![image](https://github.com/user-attachments/assets/08f5ec7a-1ad1-42ff-b032-9c728d3270f1)

PR #752, opened 2021-11-17
Local File System Adapter

mrloop:
WIP


leira:
I'm interested in a local file system adapter, then I can use any sync backend rather than rely on the ones supported by organice. Is this PR working? Why is this PR closed? What is the problem of it? Thanks~

munen:
Closed, because it doesn't and cannot work: https://github.com/200ok-ch/organice/issues/730

Here’s what could work: https://github.com/200ok-ch/organice/issues/663

Good luck working on local fs support. Happy to support you with testing and documentation. Looking forward to it!

PR #751, opened 2021-11-14
Search and Task List in one Modal

tarnung:
This PR sits on top of #748 and #747 and adresses task one in #211.

- Search and Task List open in one modal with tabs
- modal opens to "Search" by default and remembers what was opened last (not persisted to backend)
- **NO LONGER TRUE:** if there are open clocks, the last opened modal is ignored and it always opens to "Search" for "clock:now".
- **instead:** if there are open clocks, a third modal "Clock List" is added that looks like "Search" for "clock:now" without the search input field.

Feel free to discuss if this is a sensible behaviour. 

A caveats:
This is a design refactor. It still uses SearchModal and TaskListModal under the hood so there was no cleaning up of duplicated functionality.

PR #753, opened 2021-11-18
Option to not edit raw values

tarnung:
This PR builds on top of #751..

It changes the default of editing the title or description of a header from "edit full value" (such that title + description = full text) to "edit content text only" (no todo keywords, tags, schedule items, properties etc.)

It's possible to toggle between new and old behaviour by clicking the "edit title" or "edit description" icon multiple times.
It's important to have the old behaviour accessible, because otherwise some description parts like repeater logs would no longer be editable anywhere (plus emacs people probably like directly editing text).

There is a setting to indicate that you prefer to see "full values" first and "just content text" only on repeated clicks.


munen:
After a comprehensive code walk-through and demo with @tarnung, this EPIC will be happily merged! :pray: :bow:

PR #755, opened 2021-11-27
custom queries

tarnung:
Basic setup for predefined queries. Code still very messy but working.

Define queries in the settings menu to have easy access to them in the search popup.

Feedback is very welcome before I start to clean up the code changes a bit.


munen:
This is an exciting UX concept in that it allows for different scopes with different queries that are available at all times.

It's completely different from the bookmark feature for search, but it could be precious for people with many projects that want different scopes for various searches.

On a first try, I like it and am stoked that you've already come up with the next major epic in a usable PR! :pray: :bow:

tarnung:
I started out with the "bookmark search" feature in mind, but I started to build from the "stored query" that we introduced for the clock list. 

I'm not sure yet if this feature is useful. I have it on staging and i'll see if I use it at all.
There's the possibility that a "bookmark search" feature could be much more flexible and quicker to use.

munen:
> I'm not sure yet if this feature is useful. I have it on staging, and I'll see if I use it at all.
> There's the possibility that a "bookmark search" feature could be much more flexible and quicker to use.

It's pretty awesome that you've implemented both options! 

The way you've implemented the bookmarks feature is straightforward (in UI surface and code), so it was trivial for me to merge right away. Also, I think I'll use it more than the custom queries. However, you've put a lot of effort into both options, and I'm not against merging the custom queries. So I'll leave the PR open until you give some feedback on your insights on whether or not you're going to make use of custom queries personally. If you do, then this feature will undoubtedly land in `master`!

tarnung:
I'll use the bookmark feature for a bit. If it does most of what I want, there is no need for another concept. Don't expect feedback on this anytime soon. I gues it does not hurt to let this sit around as a draft pr for the time being.

munen:
> I'll use the bookmark feature for a bit. If it does most of what I want, there is no need for another concept. Don't expect feedback on this anytime soon. I guess it does not hurt to let this sit around as a draft pr for the time being.

It does not hurt at all. Take all the time you need!

PR #757, opened 2021-11-29
Feat/scroll layout

tarnung:
A year ago I messed around with the bar at the top of the app and the padding at the bottom that would allow you to scroll content above the action buttons at the bottom of the screen. That had some bad consequences. Most notably even when there isn't enough content to fill the screen, the padding at the bottom still is there and allows to scroll content upwards out of sight. This is true even for pages like login and settings that have no need for padding since there are no action buttons. Sometimes a view even starts out scrolled down so that the content comes into view partially obscured.

I tried to fix this again and I seem to have found a better solution. My knowledge of css is lacking so I mostly try things that seem reasonable until I'm happy with the results.

- the app-bar at the top is now an actual element above the content not a sticky box in front of it.
- on mobile everithing seems to be ok
- on desktop this has the effect that the scroll bar now appears at the boarder of the app content, not at the boarder of the browser window. That has a little less of an elegant feel, but if we ever want to allow for an expanded view on desktop machines, we would need to do that anyways.

You can test the effects of this change on staging, where it currently is deployed alongside the "custom queries" feature.

### DIFF
--- a/src/base.css
+++ b/src/base.css
@@ -34,7 +34,6 @@ body,
 #root {
   color: var(--base00);
   background: var(--base3);
-  scroll-padding-top: 60px;
   user-select: none;
 }
 
--- a/src/components/Entry/stylesheet.css
+++ b/src/components/Entry/stylesheet.css
@@ -3,9 +3,9 @@
 
   display: flex;
   flex-direction: column;
-  height: 100%;
-  min-height: 100vh;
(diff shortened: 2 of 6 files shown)


munen:
This is an excellent change, and it works well for me (also on iOS).

Happy to merge! :pray:

PR #758, opened 2021-12-03
bookmark searches

tarnung:
Simple search string bookmarking.

The initial idea of saving the last 10 used search strings is not practical because we search on every keystroke. The most obvious way to identify useful complete search strings is to let the user decide when to bookmark.

Bookmarked search strings populate the suggestions if no search string is entered into the input field.
Bookmarks are saved by context, so there are separate bookmarks for search, task-list and refile.
There are at most 10 bookmarks for a context. New saved bookmarks are inserted at the top of the list. If the list gets too long, the last search strings are dropped. Duplicate bookmarks are dropped too. The list of bookmarks is ordered by last used.

Bookmarks are at the moment unaware of file context. You always have the same bookmarks. It would be possible to change that, but I think it would not be intuitive to use.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -7,8 +7,16 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2021-12-05 Sun]
+** Added
+   - *EPIC* Bookmark functionality for search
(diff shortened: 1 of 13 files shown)


munen:
@tarnung I love it :heart: 

Simple, but super powerful implementation! I'm sure many people will be able to make good use of this. I certainly will, so I'll merge right away^^

I have updated the help and changelog accordingly.

PR #763, opened 2021-12-06
fix title editor text loss

tarnung:
The title editor modal in semantic edit mode persists its content on different schedules.
- the todo keyword is saved on click
- the title text is saved on close

This leads to the following behaviour:
If you modify the text of a title and then click a todo, the todo keyword gets saved which triggers a rerender before the text got saved. Therefore the text is reset to the state it was in when the title editor was opened.

As a solution I save the title text before updating the todo-keyword.


munen:
@tarnung Good job! I confirm the bug and solution :pray: 

I've updated the changelog and will merge right away :bow:

Issue #764, opened 2021-12-09
Update Sign In page


munen:
- [ ] Make link to Github more dominant
- [ ] Decouple "Sign In" and Github icon. Right now it looks like they are coupled.
- [ ] Bring GitLab into the list of sync back-ends

amm98d:
Hi, I would like to take this, if possible.

munen:
Hi @amm98d 

Thank you for volunteering :pray: 

Looking forward to the Pull Request.

Happy hacking^^

amm98d:
Hey @munen, I'm facing issues in setting up the environment. I have a windows system, thus I used [nvm-windows](https://github.com/coreybutler/nvm-windows). However, most of the codebase is still Linux-based with .sh files, etc...not sure how to proceed around that.

tarnung:
Hi @amm98d a year ago I worked on organice on a windows system. I had some small issues with line endings and such, but that should be taken care of by the git config. You mostly only need 'yarn start' (and 'yarn test') to fire up the organice dev instance. No need to dig into the scripts behind that. Do you have a specific problem with something not working?

A short while ago I did again set up my organice dev environment. This time I cloned the repo inside of WSL and I use the vscode to wsl connector. That works fine as well.

munen:
Thanks for jumping in, @tarnung. @amm98d He is a maintainer, too.

Generally speaking: I'm afraid that I haven't used Windows a whole lot since '99, so I cannot be a big help. From what I hear, [WSL](https://docs.microsoft.com/en-us/windows/wsl/install) provides a proper Linux environment in Windows 10 and 11. If you don't have that, I used to use [Cygwin](https://www.cygwin.com/) a long time ago.

On the other hand, tarnung is right - as long as you can run `yarn start`, you should be good to go for many tasks on organice(;

amm98d:
Thanks for the input guys @tarnung @munen, I actually did try to go past all the scripts and Linux stuff and run `yarn start` as that is all I would be needing. However, there was an error in that. I will share it in a short while. 

P.S. I have worked with WSL in the past, if nothing works, I will switch to that.

PR #767, opened 2021-12-12
Update Sign In page

amm98d:
## Changes:
- [x]  Make link to Github more dominant
- [x]  Decouple "Sign In" and Github icon. Right now it looks like they are coupled.
- [x]  Bring GitLab into the list of sync back-ends

## Original:
original

## New:
1em - 2x

--
P.S. Let me know if changes needed.
Referenced Issue: #764

### DIFF
--- a/src/components/HeaderBar/index.js
+++ b/src/components/HeaderBar/index.js
@@ -281,7 +281,7 @@ class HeaderBar extends PureComponent {
 
           {!isAuthenticated && (
             <ExternalLink href="https://github.com/200ok-ch/organice">
-              <i className="fab fa-github header-bar__actions__item" />
+              <i className="fab fa-github header-bar__actions__item fa-2x" />
             </ExternalLink>
           )}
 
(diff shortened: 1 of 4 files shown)


amm98d:
Hey @munen, let me know if the PR is okay or needs any updates...

munen:
Hi @amm98d 

Good job on your first PR; congratulations! :tada: 

I realize the first two todos are unclear in the original issue. They say:

- Make link to Github more dominant
- Decouple "Sign In" and Github icon. Right now it looks like they are coupled.

Those two are coupled. Having a "Sign in" link next to a Github icon implies that the sign in kind of works with Github. That's not true. Now that you've made the icon bigger, the implication is even more significant(; But good job on finding out that you could do that with only a change to the markup!

The solution would be to remove the GH icon in the top right and instead add a button for GH after the "Documentation" button.

Minor nitpick: Since we're updating the list of sync back-ends, we could use the same order as we do in the documentation: Dropbox, GitLab, WebDAV, and Google Drive.

After adding those two small changes, I'm happy to merge in your very first PR :pray:

munen:
Thank you for the effort :pray: 

This PR is made obsolete by the new LP https://github.com/200ok-ch/organice/pull/809

Issue #732, opened 2021-10-16
Deployment under a URL subfolder


alethiophile:
**Describe the bug**
I'm trying to deploy Organice under a URL subfolder. There's a Nextcloud instance hosted on another subfolder of the same domain, and the goal is to be able to use Organice on those files without having to loosen the CORS settings. Having Organice hosted under the same subdomain should allow this due to same-origin (AFAICT).

However, it seems that Organice always assumes it will be hosted at the root of the subdomain. When I configure my server to host the Organice build directory at the URL subfolder `organice`, it successfully downloads the index.html, but all requests for static resources go to `/static/js/etc` directly, rather than `/organice/static/js/etc`. This makes organice fail to load.

I've tried adding a `homepage` directive to the `package.json`, as described in the Create React App docs [here](https://create-react-app.dev/docs/deployment#building-for-relative-paths). When I build with this set, organice loads successfully; however, for some reason, on first loading `index.html`, the client-side router redirects me to the root of the subdomain (rather than `/organice/`). How do I make it use the correct URLs on the client side?

**To Reproduce**
1. Configure your Web server to serve organice under a URL subdirectory e.g. `/organice/`
2. Watch it fail to load all the static assets

munen:
Relabeled, because this is a change request, not a bug in existing behaviour.

munen:
Hi @alethiophile 

Thank you for opening a new issue. Are you planning on working on this?

munen:
Closing this issue as per https://github.com/200ok-ch/organice/blob/master/CONTRIBUTING.org#how-we-work-with-issues. If you choose to work on it at some point later, please re-open the issue.

PR #771, opened 2022-01-11
Enable hosting from subdirectory

Pascal-So:
As a server admin I want to host organice from a subdirectory URL in order to arrange my server's routes to my liking and to have more technical flexibility (i.e. host organice from the same domain as nextcloud so that I don't have to set up CORS).

This fixes issue #732.

This implementation uses [CRA's `homepage` feature in `package.json`](https://create-react-app.dev/docs/deployment/#building-for-relative-paths). The person wanting to host organice from a subdomain should then adjust the `homepage` value to the full desired domain and path, for example `"homepage": "https://example.com/organice/"`.

After running `yarn build` as usual, they can then upload the resulting build output to that subdirectory on their server. Before this change, this was not possible because the router would have kept redirecting to the "/" route, the assets were not found, etc.

Internally, CRA takes the `homepage` value and extracts the path part of it, which is then assigned to `PUBLIC_URL`. It would also be possible to build organice without adjusting the `homepage` value and instead just calling `PUBLIC_URL=/organice yarn build`.

In my PR I left the `homepage` value as `""` which should be safe as that should lead to the same behaviour as not specifying it. See the relevant implementation in CRA [here](https://github.com/facebook/create-react-app/blob/9673858a3715287c40aef9e800c431c7d45c05a2/packages/react-dev-utils/getPublicUrlOrPath.js#L47) (the code doesn't check for undefined, it just casts to bool). 

Do you want me to add some documentation for this? If yes, where? Should I add a new section in the FAQ for this?

### DIFF
--- a/src/components/SyncServiceSignIn/index.js
+++ b/src/components/SyncServiceSignIn/index.js
@@ -35,7 +35,7 @@ function WebDAVForm() {
             persistField('webdavEndpoint', url);
             persistField('webdavUsername', username);
             persistField('webdavPassword', password);
-            window.location = window.location.origin + '/';
+            window.location = window.location.origin + process.env.PUBLIC_URL;
           }}
         >
           <p>
@@ -165,9 +165,11 @@ export default class SyncServiceSignIn extends PureComponent {
       clientId: process.env.REACT_APP_DROPBOX_CLIENT_ID,
       fetch: fetch.bind(window),
     });
-    dropbox.auth.getAuthenticationUrl(window.location.origin + '/').then((authURL) => {
-      window.location = authURL;
-    });
+    dropbox.auth
+      .getAuthenticationUrl(window.location.origin + process.env.PUBLIC_URL)
+      .then((authURL) => {
+        window.location = authURL;
+      });
   }
 
   handleGoogleDriveClick() {
@@ -185,7 +187,7 @@ export default class SyncServiceSignIn extends PureComponent {
 
             gapi.auth2.getAuthInstance().signIn({
               ux_mode: 'redirect',
-              redirect_uri: window.location.origin,
+              redirect_uri: window.location.origin + process.env.PUBLIC_URL,
             });
           });
       });
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "engines": {
     "node": "^12.13.1"
   },
+  "homepage": "",
   "dependencies": {
     "@bity/oauth2-auth-code-pkce": "^2.13.0",
     "bowser": "^2.11.0",
--- a/src/App.js
+++ b/src/App.js
@@ -216,7 +216,7 @@ export default class App extends PureComponent {
   render() {
     return (
       <DragDropContext onDragEnd={this.handleDragEnd}>
-        <Router>
+        <Router basename={process.env.PUBLIC_URL}>
           <Provider store={this.store}>
             <div className="App">
               <Entry />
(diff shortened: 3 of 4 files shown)


Pascal-So:
hmm, [circleci fails with `error 'process' is not defined`](https://app.circleci.com/pipelines/github/200ok-ch/organice/1721/workflows/9278e0e9-0086-4841-a65e-b41359c82419/jobs/3641/parallel-runs/0/steps/0-106)...

But I noticed that the code is already using the `process` object in `SyncServiceSignIn`, so does anyone know why this fails for my code?

Pascal-So:
Ah I see, we need to tell eslint that the `process` global exists: https://eslint.org/docs/user-guide/configuring/language-options#specifying-globals

Instead of adding the `/* global process */` comment everywhere, would it make more sense to add that to the `.eslintrc.yml` file instead?

src/components/SyncServiceSignIn/index.js
```diff
@@ -35,7 +35,7 @@ function WebDAVForm() {
             persistField('webdavEndpoint', url);
             persistField('webdavUsername', username);
             persistField('webdavPassword', password);
-            window.location = window.location.origin + '/';
+            window.location = window.location.origin + process.env.PUBLIC_URL;
```
schoettl, line 38: How does it behave if PUBLIC_URL is undefined instead of `/`?

Pascal-So: Ah, good catch, thanks. It just turns into "" so the trailing slash would be missing. It looks like trailing slashes are removed from the `PUBLIC_URL` anyway (for both the `homepage` method and directly specifying `PUBLIC_URL`) so I'll simply re-add the `+ '/'` after the `PUBLIC_URL`.

schoettl: Alright, but does it also work for `export PUBLIC_URL='/'` (which should be allowed IMO)? Sorry for that annoyance; I also don't have a solution by hand right now...

Pascal-So: Yes I also tested that, this behaves the same as the other cases i.e. the slash is removed and `PUBLIC_URL` ends up as "" which is what we want.

Pascal-So:
I was also wondering what would happen for a subdirectory that contains the string 'files', see this snippet in [`renderFileBrowserBackButton()`](https://github.com/200ok-ch/organice/blob/47f2ccbe34b72900765b3ab9d7bd5ba4ee81bd60/src/components/HeaderBar/index.js#L56):

```js
const fileParts = window.location.href.split('/').map((e) => decodeURIComponent(e));
if (_.includes(fileParts, 'files')) {
  backPath = _.last(fileParts);
}
```

Hosting this at `https://example.com/organice/files` now leads to the problem that the settings page renders like this: 

image

instead of like this:

image

There are however two remarks I have about this situation:

1. It doesn't seem very likely that someone would use `files` as one of the directory names in the path to their organice instance, worst case we can warn against it in the documentation.
2. The back button text appears to be misappropriated for the directory name when browsing files. In my opinion the back text should always be "Back", and the current directory name should be printed somewhere else:
  a. either centered in the title bar just like the "Settings" text on the settings page
  b. or if that leads to problems with the directory name being too long then maybe add a breadcrumb line below the title bar. That way the whole path would be visible, not just the name of the current directory.

With the directory name being displayed somewhere else, we could then change the `renderFileBrowserBackButton()` to always display the "Back" text instead of something dynamic.

image
  
Therefore I don't think this issue is significant.

PR #781, opened 2022-02-23
Fixed manifest for PWAs, fixes #779

### LINKED ISSUE #779: PWA not installable
I cannot install organice as a PWA on my mobile browser (the "install" button does not appear in any browser, while it works on other PWAs like https://mdn.github.io/pwa-examples/a2hs/)

On desktop, Chrome & firefox both report several errors with the app manifest (Developer tools -> Application -> Manifest)
- "Manifest start URL is not valid"
- "Manifest does not contain a suitable icon [...]"

I have debugged the issue locally, and it seems to be a problem with the dynamically generated blob url, introduced in https://github.com/200ok-ch/organice/pull/465 . Apparently the relative urls (`"/"` for `start_url`, or `"organice-512x512.png"`for the icon `src` fields) are not resolved correctly when they appear inside `createObjectURL`?
When I manually change `start_url` and `src` to absolute urls, the errors from firefox & chrome disappear and both report the PWA as "installable".
The errors also disappear when the JS logic is removed, and a static `manifest.json` with relative urls is referenced in the `<link>` tag.

A possible fix could be to create absolute urls on the fly, that should not lead to any issues I guess? I can create a pull request for that if you like.

**To Reproduce**
1. Go to https://organice.200ok.ch/files
2. Open developer tools: Application -> Manifest
3. Error messages appear, PWA is shown as "not installable"

**Expected behavior**
PWA is installable, Browser report no errors with the app manifest

**Screenshots**
image

Test with manually set absolute urls (with the same JS logic):
image

 - OS: Linux
 - Browser: Firefox 97.0.1 (64-bit), Chromium 98.0.4758.102 (Official Build) Manjaro Linux (64-bit)

tbruckmaier:
As discussed in #779 

Since the manifest generation happens inline, `absoluteUrl()` is defined in the global namespace, hopefully does not lead to problems.

I have based the PR on `master`, if it shall be based on another branch, I'm happy to change it!

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -48,27 +48,31 @@
     <script>
      var {pathname} = window.location;
 
+     const absoluteUrl = function(path) {
+        return new URL(path, document.baseURI).href;
+     };
+
      var manifest = {
          "short_name": "organice",
          "name": "organice",
          "icons": [
              {
-                 "src": "organice-512x512.png",
+                 "src": absoluteUrl("organice-512x512.png"),
                  "sizes": "512x512",
                  "type": "image/png"
              },
              {
-                 "src": "organice-192x192.png",
+                 "src": absoluteUrl("organice-192x192.png"),
                  "sizes": "192x192",
                  "type": "image/png"
              },
              {
-                 "src": "favicon.ico",
+                 "src": absoluteUrl("favicon.ico"),
                  "sizes": "64x64 32x32 24x24 16x16",
                  "type": "image/x-icon"
              }
          ],
-         "start_url": pathname || "/index.html",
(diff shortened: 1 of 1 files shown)


munen:
@tbruckmaier Thank you for the fast implementation :100:. I have deployed it to https://staging.organice.200ok.ch/ and confirm that:

- The manifest error/warnings in Firefox are gone
- Installation on iOS still works

Happy to merge, thank you for the contribution :pray: :bow:

munen:
I have also added the fix and attribution to the changelog: 4903142e

PR #138, opened 2019-12-17
[WIP] Feature: Webshare

dotcs:
This PR tries to add [webshare support](https://wicg.github.io/web-share-target/level-2/). It's currently WIP. This addresses issue #52. Please join the discussion there.

PR #795, opened 2022-05-12
WIP: organice as share target

tbruckmaier:
Hi,

this WIP pull request aims to make organice show up as a share target on android, as described in #52

First of all, I cannot get *any* PWA to show up as share target on my phone, not even the sample PWAs like https://github.com/GoogleChrome/samples/blob/gh-pages/web-share/README.md#web-share-demo
That being said, I won't be able to finish this PR since I can neither use nor test it properly.
Maybe someelse wants to finish the PR though (there is not much left to do imo). Or maybe the new `/capture` page is useful on its own already.

Implemented so far:
- a new route `/capture` is implemented, which lets the user select capture template & target file (capture template settings are used here, e.g. which capture template is available in which files). Either one or both parameters can also be passed as URL parameters, the form is then populated with these values
- on submitting the form, the user is forwarded to the selected file & the capture process is triggered
- any `captureVariable_` url parameters are forwarded through the process, so you can open the initial page `/caputre?captureVariable_foo=bar`, select file & template interactively, and the capture variables are still used in the following capture process
- "Share targets" on android receive three parameters: `title`, `text`, `url`: these will be mapped into the capture variables `v_title`, `v_text`, `v_url` (see below for the strange variable names)

TODOs
- The url `/capture` should be used as the share target, with the `captureVaraible_*` parameters filled. I have added this already to the manifest, so it *should* work already. No idea why this does not work on my phone, but maybe someone else can test it.
- probably the capture page can be styled a bit nicer

Some open questions / further ideas:
- the replacement of capture variables seems odd to me (or maybe I was doing something wrong?): the single letter placeholders like `%u` or `%t` get replaced *before* any custom variables (like `captureVariable_url` / `%url`), so there can be no capture variables starting with certain letter (for example `%title` woiuld be replaced to `[1970-01-01]itle`). Is that behaviour intended? Or am I doing something wrong? I have prefixed the capture variables with `v_` meanwhile
- this PR makes organice show up as a *single* share target. It might also be possible to make certain capture templates show up as share targets on their own (mabye with a template setting). This should be a separate issue imo
- when the user submits the form on the `/capture` route, the capture process is directly triggered and the captured content saved to the file. Under some circumstances it could be preferable to show the capture popup and let the user change the captured content interactively

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -75,8 +75,16 @@
          "start_url": absoluteUrl(pathname || "/index.html"),
          "display": "standalone",
          "theme_color": "#000000",
-         "background_color": "#ffffff"
-     }
+         "background_color": "#ffffff",
+          "share_target": {
+            "action": "/capture",
+            "params": {
+              "title": "captureVariable_v_title",
+              "text": "captureVariable_v_text",
+              "url": "captureVariable_v_url"
+            }
+          },
+     };
 
      const manifestBlob = new Blob([JSON.stringify(manifest)], {type: 'application/json'});
      const manifestURL = URL.createObjectURL(manifestBlob);
--- a/src/components/Capture/index.js
+++ b/src/components/Capture/index.js
@@ -50,7 +50,7 @@ const Capture = ({
           <p>selected template: { template.get('description') }</p>
 
           <Link
-            to={`/share?${getQueryString()}`}
+            to={`/capture?${getQueryString()}`}
           >select different template</Link>
         </div>
       );
@@ -64,7 +64,7 @@ const Capture = ({
             <li>
               <Link
                 key={template.get('id')}
-                to={`/share/${template.get('description')}?${getQueryString()}`}
+                to={`/capture/${template.get('description')}?${getQueryString()}`}
               >{template.get('description')}</Link>
             </li>
           ))}
--- a/src/components/Capture/index.js
+++ b/src/components/Capture/index.js
@@ -1,0 +1,207 @@
+import React, { Fragment, useState, useEffect } from 'react';
+import { connect } from 'react-redux';
+import { useParams, Link, useHistory } from "react-router-dom";
+// import './stylesheet.css';
+
+import { List, Map } from 'immutable';
+import { STATIC_FILE_PREFIX } from '../../lib/org_utils';
+
+import parseQueryString from '../../util/parse_query_string';
+
+const Capture = ({
+  captureTemplates,
+  loadedFilePaths,
+  state,
+}) => {
+
+  /**
+   * Return a list of possible files for a capture template. Some templates
+   * force a certain file, some allow a specific whitelist and some allow all
+   * files
+   */
+  const getPossibleFilesForTemplate = (template) => {
+    if (!template) {
+      return [];
+    }
+
+    if (template.get('file')) {
+      return [template.get('file')];
+    }
+
+    if (!template.get('isAvailableInAllOrgFiles')) {
+      return template.get('orgFilesWhereAvailable').toJS();
+    }
+
+    return loadedFilePaths.filter(path => !!path);
+  };
+
+  /**
+   * Render the template selector or the current template
+   * 
+   * If no template is selected, show the list of all templates and let the user
+   * select one. If a template is selected, show the selected template and a
+   * button to unselect it
+   */
+  const renderTemplateSelector = () => {
+
+    if (template) {
+      return (
+        <div>
+          <p>selected template: { template.get('description') }</p>
+
+          <Link
+            to={`/share?${getQueryString()}`}
+          >select different template</Link>
+        </div>
+      );
+    }
+
+    return (
+      <div>
+        <p>select a template:</p>
+        <ul>
+          {captureTemplates.map((template, index) => (
+            <li>
+              <Link
+                key={template.get('id')}
+                to={`/share/${template.get('description')}?${getQueryString()}`}
+              >{template.get('description')}</Link>
+            </li>
+          ))}
+        </ul>
+      </div>
+    );
+    ;
+  }
+
+  /**
+   * Render the file selector, if a template is selected. Depending on the
+   * template, show a dropdown with possible files or just an automatically
+   * selected file
+   */
+  const renderFileSelector = () => {
+
+    if (!template) {
+      return null;
+    }
+
+    return (
+      <select onChange={(e) => setCaptureFile(e.target.value)} style={{ width: '90%' }} value={captureFile}>
+        {getPossibleFilesForTemplate(template).map((path) => (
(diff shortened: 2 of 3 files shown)


chasecaleb:
I'm not a maintainer here, but I think this would be a great feature!

Unfortunately I have too many other things going on so I don't think I'll be able to contribute to this anytime soon, so hopefully someone else can. I could at least do some quick testing if/when this gets further along though.

tbruckmaier:
Hi @chasecaleb , I have deployed my branch on https://organice.24hoursparty.de/. Would you be open to install that page as PWA and see if sharing works for you? That would at least give us a hint if the missing share functionality is a issue with my phone or the code.

munen:
Hi @tbruckmaier 

First, thank you for coming up with such a fantastic PR!🙏🏻

Unfortunately, I’m not an Android user myself, but I did install your deployed version to my iPhone to check if it would work there. According to https://caniuse.com/web-share, it just might. Having said so, I haven’t used any PWA as a share target yet. But I’d very much like organice to work like that(;

Unfortunately, so far, I have had no luck with your installation. However, I also tried the Scrapbook PWA from your initial post, and it also didn’t show up as a share target. I tried finding out if I need to configure something somewhere but haven’t found anything so far.

munen:
@schoettl or @branch14 Does the [scrapbook share PWA](https://github.com/GoogleChrome/samples/blob/gh-pages/web-share/README.md) show up as a share target on your Android phones? If so, does organice show up after installing from https://organice.24hoursparty.de/?

schoettl:
Hi all, I just tried the sample PWA https://scrapbook-pwa.web.app/ on Android 10 and tried to share a photo. It doesn't show up as a share target :( I even tried a smartphone reboot.

munen:
That’s really interesting. So… the official Google samples don’t work, but the API still exists. Curious. 

Thank you for checking it out @schoettl 🙏🏻

tbruckmaier:
That is really strange, I have blamed the issue on my non-standard phone setup (quite old LineageOS 14 ~= Android 7, no google play services, root, Chromium instead of regular Chrome). @schoettl do you have a more regular setup? Also, is Chrome your default browser? (I think I have read someone having issues because of that).
And could you maybe try https://mconverter.eu/ ? It is also a PWA which accepts files through the android share menu, and they claim in a [recent blog post](https://mconverter.eu/blog/web_share_target_api/) (march 2022) that _"nearly a quarter of all daily conversions getting initiated through the share menu"_ . So surely we are doing something wrong?

@munen If we cannot solve this issue, do you think the new /capture route is beneficial nevertheless?

munen:
@tbruckmaier The demo video on https://mconverter.eu/ is pretty impressive. I'd love this workflow to be available on organice! I did try it (on my iPhone 13 pro) and it did not work. Having said so, they don't claim it should work on iOS(; That itself is not a blocker for inclusion to organice, of course. It only means I unfortunately cannot really help with the issue.

> If we cannot solve this issue, do you think the new /capture route is beneficial nevertheless?

I mean, the sharing API seems to be a thing (we had the first [draft PR in 2019](https://github.com/200ok-ch/organice/pull/138)). If anyone can get it to work, I'm all for pulling this in. If it won't work, then we can always leave the PR in a draft state for someone else to pick up later. Or do you see another possible use case for the `/capture` route? I'm open to suggestions, of course.

schoettl:
Oh, I forgot that my default browser on Android is Firefox!

Installing the demo app via Android Chrome works, now I've got the share target.

I'll try organice.24hoursparty.de tomorrow...

schoettl:
OK, I installed Organice from 24hours... as app in Chrome on Android 10. I can't sign in, Dropbox is not working and I can't get WebDAV working ATM, the docker test server is not reachable from other than localhost... although it should because it's bound to 0.0.0.0

But so far, Organice does not show up as share target. @tbruckmaier For which types of shares it's supposed to work? Webseites shared from browser? Files?

munen:
@schoettl It makes sense that Dropbox won’t work (unless it’s been set up as an app). We could merge to develop and deploy to Staging for that. 

However, it should show up as a sharing target anyway, I suppose. @tbruckmaier Or should it only show up when logged in?

branch14:
@munen I "installed" organice from 24hoursparty.de via Chrome and Firefox on Android. I tried to log in via Gitlab, which did not work. "Client authentication failed due to unknown client, no client authentication included, or unsupported authentication method." Organice did not show up as a share target.

munen:
@branch14 thank you for checking it out! That’s GitLab synchronization did not work is expected if OP did not set it up. However that should be independent of the shared target functionality.

@tbruckmaier Seems we have no dice on the matter so far 😢

tbruckmaier:
Thanks everyone for testing, what a pity that it still does not work.

A last idea which came to my mind, maybe the integration does not work with the dynamically generated manifest (right now, the manifest is generated in JS dynamically instead of the regular `manifest.json`? I have (re-)added `manifest.json` therefore.

I have deployed it to 24hoursparty.de again, would you all be so kind and test it again? @branch14 @schoettl Probably best to uninstall the PWA and reinstall it freshly.

And if that does not work, as a last straw, could you have a look at the changed manifest code? Maybe it is a stupid typo or other issue which I somehow don't see?

Regarding the login issues with dropbox and gitlab, I have not done anything special to that part of the code or the server setup. From my understanding, this should not interfere with the sharing part though.

schoettl:
I reinstalled the app from 27hours... and tried to share selected text, websites, and photos. No success, it doesn't show up :/

tbruckmaier:
Alright, thanks for the feedback, I am officially out of ideas.

@munen I leave it up to you what to do with this PR. Honestly, I will probably not be using organice a lot, since I mostly need something on the phone to capture notes, and will continue to use orgzly for that.

I personally would prefer to have something like the `/capture` route, which I built with the idea of emacs' interactive capture menu in mind. my idea was, whenever you want to capture something, you can use the same shortcut to that route, and then choose the appropriate template interactively. But since you are using the app a lot more than I do, you probably have a better idea if that is a useful addition.

munen:
For the moment, let's leave the PR here for someone to pick up at a later point in time and good luck to this future person!

tbruckmaier:
@munen Thanks for the tipp, that is indeed my backup plan, once [firefox on mobile supports bookmarklets again](https://github.com/mozilla-mobile/fenix/issues/2871) ;)

czlucius:
> @munen Thanks for the tipp, that is indeed my backup plan, once [firefox on mobile supports bookmarklets again](https://github.com/mozilla-mobile/fenix/issues/2871) ;)

I believe they already do, and that issue was just not closed.
Can try opening up a bookmarklet in Stable, and it will open up in the current tab and not a new tab.

PR #800, opened 2022-05-29
Assign labels to their inputs

schoettl:
Even if this is not a website but an app, I like it if labels are assigned to their inputs. Click/touch on the label will focus the assigned input.

What do you think?

And do we have special rules for choosing ID names?


munen:
Sounds good to me👍

As for naming conventions, we don’t have them, yet. But if you have a preference, you can create them^^

PR #802, opened 2022-05-30
fix: use input type url instead of text for gitlab url

branch14:
Input type text on Android will automatically insert a space after each dot. This is unwanted behavior when entering a URL. Input type url will AFAIK not yield any other benefits nor side effects apart from not inserting the space after the dot.

### DIFF
--- a/README.org
+++ b/README.org
@@ -81,7 +81,7 @@ homescreen follow these platform specific instructions:
 
   The exact procedure may differ depending on your browser and Android
   version.  If you discover improvements to the following procedure,
-  please [[#contributing][let us know]]!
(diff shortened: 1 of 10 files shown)


munen:
Lgtm :pray: 

Let's merge to develop and test on iOS and Android.

schoettl:
The webdav sign-in also has this input type url. But I think type url requires the protocol (https://) which was not required before... don't know how the webdav backend handles this value.

schoettl:
... so now, if I try to input an URL without protocol (like the placeholder text says), I got an browser error hint when trying to submit ("Sign-in").

schoettl:
I think type url is OK. I tested on Firefox (Desktop). But 1. and 2. should be fixed.

PR #805, opened 2022-05-31
chore: setup capacitor for android and document how to use it

branch14:
- [x] setup capacitor for android
- [x] document how to use it


munen (approved):
LGTM :partying_face:

PR #808, opened 2022-05-31
Refactor signin markup

branch14:
Based on #807 to avoid merge conflicts.

Fixes an issue where all content inside the signup form would have pointer as cursor.


munen:
LGTM! :+1:

Issue #127, opened 2019-12-01
Waiting for Google to put "Google Drive" API into production mode 


ahyatt:
**Describe the bug**
On signing into Google Drive, I get the following error:

"Sign in with Google temporarily disabled for this app
This app has not been verified yet by Google in order to use Google Sign In."

**To Reproduce**
1. Go to the organice page
2. Click on Sign In
3. Select Google Drive
4. Select account

**Expected behavior**
I would expect to be logged in

**Screenshots**
[Sign in Issue.pdf](https://github.com/200ok-ch/organice/files/3908747/Sign.in.Issue.pdf)

**Desktop (please complete the following information):**
 - OS: Mac OS X
 - Browser: Chrome & Safari

**Additional context**
My account has 2-factor authentication enabled, not sure if that affects the issue.

munen:
@ahyatt Thank you for the report. It actually does work for me (I just tried logging out and in again). That is not to say that there's something amiss, of course!

@branch14 Can you look into this or give me access to the Google APIs, please?

branch14:
@ahyatt Thanks for the heads up! :+1:

I guess it was working for users that had passed the consent screen. But not for new users. The Google Dev Console "hid" the reason in a popup: "Because you updated the application logo, your consent screen requires verification by Google."

I just resubmitted it for verification. :crossed_fingers:

I'll leave this issue open as it is not fully resolved, yet.

munen:
Update: This is what we see in the Google Cloud Console:

![image](https://user-images.githubusercontent.com/505721/72259118-f3263c00-360f-11ea-9f0c-b6bad310af26.png)

They are taking their time...

munen:
Update: Google still shows no update in the Google Cloud Console:

![image](https://user-images.githubusercontent.com/505721/73203301-12a78380-413d-11ea-99fc-805b90ecec09.png)


According to their estimate, it should have been processed any day in the last two weeks. Hence, it should be any day now(;

ahyatt:
I checked internally at Google, and evidently they have no record of this being a production system and no request for verification.  Perhaps this was requested in the wrong way, they referred me to this documentation: https://support.google.com/cloud/answer/9110914#submit-howto.

ahyatt:
Update: evidently my previous update was wrong - and was about another project named "Organice".  Evidently there are several.

At any rate, there was a mistake about not updating scopes during the request submission, and that needs to happen before this can be closed.  They will reach out to you and let you know what needs to be done.

[...]

Issue #109, opened 2019-11-21
Replace integration of Google Drive through <script> include in index.html with NPM module


munen:
organice adds the Google Drive API SDK through an old school `<script>` tag in the index.html file ([here](https://github.com/200ok-ch/organice/blob/master/public/index.html#L41)).

It would be better to put it as a proper dependency through `yarn add googleapis` and then follow the [documentation](https://www.npmjs.com/package/googleapis) to refactor the [sync backend](https://github.com/200ok-ch/organice/blob/master/src/sync_backend_clients/google_drive_sync_backend_client.js).

Rationale:

1. It's bad practice to add dependencies through in tags.
2. With the current integration, it requires the browser to allow third party cookies which is proper sane to have turned off for most folks.

munen:
Is anyone interested in taking responsiblility for the Google Drive integration? There's a number of open issues which only relate to Google Drive(#107, #109, #127).

From the current maintainers, nobody is using the Google Drive integration and nobody currently has time to spend on it. Generally speaking, the current integration works (if you're an early adopter on organice.200ok.ch or run your own instance to provide your own Google Drive API key). However, the three issues linked above prevent new users to use Google Drive easily.

It would be great if there's at least one person willing to take on proper Google Drive integration. Otherwise, I fear, that the easiest option would be to disable Google Drive until such a time comes.

If someone speaks up, I'd be absolutely willing to share information, help out, do pairing sessions (on Zoom or the like) and do code reviews.

asymmetric:
I think considering the cost of leaving GDrive on (i.e. the necessity of allowing 3rd party scripts), it's better to remove support for it ASAP, and re-add it once someone shows up to maintain it.

markamaze:
tried to implement googleapis npm package and ran into the same problems described in these issues:

[https://github.com/googleapis/nodejs-dialogflow/issues/716](url)
[https://github.com/googleapis/nodejs-dialogflow/issues/314](url)

munen:
@markamaze Thank you for looking into this :pray: 

The finding is unfortunate, though. From the first issue you're linking to, it looks like the googleapis package cannot be used from a create-react-app or next.js app at this point - maybe even not possible from an app using webpack: https://github.com/googleapis/nodejs-dialogflow/issues/314#issuecomment-479672058

[...]

Issue #107, opened 2019-11-15
Sign in to Google Drive from new versions of Chrome disfunctional


klapauciusisgreat:
When I try to sign in with Google drive, I can. However:

1) Chrome wants to prevent me, claiming the website is "unsafe", so need to go through 'advanced' and explicitly tell Chrome to ignore.

2) After signing in, I end up on the homepage that now looks identical to  when I was not signed in (e.g., shows sample, and sign in). When I now try to sign in again, the click on Google drive does nothing.

I have not looked deeper, but I checked with Chrome and Safari on MacOS (using the demo site at organice.200ok.ch)

I'm probably do something super stupid, but I'm stuck :) If this is a known problem, many thanks in advance for telling me.

klapauciusisgreat:
Actually, I was testing with the Chrome canary version, which gave me (among others) this warning:

[Report Only] Refused to create a worker from 'https://organice.200ok.ch/service-worker.js' because it violates the following Content Security Policy directive: "worker-src 'none'".

Sign in actually works on the main Chrome branch. Safari, OTOH, also did not work for me.

I'm not sure if I should close it, or if someone wants to look at this to prevent this from becoming a real bug as new chrome versions get rolled out.

munen:
Thank you for reporting this @klapauciusisgreat 

I do confirm that logging in to Google Drive doesn't work in newer versions of Chrome. It also doesn't work in Mobile Safari. It used to, though. Curious how Google is keeping users out when two of their products try to connect without informing the API admins.

I'm personally using Dropbox on a daily basis with organice, so I wouldn't have caught this until somebody spoke up. Thank you for doing so and providing debug information! :pray: 

I'll look into this as soon as I can. In the meantime, I recommend to use another browser. I tested access to Google Drive sucessfully from Firefox on the same machine where Chrome didn't connect. In case the same problem exists on Android/Chrome, using Firefox in the interim should also work. According to [another ticket](https://github.com/200ok-ch/organice/issues/89), I just heard that this combination is working.

Sorry for the inconvenience and thanks again for helping out!

munen:
Update: Turns out I _cannot_ reproduce this bug. I ran into another bug with the Google Drive integration for which I created a new issue #109 

But I'm also not running on Chrome Canary, I'm using Chromium from Debian testing (76.0.3809.100 (Developer Build)). And I can't test using Chrome Canary, because Google requires me to run a non-free operating whilst my dev boxes don't run Windows or macOS:

[...]

PR #817, opened 2022-06-03
fix: Remove Google Drive

munen:
Closes: #127, #107, #109

This removes Google Drive in code and documentation. The rationale is documented in the changelog.

@schoettl, @branch14 Any last words you want to say before we bury this for good? :smirk:


branch14:
So Long, and Thanks for All the Fish?

I like that this is classified as a "fix". :smile:

LGTM :+1:

PR #822, opened 2022-06-08
feat: login to gitlab from native app

branch14:


munen:
Successfully tested on my machine. Native functionality is coming up superbly :100: :rocket:

PR #825, opened 2022-06-09
fix: Rename sample demo file in bundled WebDAV image

munen:
Otherwise there's a routing conflict between the actual sample.org file and the demo file.

That means, that anyone who has a `sample.org` file might still run
into the problem. The better fix would be, of course, to treat the
sample.org file differently. However, that would be a bigger refactor which could be done by the person requiring to have a file in her root directory to be called `sample.org`.


branch14:
:+1:

PR #826, opened 2022-06-09
feat: login to webdav from native app

branch14:


munen:
Sweet, works as advertised! Now all sync-backends work in the native mobile app versions :100: :rocket:

PR #834, opened 2022-06-13
doc: Set a default WebDAV server

munen:


munen:
While developing a new feature and regularly logging out and in for tests, I realized that this feature is actually very helpful for development, too.

Nice PR you have created! It saves quite a bit of typing during development^^

PR #835, opened 2022-06-13
feat: transient env vars

branch14:
This uses an 2 step approach (**bait & switch**) to inject the values of environment variables post compilation.

Step 1: Scan `.env.sample` and build a dummy `.env` file (**bait**) that injects dummy strings that can be replaced post compilation.

Step 2: After compilation replace (**switch**) the dummy strings with the value of corresponding env vars. (This is cheap and runs on every start of the docker container.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN yarn install
 
 COPY . /opt/organice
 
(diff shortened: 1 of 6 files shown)


munen:
Looking great :sparkler:

PR #838, opened 2022-06-13
Added capture template variables to enable advanced dates

dmorlitz:
Added %r and %R to the capture template input form and substitution code, so you can specify dates with modifiers in capture templates, such as <%r ++60d>

Currently if you use %t you would get (as an example) <<2022-06-13> ++60d> (which is invalid) - and the same is true for %u but with brackets

### DIFF
--- a/src/components/CaptureTemplatesEditor/components/CaptureTemplate/index.js
+++ b/src/components/CaptureTemplatesEditor/components/CaptureTemplate/index.js
@@ -292,6 +292,12 @@ export default ({
             <code>%U</code> - Inactive timestamp, with date and time.
           </li>
           <li>
+            <code>%r</code> - Timestamp, date only, no surrounding punctation - for dates with modifiers.
+          </li>
+          <li>
+            <code>%R</code> - Timestamp, with date and time, no surrounding punctuation - for dates with modifiers.
+          </li>
+          <li>
             <code>%{'<custom variable>'}</code> - A custom variable from a URL param capture. See{' '}
             <ExternalLink href="https://organice.200ok.ch/documentation.html#capture_templates">
               the README file
--- a/src/components/CaptureTemplatesEditor/components/CaptureTemplate/index.js
+++ b/src/components/CaptureTemplatesEditor/components/CaptureTemplate/index.js
@@ -292,10 +292,10 @@ export default ({
             <code>%U</code> - Inactive timestamp, with date and time.
           </li>
           <li>
-            <code>%r</code> - Timestamp, date only, no surrounding punctation - for dates with modifiers.
+            <code>%r</code> - Timestamp, date only, no surrounding punctation.
           </li>
           <li>
-            <code>%R</code> - Timestamp, with date and time, no surrounding punctuation - for dates with modifiers.
+            <code>%R</code> - Timestamp, with date and time, no surrounding punctuation.
           </li>
           <li>
             <code>%{'<custom variable>'}</code> - A custom variable from a URL param capture. See{' '}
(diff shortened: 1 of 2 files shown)


dmorlitz:
I just want to start by saying that it is entirely possible that I am using capture templates in the wrong way.  I actually don't use Emacs at all - I am exclusively managing my Org mode file through organice (and nano for quick text edits)

What led me to look at capture templates was that I was bulk entering tasks from my previous solution (Kanboard) into organice.  I have a lot of recurring tasks so I was primarily using the + button to enter them and looking for a way to minimize the number of clicks that it took.

For example, I was entering birthdays (and other annual events) using this capture template:
* BIRTHDAY %?
  DEADLINE: <%r ++1y -30d>

In this case, through the + button I still had to change the description and date field - but by having the date without punctuation I was able to add the ++ and - parameters automatically.  I have since made this more generic by using (although I have only tested this in a very limited case for other input methods, like Siri):
* BIRTHDAY %?
  DEADLINE: <%<date> ++1y -30d>

I was also using this template for recurring tasks to enter monthly events:
* TODO Monthly - %?
  DEADLINE: <%r .+1m>

I am not sure what the equivalent Emacs usage would be - since that is not how I am managing my data.

I would be fine either way if you decided to merge or discard this change.  I will admit that (after thinking about it) other than the bulk data entry, I am not sure how often I would be using this feature..

munen:
> I just want to start by saying that it is entirely possible that I am using capture templates in the wrong way.

The way you've outlined is perfectly reasonable.

Some capture templates you may keep forever, others can be temporary like the bulk entry you have described.

> I actually don't use Emacs at all - I am exclusively managing my Org mode file through organice (and nano for quick text edits)

That is perfectly alright and actually great use. I know of some other users who use organice exclusively for Org mode. Since we see this user base growing, we are currently in a sprint to make organice more accessible to this kind of user - somebody who does not have a strong attachment to Emacs. When we are done, ultimately the user shouldn't even require a strong technical background. In that regard, we're pushing new features like 'create new Org file' from the file browser, creating a landing page to explain what organice is, and we are currently working on bundling organice as mobile apps for Android and iOS. There's more coming, too(;

Great to hear that you can make good use of organice as is. You're also a good reference in that regard. So if you feel that something important is missing (like this use-case for capture templates), please always speak up! I cannot promise we will/can implement all wishes, but getting feedback is ultimately very valuable and we are always happy to listen, learn and adapt.

> ...
> I am not sure what the equivalent Emacs usage would be - since that is not how I am managing my data.

Thank you for explaining your use-case with actual examples. All of them make sense. They are the scenario 1 that I described above. Personally, I haven't created a capture template with this kind of recurring task. Hence, I will need to think and research this for a bit.

> I would be fine either way if you decided to merge or discard this change. I will admit that (after thinking about it) other than the bulk data entry, I am not sure how often I would be using this feature..

If we cannot come up with a way how Emacs does it, this absolutely is a viable PR to pull in as is!(;

munen:
Here's my interim result. Your example:

```
BIRTHDAY %?
DEADLINE: <%r ++1y -30d>
```

Could look like this in Emacs:

```
* BIRTHDAY %?  
DEADLINE: <%(org-read-date nil nil \"\") +60d>
```

So basically, the solution is to use elisp to construct the date here. The `%(` starts an elisp expression. `org-read-date` might be a bit overkill actually. Leaving all arguments empty (nil, nil and "") takes away from most of its power. It's more powerful when some options are given, so you can pick a date from the calendar, for example.

For full reference, here's the docstring of org-read-date https://gist.github.com/munen/682ee46dfdf49776570c0bf2751c16b7

A more simple solution (if only 'today' is interesting) would be to use:

```
* BIRTHDAY %?
 DEADLINE: <%(format-time-string "%Y-%m-%d") +60d>
```

That just fills in the current date.

These options are more configurable than `%r`, but way less sexy to type for your use-case.

TL;DR: I'll go over it one more time with @branch14. But at this point in time, I think your PR is going to be merged :+1:

dmorlitz:
I don't entering something more complicated into the capture string, if I know what it is and it makes it easier on your end.  How would I use a capture string with your elisp example to convert the input to a date?  When I tried 'DEADLINE: <%(format-time-string "%Y-%m-%d") +60d>' organice just left that string in the org file.  Does it need to be processed inside emacs?  Or can organice convert that to an actual date for me?  (I don't use emacs right now, so I am hoping organice can process that field somehow)

munen:
@branch14 agrees. This PR is a go. I'll make some changes to the docs and the changelog and will then pull it in.

PR #839, opened 2022-06-13
chore: minimal fastlane setup for android

branch14:
Test it with `cd android && bundle exec fastlane run validate_play_store_json_key`


munen:
Classic PR that looks easy, but takes two days to do. Great job and thank you for the extra effort :pray:

PR #844, opened 2022-06-14
refactor: `dirName` function

munen:
Use Node.js `path.parse(pathString)` ponyfill instead of custom code.


branch14:
TIL "ponyfill" :pray:

PR #847, opened 2022-06-16
fix: Conflated Routing

munen:
We are running different kinds of pages in our SPA (see [ADR-002](https://organice.200ok.ch/documentation.html#adr-002)).

Currently, the `<Entry />` component has conflated logic for rendering the LP and the actual application. This makes the code more complex and can yield unexpected side effects.

Todos:

- [x] Create a new component `<SwitchComponent />` that has the responsibility to render either `<Landing />` or `<Entry />`
- [x] Render the new `<SwitchComponent />` in `<App />`
- [x] Refactor relevant Routing logic from `<Entry />` to `<SwitchComponent />`
- [x] Remove LP boilerplate from `<Entry />` 
- [x] Do extensive manual testing to ensure that all routes still work 
- [x] Come up with better naming than SwitchComponent, it's just a working title
  - [x] @branch14 proposes `Turnout` as in railway lingo

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -29,7 +29,7 @@ import createGitLabSyncBackendClient, {
 
 import './base.css';
 
-import SwitchComponent from './components/SwitchComponent';
+import Turnout from './components/Turnout';
 
 import {
   listenToBrowserButtons,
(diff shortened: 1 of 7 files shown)


branch14:
Way better. :smile:

PR #853, opened 2022-06-21
feat: Handle empty files

munen:
Original issue: https://github.com/200ok-ch/organice/issues/583

This takes care of a big UX problem for users who solely rely on organice. So far, we've shown a parser error if the file is empty *or* if the file did not contain a headline.

Now, if the user opens a file that is completely empty, she has this flow:

Screen Shot 2022-06-21 at 17 13 40

On Click:

Screen Shot 2022-06-21 at 11 29 16

If the file already contains content (but not a headline), the user is shown this optimized screen:

Screen Shot 2022-06-21 at 17 13 21

Of course, this also works when the user deletes the final headline.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -7,6 +7,12 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2022-06-22 Wed]
+** Added
+- Ability to handle empty files
+  - This takes care of a big UX problem for users who solely rely on organice. So far, we've shown a parser error if the file is empty *or* if the file did not contain a headline.
+  - Now, if the user opens a file that is empty, or has no headlines, she will see the call to action to create a new headline.
+  - PR: https://github.com/200ok-ch/organice/pull/853
 * [2022-06-16 Thu]
 ** Added
 - Capture template variables =%r= and =%R= expanding to raw timestamps
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -315,6 +315,11 @@ export const addHeader = (headerId) => ({
   dirtying: false,
 });
 
+export const createFirstHeader = () => ({
+  type: 'CREATE_FIRST_HEADER',
+  dirtying: true,
+});
+
 export const selectNextSiblingHeader = (headerId) => ({
   type: 'SELECT_NEXT_SIBLING_HEADER',
   headerId,
--- a/src/components/OrgFile/OrgFile.integration.test.js
+++ b/src/components/OrgFile/OrgFile.integration.test.js
@@ -113,6 +113,18 @@ describe('Render all views', () => {
       getByPlaceholderText = res.getByPlaceholderText;
     });
 
+    describe('Works with Org files without headlines', () => {
+      test('Works with a completely empty files', () => {
+        store.dispatch(
+          parseFile(STATIC_FILE_PREFIX + 'fixtureTestFile.org', readFixture('empty_file'))
+        );
+        expect(queryByText('This file has no headlines')).toBeTruthy();
+        expect(queryAllByText('Yes, your file has content.').length).toEqual(0);
+        // Sanity check, ensure that not the regular test file is loaded.
+        expect(queryByText('Top level header')).toBeFalsy();
+      });
+    });
+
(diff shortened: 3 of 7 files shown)


dmorlitz:
I did have issues with blank files - mostly because I have not really used Org mode before using organice.  I ended up taking the sample file, putting it directly on my WebDAV server, and removing all of the content.  This gave me the header and now I just copy the file directly on the WebDAV server when I need to make a new one.  Definitely not the easiest.

I will also tell you that it took me some time to figure out that the file needed to have the .org extension to be recognized by organice.  I probably should have figured that out more quickly - but it definitely slowed me down for 30 minutes or so.

dmorlitz:
Oh - and all of my 5 org files have "ORG FILE - DO NOT DELETE" as their first header.

munen:
Oh my... I figured that it would be a pain, but that's even a bit worse than I thought. My apologies for the inconvenience🙏

Thank you for sharing your previous workflow. That gives me extra motivation to ship these new features!

Since last week, you can create new files right from organice. If you don’t add a .org extension, organice adds it automatically, so these 30min should be saved for the next user. 

With this PR, organice will work with empty files, files with content and headlines and files with content, but without headlines. That means it’ll still work when you delete the last headline.

I hope that takes care of all the various workarounds. Again, I apologize for the complicated experience. Some background: organice started out as a tool for power Emacs users who wants to have access to their Org files on the go. For this use case, it has proven to be a solid choice and has been for some years. In the meantime, the user base grew beyond old school Emacs users. Even though organice is rock solid for the old school Emacsers use case, there’s certainly big improvements to be done for the new user group. 

With this change in user base, we are committed to remedy that and support organice for everyone in an equal manner.

munen:
Thank you for your kind words🙏

It's great to see that you have arrived at using organice after first being seriously committed to another tool and then evaluating a lot of others. Kudos btw for the contributions to Kanboard👍

Your Inbox solution sounds like a wonderful workflow, btw. I do know some people who actually do still email themselves important todos even though they have dedicated PM tooling with an inbox. If you wanted to share that the PHP script in [contrib](https://github.com/200ok-ch/organice/tree/master/contrib), I'd be happy to pull that one in!

branch14:
@dmorlitz Thank you for sharing your use cases so openly. This is very interesting to read.

@munen This will indeed be much better for onboarding users that do not bring their own org files (or for some reason do not use headlines). :100:

dmorlitz:
My goal is really to get away from managing tasks by mail.  The reason I have these scripts are so that I can set up mail filters which forward messages to a webhook (through Mailgun, IFTTT or something like that) so they end up in my Org file.  I use it for things like bills, bank statements, integrations from other systems that produce notifications, etc.....

I will post the scripts and some instructions in the contrib/ folder later today.

munen:
Your use cases make total sense. Org mode is great for them!

I do not want to spoil too much, because it's not yet 100% certain, but on the `develop` branch we've got organice ported to a functional Android app (iOS will also be worked on). That new version already has a POC included so that the user can share different MIME types into organice. There's nothing happening with these specific shares, yet, other than being logged. But for every MIME type, the idea is that the user will be able to capture them - either in a default inbox or with custom capture templates.

That would enable the use cases that you outline above. Personally, I'm already doing all of those in Org mode since.. at least 6 years (plus some more specialized ones like tracking expenses from email bills). However, I'm currently using Emacs to do it. With the upcoming functionality, it should totally be not just possible, but nice to do it with organice(;

But again, I don't want to say too much. It's not done, yet. There can be issues along the way that we're currently not aware of.

dmorlitz:
That sounds wonderful.  I am really glad I found this project.  I am forced to use iOS (unfortunately) for work, so my device is an iPhone (I do have many Android devices laying around at home).  I currently have beorg installed to synchronize Org files - mostly because this app generates alarms.

The model you described would work very well for sharing content from a mobile device directly into Organice.  I am not sure I see a "server side" way to do this (and I do understand that Organice does not have a server side).  The reason I mention this is because if the code you built for Android was modular in the right way, you could effectively have Organice run behind a web-server and listen for automated requests (like mail forwards, IFTTT messages and other external triggers) if the right APIs were available.

Then you could effectively make a server act like an Organice client-side app and rely on the synchronization of the backend so that you could have unattended actions take place.

PR #854, opened 2022-06-21
contrib/organice-webdav-traefik: Fix docker-compose file and update README.org

edgar-vincent:
The code from the previous commit hadn't been updated in a while and had stopped working - this fixes it.


munen:
Hi @edgar-vincent 

Thank you for maintaining your original contribution :pray: 

Your setup looks very nice, indeed! At 200ok, we're using a similar docker-compose setup for different products and projects.

Of course, I'm happy to pull the update :rocket:

PR #856, opened 2022-06-21
doc: Fix several links for the documentation

munen:
Prior, they worked only in Emacs when the repo was checked out, not in the online documentation.


branch14:
:+1:

PR #857, opened 2022-06-22
Contrib webdav post

dmorlitz:
Here are the script files I use for posting tasks directly to Org files via email and web browsers.  The script does some limited data validation and removes dangerous characters.  The content is also formatted and indented properly for the submitted content.

The logic is pretty simple and could possibly be written in other languages.  I am currently using 2 PHP files and 1 shell script.

### DIFF
--- a/contrib/organice-webdav-post/Readme.md
+++ b/contrib/organice-webdav-post/Readme.md
@@ -1,0 +1,31 @@
+* Receive an HTTP post with a message to post to an Org file in organice
+
+These scripts *directly* update Org files on your WebDAV server
+
+*NOTES*:
+  - This set of scripts was designed to be modular and used
+  - There is limited content checking - so please be careful with what you post
+  - Watch out for race conditions if you edit your Org files in multiple places simultaneously
+
+** Configuring the scripts
+
+In this example, the file =home.php= contains all of the configuration information.  There should be no reason to edit the other files, unless you wish to change some default behaviours.
+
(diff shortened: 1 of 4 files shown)


munen:
Sweet, thank you for sharing :pray: 

With the nice documentation you've provided, I can absolutely see other people taking it up and making good use of it!

Regarding having organice as a back-end, we'll be keeping it in mind. The additional use cases are definitively interesting. For the moment, organice being 'just' a SPA gives the user the certainty that their data is 100% free and will never be seen by an entity other than them and their storage provider. But @branch14 and I discussed your comment and looked at your code and we see a lot of value that could be created. We will keep it in mind :+1:

dmorlitz:
I am glad that other people could benefit from the code that I sent you.

As to the "back-end" - that might have been poor wording.  I know that the storage provider will always be the "back-end".  I was just trying to suggest keeping in mind that people could run organice as a headless-SPA.  You already have the capture templates and URL parameters for posting tasks (as I recall).  If I could somehow store the back-end credentials on a server that I run so that I could have Apache (or another web server) invoke the local URL for organice, then I can have requests being handled by organice and benefit from the storage back-ends and other parsing features already put into place.

Right now, I just don't have a great way to have an Apache server on a system I host connect to either my iPhone or laptop (not on a public IP) to make those automatic connections.

Just a different way of thinking about deploying the app.......with the security concerns that come with storing credentials any place.

Issue #843, opened 2022-06-14
Undo looks like it is persisting changes, but it does not


munen:
The UI looks like it is making a synchronization and there are HTTP requests being dispatched. But there's no change in the actually persisted file.

Corresponding: Reloading the page will bring back the version of the file before undoing. 

Tested with WebDAV and Dropbox.

PR #861, opened 2022-06-28
fix: Undo/Redo persists to storage, again

munen:
Closes: #843

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -8,6 +8,11 @@ All user visible changes to organice will be documented in this file.
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
 * [2022-06-29 Wed]
+** Fixed
+- Undo looks like it is persisting changes, but it persisted old
+  state. Hence, reloading the file was bringing back the version of
+  the file before undoing.
+  - PR: https://github.com/200ok-ch/organice/pull/861
 ** Changed
 - Simplify UX in task modal
   - "Change todo keyword set" is now an explicit button. Also, it's
--- a/src/actions/org.js
+++ b/src/actions/org.js
@@ -161,17 +161,15 @@ const doSync = ({
 
       if (isAfter(lastSyncAt, lastServerModifiedAt) || forceAction === 'push') {
         if (isDirty) {
-          const contents =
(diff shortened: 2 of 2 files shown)


munen:
@tarnung I found a bug that was introduced in the multiple file epic. If you have the time, I have a question about it.

Before this PR, the UI looked like it is synchronizing and there are HTTP requests being dispatched. But there's no change in the actually persisted file. Actually, it does push changes, but it always sends the  the old state (before undo).

Corresponding: Reloading the page will bring back the version of the file before undoing.

I found the reason in `doSync()` in `actions/org.js`. There's a new line from your PR [EPIC: Multiple files](https://github.com/200ok-ch/organice/pull/570/files#). This line finds the content that is to be persisted. Before, we used the state from the store. Newly introduced was the preferred option to look at what's in `localStorage`. However, in the case of undo/redo `localStorage` is always outdated. It does not know the concept of Redux Past/Present/Future. Hence, removing this line fixes this particular bug.

Having said all that, I'm quite certain that there's a good reason why you introduced the `localStorage` bit. If so, then I'm fixing one bug by introducing another unknown one(; Do you by chance remember why you've introduced it? 

I tried thinking of a reason, but couldn't come up with it. In my understanding, when reaching this line of code, the Redux store is always populated and I don't see when `localStorage` should have preferrable information in it.

P.S.: Looking at the PR with "hide whitespace" helps readability. I actually only deleted the `localStorage.getItem('files__' + path)` bit.

tarnung (approved):
As stated above this was most likely a misguided effort to do things a bit more efficiently.
Since introducing a special case here to check if it's an undo/redo does not help code comprehensibility, i suggest to go with the change as proposed.

munen:
Nice, thank you for looking it that fast! 🚀🙏🏻

As for speed, I don’t think we lose anything here. The Redux store is already hydrated at that point. Looking up the file content is pretty much a hash lookup either in Redux or localStorage.

tarnung:
@munen Well you do call exportOrg one more time, which transforms the redux state to its file representation thereby doing a lot more than a map lookup.
But i very much agree that's a cost worth paying for fixing that bug.

PR #873, opened 2022-07-01
Added setting to change textbox height in the description editor

dmorlitz:
The settings screen works just fine - but the Description Modal does not yet load the setting and use it - so the default height is always used.


munen:
I cannot push to your branch:

```
To github.com:dmorlitz/organice.git
 ! [remote rejected]   dmorlitz-master -> dmorlitz-master (permission denied)
error: failed to push some refs to 'github.com:dmorlitz/organice.git'
```
So I created a patch for you with `git format-patch -1 HEAD`:

[0001-fix-Add-React-Redux-to-add-necessary-props-to-compon.patch.gz](https://github.com/200ok-ch/organice/files/9030354/0001-fix-Add-React-Redux-to-add-necessary-props-to-compon.patch.gz)

I used `gzip` on it, because GitHub won't allow `.patch` files to be uploaded to comments. Tbh, I'm not sure what the correct workflow is here. Usually I'm allowed to push to the PRs of contributors. I apologize if I'm making this more complicated than necessary.

The patch contains the [React Redux](https://react-redux.js.org/) bridge to get the required value into the props of the component. You were very close already.

munen:
I realize this is a total Draft PR, but I'll add some notes to the code. If you wanted to take care of those, anyway, just ignore them(;

munen:
You've added a `.swp` file to the PR. It's probably just an accident, because you wanted to create the PR quickly(; But those accidents happen. Personally, I have added such tmp files to my `~/.gitignore` files, because there's never a good reason to add them to git.

Having said that, kudos to using VIM! I've been a happy user for a good decade, as well^^

src/components/OrgFile/components/DescriptionEditorModal/index.js
```diff
@@ -17,6 +17,9 @@ export default class DescriptionEditorModal extends PureComponent {
       descriptionValue: props.editRawValues
         ? this.calculateRawDescription(props.header)
         : props.header.get('rawDescription'),
+      editorDescriptionHeightValue: props.editorDescriptionHeightValue
+        ? props.editorDescriptionHeightValue
+        : "8",
```
munen, line 22: When creating a new persistable field, we add it to https://github.com/200ok-ch/organice/blob/a7f57ce8520778625ba634617be681e2de6c117d/src/util/settings_persister.js#L74

There, you can also set a default.

This functionality is a bit hidden. Adding this to some part of the dev documentation could be helpful.

munen: Hah, I just saw that you actually added the flag there! Good job, so it wasn't too hidden for you(;

Just the default is missing.

Nice! :100:

dmorlitz:
I actually was basing my code on the agendaDefaultDeadlineDelayValue
setting - since that was the only number field on the Settings screen.
Then I just used grep to find all of the references to that variable, which
is how I found the persistent settings file.

I am not sure why you can not push to my PR.  I didn't change any settings
so I will have to go check my defaults.

I will also see if I can figure out how to remove the .swp file (my git
skills are limited)

munen:
> I actually was basing my code on the agendaDefaultDeadlineDelayValue setting - since that was the only number field on the Settings screen. Then I just used grep to find all of the references to that variable, which is how I found the persistent settings file.

Totally makes sense. For such cross-cutting concerns, in a non-trivial application as organice, I'm doing the same(;

> I am not sure why you can not push to my PR. I didn't change any settings so I will have to go check my defaults.

I think the difference is that your PRs are from your `master` branch to the upstream `master` branch. Usually, you'd create a branch on your repository from which to create the PR. I think that should already do the trick.

If not, I'll be happy to look over the [Github documentation on Pull Requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) with you.


> I will also see if I can figure out how to remove the .swp file (my git skills are limited)

Don't worry about this one committed file, then. I can remove it later. To make it easier on yourself on the future, I would suggest telling git to automatically ignore certain files. A VIM swap file usually never needs to be committed. There's different ways of doing so, but I suggest saving it to your personal git profile. Then, it cannot happen on any project on your computer. This is how to do it:

``` shell
# Tell git that you want to have a global excludes file:
git config --global core.excludesfile ~/.gitignore

# Add VIM swap files to said file
echo '*.swp' >> .gitignore
echo '*.swo' >> .gitignore
```

munen:
*hihi*, thanks for giving me push access to your repo :pray: That should do the trick, as well :+1:

dmorlitz:
Something must have gone wrong when I made the first commit - because I
actually did make a branch for this feature and then my development machine
pushed back to master.  Oh well.

I am going to try and update the PR.  I have merged in your changes and it
appears to work well now.  Thank you for the guidance - I doubt I ever
would have found the react_redux.

I have done limited testing on my devices and when the text box is taller
than the screen, you can still easily scroll down the screen to find the
icons at the bottom.  The keyboard does not appear to get in the way.  I
have tried this on both an iPhone 13 Pro XL and a Google Pixel 3XL.

I still agree that having a logical switch for a mobile device would
probably be the better solution.  People might not notice that the screen
is scrollable.

dmorlitz:
This PR is ready for a re-review - and still needs the mobile browser check added.

munen:
> I don't seem to be able to rerun the CircleCI checks - is there any way for you to restart those?

You mean on your fork? You'd have to set up [CircleCI](https://circleci.com/) for that. CircleCI doesn't automatically configure itself for forks. To get even faster feedback, you can also run the checks locally on your machine: `yarn eslint` and `yarn test`.

munen:
The setting works. The first obligatory test I made was to add a rather large number. Assuming that some user might not be happy with the small textarea, she will enter something like 77. This yields:

image

It would be good to ensure that the task modal can always be rendered, independent of what actual number was entered. That should be possible with some CSS.

munen:
A first option I tried was to restrict the max size of `drawer-inner-container`. However that won't work, because the textarea inside will still be too big and overflow. The necessary change will be a bit more complex. It probably has to take into account how big `drawer-inner-container` can become in total and then restrict the size of the textarea appropriately. Maybe there's an easier way that I don't see right now. 

I'll have to sign off for today. Thank you for your good work and good luck on continuing with this PR :pray: 

Have a nice evening and a relaxing start into the weekend :wave:

dmorlitz:
This code still has the isMobile check in to set the height of a textarea to 8 when the run on a mobile device.  Given the device height restrictions, I am not sure if that is still necessary but I left it in anyway.

dmorlitz:
The build failed on prettier-eslint without showing any error messages, other than

success formatting 9 files with prettier-eslint
101 files were unchanged
error Command failed with exit code 1.

How do I figure out what is failing with the prettier-eslint step, so I can finish submitting this PR?

munen:
What's interesting is that `yarn.lock` has pretty big changes while the PR does not include changes in `package.json`.

munen:
This is really weird. I ran prettier with `yarn prettier-eslint --write` in https://github.com/200ok-ch/organice/pull/873/commits/755aee02d96b699e2b6b0054bfa71ebca9d9c7ed - meaning it was happy on my machine, but CI still isn't. Since this branch has many changes in `yarn.lock`, I also did:

```
rm -r node_modules
yarn install
yarn prettier-eslint --write
```

And now I see changes in 7 more files.

I don't know how/why `yarn.lock` includes these changes and propose to remove them from the PR. There are some changes to core library versions and some core tools (prettier) behave differently on `master`. Even if they are reasonable upgrades, I propose to do them on a separate branch.

munen:
I have reviewed and tested this PR. Apart from the change request above, it's perfect and ready to go. Good job and nice addition! :pray: :bow:

dmorlitz:
When I created the build environment on my system and started the development server with "yarn build" there were several messages that said I needed to run "yarn upgrade".  I was not sure if that was required to get the development server running, so I did run the upgrade.  Apparently that was a more broadly reaching action than I thought.

There is no reason that I am aware of that this code requires any module upgrades.  I will try reverting the yarn.lock with the instructions above.

For future PRs, I will rebuild my development environment and simply ignore those upgrade warnings.

munen:
I suppose, you have seen this:

```
munen@lambda:~/src/200ok/organice% yarn build
yarn run v1.22.4
$ ./bin/compile_search_parser.sh && react-scripts build
Creating an optimized production build...
Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`
Compiled successfully.
```

Yes, it's true that the `caniuse-lite browserslist` package is not the newest. We can totally upgrade it. Running `yarn upgrade` upgrades all packages it can find, however. Generally speaking, that's potentially dangerous to do in a slightly bigger than very small JS codebase(; We do run upgrades and have a semi-automated process for that using dependabot. Here are some recent updates that came in through this:

image

Using the semi-automated PR based process, it's usually easy to check out a single upgraded package, make sure all automated and manuell tests pass and only then merge it in.

munen:
In c93997afed0ffe6e1ce2d84de9213e320fdbbd1e, you've updated the 7 additional files I mentioned in https://github.com/200ok-ch/organice/pull/873#issuecomment-1173561755.

Since we've found out that upgrading dependencies is not intended for this PR, if it's alright with you, I'll clean up this branch and merge it in.

To get your setup in a working and synchronized state, I suggest you:

```
git checkout master # upstream, not your clone
rm -r node_modules
yarn install
```

dmorlitz:
Please feel free to clean this up.  I am still learning my way through Git
and I will get better on future requests, like properly using branches, the
right .gitignore settings, etc.......

munen:
:+1: :pray:

munen:
Closing in favor of https://github.com/200ok-ch/organice/pull/875

Issue #868, opened 2022-06-30
Sign in to self-hosted GitLab instance


MaxSchlueter:
**Describe the bug**
In organice I'm trying to sign in to my own GitLab instance. I get the following error message:
```
Project does not appear to be a valid gitlab.com URL
```
Naturally, it's not a gitlab.com URL. This should in principle be possible though, right?

munen:
Hi @MaxSchlueter 

Thank you for using organice :pray: 

Unfortunately, what you want is not possible to my knowledge. This has been researched a couple of times, at least. There should be some information in old issues, discussions or the community chat somewhere.

If this has changed, you're very welcome to provide a PR with the necessary changes. Until then, this is not a bug, but a feature request pending on changes by a third party.

Thank you for your effort :+1:

MaxSchlueter:
Hi @munen 

sure, I can have a look into it. Do you have any pointers to the research that has happened around this issue? I couldn't find any information yet looking through old issues.

munen:
Great!

Since you’re trying to implement it, personally I would ignore any prior comments and look at the Gitlab API docs directly. If it works, it works. And then it might even be a change of only a few bytes to make it work, too^^

poVoq:
Selfhosted Gitea and Codeberg.org support would be nice as well. They also support acting as a Oauth2 provider so the setup should be similar to Gitlab.com

munen:
@poVoq Are you interested to work on that? NB: Just signing in won't be enough. You'll have to create a sync back-end compatible with each of those new storage providers.

poVoq:
Wouldn't that be just git over http?

munen:
If you can implement that, it will be.

zaz:
When I tried to fork this project I initially, and ironically, got a 500 instead of a 200 OK.

It does look like GitLab instances use the same API as [gitlab.com](https://gitlab.com).  So simply replacing "gitlab.com" everywhere with the URL of the self-managed instance should work.  I attempted a fix at [change at zaz/organice » gitlab-instances](https://github.com/200ok-ch/organice/compare/master...zaz:organice:gitlab-instances), but have not tested the code yet.

zaz:
Thank you!

[...]

PR #874, opened 2022-07-02
Extend GitLab sync support to self-hosted instance

MaxSchlueter:
This is done by getting rid of hard-coded "gitlab.com" URLs, instead
creating a new persistent field that stores the hostname of the GitLab
instance and constructing the URLs based on this field.

Adresses #868

### DIFF
--- a/src/sync_backend_clients/gitlab_sync_backend_client.test.js
+++ b/src/sync_backend_clients/gitlab_sync_backend_client.test.js
@@ -7,9 +7,11 @@ import {
 
 test('Parses GitLab project from URL', () => {
   [
-    ['https://gitlab.com/user/foo', 'user%2Ffoo'],
-    ['https://gitlab.com/group/subgroup/project', 'group%2Fsubgroup%2Fproject'],
-    ['gitlab.com/foo/bar', 'foo%2Fbar'],
+    ['https://gitlab.com/user/foo', ['gitlab.com', 'user%2Ffoo']],
+    ['https://gitlab.com/group/subgroup/project', ['gitlab.com', 'group%2Fsubgroup%2Fproject']],
+    ['https://gitlab.example.com/user/foo', ['gitlab.example.com', 'user%2Ffoo']],
+    ['gitlab.com/foo/bar', ['gitlab.com', 'foo%2Fbar']],
+    ['gitlab.example.com/foo/bar', ['gitlab.example.com', 'foo%2Fbar']],
     ['gitlab.com/user-but-no-project', undefined],
     ['', undefined],
   ].forEach(([input, expected]) => {
--- a/src/actions/sync_backend.js
+++ b/src/actions/sync_backend.js
@@ -28,6 +28,7 @@ export const signOut = () => (dispatch, getState) => {
       break;
     case 'GitLab':
       persistField('gitLabProject', null);
+      persistField('gitLabHost', null);
       createGitlabOAuth().reset();
       break;
     default:
--- a/src/components/SyncServiceSignIn/index.js
+++ b/src/components/SyncServiceSignIn/index.js
@@ -113,9 +113,10 @@ function GitLab() {
   const defaultProject = 'https://gitlab.com/your/project';
   const [project, setProject] = useState(defaultProject);
   const handleSubmit = (evt) => {
-    const projectId = gitLabProjectIdFromURL(project);
+    const [hostname, projectId] = gitLabProjectIdFromURL(project);
     if (projectId) {
       persistField('authenticatedSyncService', 'GitLab');
+      persistField('gitLabHost', hostname);
       persistField('gitLabProject', projectId);
       createGitlabOAuth().fetchAuthorizationCode();
     } else {
--- a/src/sync_backend_clients/gitlab_sync_backend_client.js
+++ b/src/sync_backend_clients/gitlab_sync_backend_client.js
@@ -12,8 +12,8 @@ export const createGitlabOAuth = () => {
   let expiryPromise;
   let invalidGrantPromise;
   return new OAuth2AuthCodePKCE({
-    authorizationUrl: 'https://gitlab.com/oauth/authorize',
-    tokenUrl: 'https://gitlab.com/oauth/token',
+    authorizationUrl: `https://${getPersistedField('gitLabHost')}/oauth/authorize`,
+    tokenUrl: `https://${getPersistedField('gitLabHost')}/oauth/token`,
     clientId: process.env.REACT_APP_GITLAB_CLIENT_ID,
     redirectUrl: window.location.origin,
     scopes: ['api'],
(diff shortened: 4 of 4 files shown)


chasecaleb:
Hi, do you have a plan for dealing with the OAuth secret? The reason this only works for gitlab.com is because you need a separate OAuth integration for each GitLab instance, so unfortunately you can't just change the URL. The OAuth secret is configured at deploy time, so this won't allow using a different GitLab instance with organice.200ok.ch unfortunately. See https://github.com/200ok-ch/organice/blob/master/.env.sample

munen:
@chasecaleb Thank you for chiming in🙏

@MaxSchlueter Thank you for giving it a shot and providing a PR🙏To the best of my understanding, @chasecaleb is right. I don't see how to accommodate that requirement on the official instance. 

I suppose supporting self-hosted Gitlab instances implies a self-hosted organice instance. If so, we could add a few lines of documentation for it. I suppose that's not so bad. Anyone willing and capable to self-host a non-trivial application as Gitlab is likely capable and willing of self-hosting a single-page application or at least the official Docker image.

branch14:
Here is an idea that I briefly discussed with @munen, and we think it's feasible.

The OAuth secret mentioned by @chasecaleb is currently injected into the build. (In the code it looks like it is read from an env var, but it actually is read from the `.env` file during build time.) This happens here:

https://github.com/200ok-ch/organice/blob/c948643116b910c82aea37a3ecc00fbb922a3016/src/sync_backend_clients/gitlab_sync_backend_client.js#L17

My suggestion is the following: The sign-in form for Gitlab gets a second input field which takes the OAuth secret (aka Gitlab app token, aka client id). The second input field is disabled as long as the first input field holds a repo url for gitlab.com. If the user enters a repo url for a Gitlab instance other than gitlab.com, the second field is enabled and a token is required.

@chasecaleb What do you think?

@MaxSchlueter Would you like to give this a shot?

MaxSchlueter:
> @MaxSchlueter Would you like to give this a shot?

Sure, thanks for chiming in. I haven't worked with OAuth secrets before, but wouldn't the client secret (`REACT_APP_GITLAB_SECRET`) as a third input field be needed as well?

munen:
@MaxSchlueter Yes, both `REACT_APP_GITLAB_CLIENT_ID` and `REACT_APP_GITLAB_SECRET` will have to be overridden in the login form.

src/sync_backend_clients/gitlab_sync_backend_client.test.js
```diff
@@ -7,9 +7,11 @@ import {
...
   [
-    ['https://gitlab.com/user/foo', 'user%2Ffoo'],
-    ['https://gitlab.com/group/subgroup/project', 'group%2Fsubgroup%2Fproject'],
-    ['gitlab.com/foo/bar', 'foo%2Fbar'],
+    ['https://gitlab.com/user/foo', ['gitlab.com', 'user%2Ffoo']],
+    ['https://gitlab.com/group/subgroup/project', ['gitlab.com', 'group%2Fsubgroup%2Fproject']],
+    ['https://gitlab.example.com/user/foo', ['gitlab.example.com', 'user%2Ffoo']],
+    ['gitlab.com/foo/bar', ['gitlab.com', 'foo%2Fbar']],
+    ['gitlab.example.com/foo/bar', ['gitlab.example.com', 'foo%2Fbar']],
     ['gitlab.com/user-but-no-project', undefined],
```
schoettl, line 15: Not your code @MaxSchlueter but I just saw this: Isn't it uncommon to use `undefined` deliberately @chasecaleb @munen ? I'm not a professional JS dev but I thought `undefined` is e.g. for a forgotten return statement or an undefined variable. Wouldn't `null` be the better option here? I mean, the parse function should return null if not successful?

kephale:
I'm super excited about this PR!

munen:
@kephale Would you like to take it over? There's one small step missing that's discussed starting here: https://github.com/200ok-ch/organice/pull/874#issuecomment-1172902067

kephale:
I can check, but it was really easy to just pull these changes into the latest `main` then (at least) just run off my own instance with my instances `.env` info. 

It would certainly be *better* to be able to resolve the secret stuff at runtime instead of build time, but even without that, this solved my use case.

munen:
> It would certainly be better to be able to resolve the secret stuff at runtime instead of build time, but even without that, this solved my use case.

It works for users self-hosting Gitlab _and_ self-hosting organice, now. But not everyone is self-hosting organice. So adding two more fields to the login will be the right way to go forward.

kephale:
Some effort in this direction, but not working yet https://github.com/kephale/organice/tree/self-host-wrapup

src/sync_backend_clients/gitlab_sync_backend_client.test.js
chasecaleb: Hi @schoettl, no idea how I missed your ping all those months ago :facepalm:

>  Imo the author of the original function wanted to be explicit about it, since it's an error case. 

Yes, this is why I didn't just do `return` -- plus I generally work in languages that have a void type so they don't allow you to mix a plain `return` and `return <something>` in the same function. 

But as for specifically picking `undefined` instead of `null`... meh :shrug:. Around the time I wrote this code I was working in a TypeScript codebase that banned `null` and stuck with only `undefined` for unimportant reasons.

In my opinion deciding whether to use `null` vs `undefined` doesn't really matter beyond perhaps picking one for style/consistency reasons. Instead, it's better to write the calling code to handle both equivalently... for example, `if (foo == null) { ... }` is true if `foo` is either `null` or `undefined`. This goes doubly so for JavaScript, as opposed to Typescript where you at least _might_ know whether a value could be null vs undefined vs neither.

Here's an interesting read: https://basarat.gitbook.io/typescript/recap/null-undefined . You'll notice that Basarat agrees with you about using null, @schoettl... but at the end he also points out the TypeScript team only uses undefined and [Douglas Crockford](https://en.wikipedia.org/wiki/Douglas_Crockford) also says to avoid null.

The _real_ best practice is to use a language with neither `null` nor `undefined` :grin:  I'll see myself out now...

munen: Hi @chasecaleb 

Thanks for chiming in and giving some detailed feedback!

I like your point about writing code that handle both cases equivalently. That's very common in dynamic languages.

For sake of completeness, and future readers, I'll say that the linked [interesting read](https://basarat.gitbook.io/typescript/recap/null-undefined) doesn't completely apply to JS, as it is meant for a TS audience. For example "Don't use undefined as a means of denoting validity". For statically typed language, it's the right call. And it's reasonable to explicitly point it out when working from a dynamically typed language into a statically typed language (as is the case with TS). In a purely dynamically typed language (like JS, Clojure, Ruby, etc), it's quite common and good practice. If anyone asks why: They have different tradeoffs. Check out the example code of the article above. It's arguable that the first example, a simple one-liner is very easy to read and reason about. The second function has  a body of 7 lines, that's quite a big hit to readability. Having said so, it's the style of statically typed language, _because_ it yields a type that's easy to reason about. That's a pro, too. Just a different one. You might be inclined to like one more or the latter. Everything is ok :pray:

schoettl: Alright, so now fixed conventions about this in organice. I personally used `x === null` for my null checks because `==` generally feels so sloppy in JS and I always fear that beginners pick it up and use it for other types like `x == ''`…

PR #876, opened 2022-07-05
WIP: Links to non-org files go to GitLab

motevets:
This work in progress / proof-of-concept shows a potential solution for
links to files that are not org files or folders (#325). It does this by:

- saving the URL to the GitLab project at login time
- parsing links to files that contain a "." but are not org files
- making a link to the file with the template
  `${gitLablProjectUrl}/-/blob/${defaultBranch}/${path}`

The commit is not production ready and has not added or updated any of
the tests, but it meant to solicit early feedback on this approach.
Notably it _only_ works for GitLab sync backends (though others will
revert to the default behaviour.)

---

A working example can be seen at: https://org.kindrobot.ca

### DIFF
--- a/src/components/OrgFile/components/AttributedString/index.js
+++ b/src/components/OrgFile/components/AttributedString/index.js
@@ -16,7 +16,7 @@ import * as orgActions from '../../../../actions/org';
 
 import classNames from 'classnames';
 
-const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
+const AttributedString = ({ org, parts, subPartDataAndHandlers, getExternalFileUrl }) => {
   let className;
 
   let location = useLocation();
@@ -38,14 +38,24 @@ const AttributedString = ({ org, parts, subPartDataAndHandlers }) => {
         if (!target.includes('/../')) {
           // Normalisation succeeded, so we can safely return a <Link>
           if (!uri.match(orgFileExtensions)) {
-            // Optimistically assume that the link is pointing to a
-            // directory.
-            target = target.replace(/^\/file\//, '/files/');
-            return (
-              <Link key={id} to={target}>
-                {title}
-              </Link>
-            );
+            if(uri.match('\\.') && typeof getExternalFileUrl === 'function') {
+              // assume that this is a file that we can link to from the backend
+              target = target.replace(/^\/file\//, '/');
+              return (
+                <ExternalLink key={id} href={getExternalFileUrl(target)}>
+                  {title}
+                </ExternalLink>
+              );
+            } else {
+              // Optimistically assume that the link is pointing to a
+              // directory.
+              target = target.replace(/^\/file\//, '/files/');
+              return (
+                <Link key={id} to={target}>
+                  {title}
+                </Link>
+              );
+            }
           }
         }
       }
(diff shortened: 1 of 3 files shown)


branch14:
Thank you @motevets for your pull request. Your contribution is highly appreciated. Your motivation and approach are comprehensible. I don't mean to be blunt, but I'll be frank since I don't want to raise false hopes that we will be going into that direction. Let me use your contribution as a good basis for a discussion on how to move forward.

I concur, it is unsatisfactory that linked files cannot be opened in organice. But organice, being a PWA, runs in a browser. It should be feasible to display most files without leaving its context.

Leaving the context of the PWA raises usability issues. What happens on the 6 different platforms[1] when opening an ExternalLink? Leaving the context of the PWA will also mean losing Authorization.

Before enhancing organice we try to answer the question WWED (What would Emacs do?) Whenever possible organice tries to replicate the behavior of Emacs in order not to violate the principle of least surprise. To answer the question: As Emacs will open local files by itself, this is what organice should do.

The current implementation, which identifies org files by there extensions (`.org` or `.org_archive`) and falls back to assuming everything else is a directory is a hack -- a hack with quite some history[2]. Tackling this topic, it might be helpful to be familiar with the history, as some valid points have already been discussed by bright minds.

Adding more conditions based on looking at the path or filename does not feel like a solid solution. IMO the only way to handle files (and directories) properly is to fetch it and deal with what is received. This might mean we have to convert raw image data into displayable blobs and maybe enhance the sync backend clients.

Organice's sync backend clients are implemented as a strategy pattern. The set of functions exposed by the modules is an interface that all clients must adhere to. Adding to this interface raises some concerns as divergence is a maintainability issue.

Again, I didn't mean to be blunt. I want to thank you for the time you put into this and I love to see the joined effort in improving organice's features and usability.


[1] The 6 platforms organice runs on

- in a desktop browser
- in a mobile browser
- as PWA installed on android
- as PWA installed on ios
- webview in android app (soon to come)
- webview in ios app (soon to come)

[2] History of organice's file browser

- https://github.com/200ok-ch/organice/pull/307
- https://github.com/200ok-ch/organice/issues/311

motevets:
Thank you for your thoughtful response @branch14 . I had a feeling this would not be a generally acceptable solution. I implemented it in my fork, and it's suiting my needs, so in the spirit of open source, I thought I should share back my modifications for consideration, no matter how remote the chances of them getting merged in. I'll go ahead and close this PR.

I'd be interested in contributing a more cohesive and comprehensive PR down the road when I have the time, unless its on someone at 200ok's todo list.

munen:
@motevets It's something that we have thought about in the back of heads since starting organice. So, if somebody beats us to the chase, we will be thankful for the contribution.

When you have time, and we haven't implemented this yet, we are happy to support you with guidance, code reviews and testing anytime :pray:

motevets:
I was thinking about this some more and was wondering if this would be an acceptable solution based on what you said.

>  This might mean we have to convert raw image data into displayable blobs and maybe enhance the sync backend clients.

Since the sync backends already have the ability to sync data, would it be acceptalbe to make blobs from the raw data and trigger a download? I think that would work for all the different PWA use-cases, and it might be an itterative step towards displaying an image inside of organice (for say, inline images). (Side note: inline images aren't a big motivation for me. I'm more likely to have links to PDFs or other files I'll need while completing the task.

branch14:
@motevets Yes, triggering a download might be a good first step. It might even be enough for many file types.

I guess there will be a lot of edge cases that need to be dealt with, e.g. what if someone links to large files. Ideally the sync backends provide ways to query metadata, like mimetype and size, and organice can take suitable actions based on that.

But even for larger files a download might be the expected outcome. (Some anecdotal evidence: Decades ago I worked at an organization that had no limit on the size of email attachments and staff would regularly send ISO images and then complain to IT that their inboxes are full and everything is slow. Good times.)

 If we had metadata we could issue a warning before starting the download.

Issue #864, opened 2022-06-28
Change log renders above previously opened view component


tarnung:
Recently the changelog acts strangely for me.
I had cases where the changelog showed up double. The view started out scrolled somewhere into the middle of the changelog and as i scrolled down to the start of the history it repeated (again showing "general" and the whole change log beneath it).
Now i discovered an even stranger bug. I opened the change log from the settings menu and the change log again started scrolled down quite a bit. As i scrolled past the end i scrolled into the settings menu which was still there below the change log.

The bug is reproducible. It happens every time i open the changelog from the settings menu
I observed this on my android phone and in the desktop browser.

The starting scrolled down part is probably due to the browser keeping its scroll position since it does not load a new view but prepends the changelog to whatever view is shown already.

munen:
Confirmed. That's a pretty ugly bug, too.

I propose fixing it by getting rid of the concept 'modal' for changelogs in the sense that it would be nice to have [proper routes wherever we can](https://github.com/200ok-ch/organice/issues/26).

munen:
I have created a PR. Sample and Changelog now only work with their own routes `/sample` and `/changelog`. There's no weird 'modal logic' that puts stuff on top of stuff.

If you would like to review it, that would be nice. If you like it, please go ahead and merge it.

PR #878, opened 2022-07-07
fix: Start from the top

munen:
Reported in https://github.com/200ok-ch/organice/issues/864

The regression was visible, for example, when going first to settings and then to the changelog. The changelog would then show in a 'scrolled down' position.


branch14:
:+1:

Issue #880, opened 2022-07-07
Documentation has escaped HTML


schoettl:
Seems to be a bug in the documentation make/deploy process...

![grafik](https://user-images.githubusercontent.com/1301843/177836179-88ceff18-c1ed-4dc5-b206-72b09199d6ce.png)

https://organice.200ok.ch/documentation.html

munen:
Good catch!

@branch14 recently introduced nice labels for the Readme. Doesn't look like they are playing well with the docs -_-

![image](https://user-images.githubusercontent.com/505721/177836462-3db0ed9f-e4e0-48fd-a967-d1245808ef37.png)

munen:
Makes actual sense, Phil replaced the 'ugly' inline `#+html:` blocks with `#+BEGIN_HTML`, but those are different When Emacs exports the latter, that's actually a code block!

munen:
I thought that setting `:results value raw` to the source block would work, but on a first try it does not. I'll look into it another time.

Relevant exporting docs: 

https://orgmode.org/manual/Results-of-Evaluation.html

schoettl:
Shouldn't it be `#+BEGIN_EXPORT html`? https://orgmode.org/manual/Quoting-HTML-tags.html

munen:
Yes, that would be a totally valid option. But then the wonderful legacy Org mode parser of Github won't show it in the Readme(;

munen:
If only there would be a good parser and Github would be FOSS, then we could make a PR to remedy that. A big and an even bigger 'if'(;

schoettl:
> I thought that setting `:results value raw` to the source block would work, but on a first try it does not. I'll look into it another time.
> 
> Relevant exporting docs:
> 
> https://orgmode.org/manual/Results-of-Evaluation.html

Do you even have an HTML babel interpreter configured (`org-babel-execute`)? What does that do? The most useful thing is lingting IMO :D

```
#+begin_src html :results value raw
  <p>
#+end_src
```

munen:
> Do you even have an HTML babel interpreter configured (org-babel-execute)?

I’m not sure I understand the implication. Of course babel understands HTML. 

> The most useful thing is linting IMO :D

Again, I don’t understand. There’s great tooling around HTML in Emacs and Org. For example this is a popular and powerful mode: https://web-mode.org/

Having said all that. For the Readme/docs, that’s all completely irrelevant. We just want to inject HTML into the exported document.

munen:
[...]

PR #881, opened 2022-07-08
fix: Rendering of HTML in documentation

munen:
Closes: https://github.com/200ok-ch/organice/issues/880

### DIFF
--- a/README.org
+++ b/README.org
@@ -14,19 +14,14 @@
 
 # REPO_PLACEHOLDER
 
-#+BEGIN_HTML
-<p>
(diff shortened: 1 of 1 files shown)


munen:
@schoettl or @branch14 If you agree that this is the best solution (or at least a temporary fix with a better solution down the line), please go ahead and merge.

branch14:
Sorry for making a mess. :pray:

munen:
@branch14 No worries. I didn't test all implications when we merged, too^^

PR #892, opened 2022-08-02
Organice emacs automation

dmorlitz:
Added instructions to automatically run Emacs commands against Organice files - in situations where automations are desired.

For Emacs users this might already be common knowledge - but since I am not an Emacs user this took a bit of time to figure out.  There can be some very handy automation possibilities knowing how to use Emacs Org mode in a batch mode that I thought others would be interested in knowing what was available.

If this is commonly known, or you do not believe it is relevant to your users, please feel free to reject the pull request.

### DIFF
--- a/contrib/organice-emacs-automation/emacs-automation.org
+++ b/contrib/organice-emacs-automation/emacs-automation.org
@@ -1,0 +1,34 @@
+#+TITLE: Automating Emacs operations for Organice
+
+* Purpose
+
+You may have a need to have some automation to help you manage your Org files.  One example, which this document will explain,
+is to automatically sort your org file based on DEADLINE's.  While these tasks might be easy to do directly inside Emacs, if you
+are not an Emacs user and rely solely on Organice, you may want to enable some automation
+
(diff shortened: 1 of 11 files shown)


munen:
Hi @dmorlitz 

Great that you have found out about Emacs batch mode. It’s opening up vast opportunities for automation. At 200ok, we use it extensively. 

There’s one more thing that’s even faster than batch mode: Using emacsclient. We recently released a blog post about it: https://200ok.ch/posts/2022-07-04_org_export_using_emacsclient.html

Maybe it’s interesting to you and you can find some use out of it. 

All the best and kind regards from a small break during the [7-day pilgrim retreat](https://zen-temple.net/pilgrim-retreat/) .

dmorlitz:
I am amazed at how much power there is here - and how it has taken me so
long to find it.

I don't currently run Emacs as a daemon - but I will definitely look into
it.  I don't have anywhere near that much automation so I am not sure if
sorting my list of birthday, tasks and family event reminders would warrant
a full daemon.......but I will think about it.

Enjoy the rest of your retreat.

PR #902, opened 2022-11-02
fix: Use correct HTML anchor tag inside public.html <noscript>

josephmturner:
Thanks for this project!!

Joseph Turner


munen:
Hi @josephmturner 

Thank you for your kind words and good job on the fix! We must have been overeager when replacing the anchors at one point in time^^

I've added the fix to the changelog and gave you attribution.

Enjoy using organice and happy hacking^^

PR #911, opened 2022-12-03
Fix typos

lechten:


munen:
Wow, I'm quite impressed with how many actual typos we had in the docs - thank you for doing the grunt work and fixing them :pray: :bow: :100:

Issue #913, opened 2022-12-06
Undesirable auto-capilalization in search text input (Android Chrome)


neildavidforrest:
**Describe the bug**
When using the search or task list text inputs the Android keyboard automatically applies the shift key, meaning that the first character is capatalized.

99% of the time this is not what I want because:
- Having mixed case triggers a case sensitive search. This is quite confusing when I did not intentionally use capitalization, but it was added anyway. For example, I type "and", and it actually inputs and searches for "And", which yields no results.
- Many of the keywords in the search grammar are lower case, e.g., `deadline:` `desc:`.  

As a consequence the first step of any search I do is to disable the shift key, which is rather annoying.

![image](https://user-images.githubusercontent.com/106537273/205782730-5b5df1d6-96cb-4ff7-9823-f622af442bea.png)

**To Reproduce**
1. Open the search modal

**Expected behavior**
Default should be lower case. IMHO.

**Screenshots**

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
 - Device: Fairphone 4
 - OS: Android 11
 - Browser: Chrome 107.0.5304.141 

**Additional context**
N/A

munen:
Closed by #914.

PR #914, opened 2022-12-06
Disable auto-capitalization for search

neildavidforrest:
This disables auto-capitalization for the search fields on Android Chrome as described here: https://github.com/200ok-ch/organice/issues/913

This is tested only on Chrome for Android only. I also checked Firefox for Android, this works fine (no auto-capitalization) with or without this change.

I'm unable to test with iOS, so that should be checked carefully for regressions.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -6,6 +6,20 @@
 All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
+
+* [2023-01-04 Wed]
+
+** Changed
+
+- Disable auto-capitalization for search
+  - Before:
+    - When using the search or task list text inputs the Android/iOS keyboards automatically apply the shift key, meaning that the first character is capitalized.
+    - 99% of the time this is not what we want because:
+      - Having mixed case triggers a case sensitive search. This is quite confusing when I did not intentionally use capitalization, but it was added anyway. For example, I type "and", and it actually inputs and searches for "And", which yields no results.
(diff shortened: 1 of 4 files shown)


schoettl:
I'm not sure about autoComplete off. Have you tested if the suggestions from the dynamic datalist still work? E.g. if I type `:` or `T`, the completion list opens with matching tags or `TODO` keywords. This at least works on master in Firefox.

neildavidforrest:
I'm also not sure about it... but I can say that in my tests `autoCapitalize="none"` is *not* sufficient to disable the auto shift key by itself.

I can confirm that the auto complete still works correctly on Chrome mobile, and Firefox desktop after this change.

image

On Firefox mobile the auto complete does not work before or after, so no change there...

I didn't find any authoritative source about this, but there is some discussion here: https://stackoverflow.com/questions/13284323/how-can-i-stop-auto-capitalization-in-a-html-text-input-field-on-an-android-sams

schoettl:
Oh good, if it still works, I'm fine :) Thanks for testing and this PR!

munen:
Hi @neildavidforrest 

Thank you for this PR! My apologies the the late answer :pray: :bow: 

Thank you for the thorough testing on Android. I have tested on FF and Chromium on Linux, there are no regressions. Furthermore, iOS had the same undesirable auto-capitalization as you described it on Android. Your PR fixes this, as well :+1: 

I have fixed the test and updated the changelog. This is ready to go.

Thank you for yet another great productivity contribution :pray: :100: :rocket:

PR #925, opened 2023-01-04
Deadline overlay UI improvements

dmorlitz:
Made improvements to the deadline overlay by making it non-breaking to avoid line wraps and increased the right margin to avoid falling off of the screen.

Still may need some work to guarantee all cases - but it solved all of the browsers and devices I tested.


munen:
Hi @dmorlitz 

Wow, that was a quick fix(;

The change is a solid improvement! I found an cases when having both overlays (clock time and deadline) enabled where one can be moved off the screen:

image

But it's certainly an edge case. No reason not to merge.

Thank you for the good work! :pray:

PR #928, opened 2023-01-06
376 - fixed text discards while check/uncheck box when text has inline-markup

saikiransathpadi:
This is related to the issue https://github.com/200ok-ch/organice/issues/376. When we have a check box and whenever we check or uncheck it if we have an inline markup in the description it is getting disappeared. 

There is a switch case for each type to convert to raw text and here the "inline-markup" case is missing. I have written a separate function to handle this text and added the case there.


munen:
@saikiransathpadi Another nice bug fix! I have tested it and updated your changes to the changelog.

Thank you for your continued effort :pray: :bow:

PR #938, opened 2023-01-28
Don't cancel swipe due to vertical movements

neildavidforrest:
Contrary to most other examples of swiping interactions, swiping a headline in Organice only works if you keep your finger/thumb/mouse inside the bounds of the item being dragged. This makes it rather difficult to swipe reliably, especially on larger devices (and/or smaller hands) since the hand naturally wants to move in an arc, while the headline bounds are linear.

This change is intended to align the swipe interaction with the behavior shared by the native interactions of both Android and IOS: that after initiating a swipe, no amount of vertical movement should cancel it.

Regarding the implementation: It works reliably in my tests, however I'm not previously familiar with React or web development in general, so please feel free to suggest better alternatives.

**Before**:
video

**After**:
video

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -7,6 +7,14 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2023-01-28 Sat]
+
+** Fixed
+
+- On a header, don't cancel swipe due to vertical movements
+  - PR: https://github.com/200ok-ch/organice/pull/938
+  - Thank you [[https://github.com/neildavidforrest][neildavidforrest]] for the PR🙏
+
 * [2023-01-09 Mon]
 
 ** Fixed
--- a/src/components/OrgFile/components/Header/index.js
+++ b/src/components/OrgFile/components/Header/index.js
@@ -32,12 +32,7 @@ class Header extends PureComponent {
     _.bindAll(this, [
       'handleRef',
       'handleMouseDown',
-      'handleMouseMove',
-      'handleMouseUp',
-      'handleMouseOut',
-      'handleTouchMove',
       'handleTouchStart',
-      'handleTouchEnd',
       'handleTouchCancel',
       'handleHeaderClick',
       'handleShowTitleModal',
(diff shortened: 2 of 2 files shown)


munen:
Hi @neildavidforrest 

You said you have no experience with React or web development - yet, I think you came up with a really nice solution. I have tested it on my iPhone 13 pro and it works a treat!

I have added the fix to the changelog and added attribution.

To give a bit feedback on the implementation: Generally speaking, it is considered non-idiomatic in React to work with `window.addEventListener`. React is component based, and every component _should_ (in optimal theory) be self-sustained and composable with other components. Hence, any component that works with any kind of global state messes with this idiom. You did catch the bug correctly. The `<Header>` component only works on the header. When leaving this component, the code stops working. Hence, your solution to increase the scope for tracking the gesture is necessary. You did it by tapping into global state - which is totally idiomatic in many cases. Let's say you write a HTML/JS app from scratch or use a library like jQuery, then this is the idiomatic way to go. In React, it will break the boundary of the component and composability. Therefore, the idiomatic approach would have been to create a wrapping component, let's say `<Swipe>` that wraps all components that need to track the swiping gesture.

Having said all that, I don't think that tapping into global state is bad in this case and that it will come back to bite us. Instead, it's a pretty neat solution packing all necessary logic into the one component that really needs it. That's the other side of the coin.

TL;DR: I like your implementation and am very happy to merge it! :pray: :bow:  Just wanted to give a tiny bit of feedback, because you said you're not familiar with web/React.

schoettl:
Thanks @neildavidforrest! This change is especially useful when using organice on desktop. It was really hard to delete a header with the mouse before!

(I wanted to test but I didn't got organice dev env to run yet under my new NixOS...)

munen:
Hi @schoettl 

Congrats to changing to NixOS(;

I'm on a similar trajectory - under Debian, I do a lot with Guix and I'm running Guix as a system on a different machine, as well. Good luck and have fun! :muscle:

neildavidforrest:
@munen Thanks for the detailed review. Not TL and DR (DID read)!

munen:
@neildavidforrest 🙏🏻

munen:
@neildavidforrest 🙏🏻

PR #947, opened 2023-04-02
Update Android/Firefox instructions in README

pakelley:
Looks like Firefox has changed how to do this. I'm on the most recent version of Firefox (on a Pixel, fwiw), but can't vouch that this is the process for most Firefox versions (or android devices) in use, so I'm happy to "append" this rather than editing the old version, if you'd prefer.

### DIFF
--- a/README.org
+++ b/README.org
@@ -106,8 +106,8 @@ instructions:
   - On Chrome, tap the "menu" button (three vertically stacked dots)
     and select "Add to homescreen".
 
-  - On Firefox, tap the home icon with the plus sign inside it which
-    is immediately to the right of the URL in the address bar.
+  - On Firefox, "menu" button (three vertically stacked dots)
+    and select "Install".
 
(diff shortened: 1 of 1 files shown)


munen:
Hi @pakelley

Thank you for taking the time to update the docs, it is much appreciated! Can you rewrite the docs to state that if the one UX does not work, then the other might? This might be prudent considering that there's usually a few users with older phones and operating systems, in any case.

Thank you, again, for your support! :pray:

pakelley:
Alright, both are included now. Lmk if the wording looks alright to you, I don't mind doing another pass if you want something changed. :)

munen:
Hi @pakelley 

Thank you for the quick update! I'm happy to merge it in.

Thank you, again, for the effort and helping us to keep the docs up to date🙏

pakelley:
no problem, thanks for your work on organice 🙂

PR #949, opened 2023-04-10
fix: Color of checkbox

munen:


neildavidforrest (approved):
Thanks for the quick fix! I confirm this restores the correct appearance of an unchecked box.

munen:
@neildavidforrest Thank you for testing :pray:  I'll merge this one in, then!

PR #958, opened 2023-06-19
fix: grammar on testimonial

jcpst:
Noticed this- I'm sure it's how I originally typed it, but would like to fix the grammar since my name is on it 😁


munen:
Hi @jcpst 

Thank you for the quick fix :pray: Happy to pull it in, of course!

I cannot remember rewriting your testimonial (as one should not do such things), but the mistake is not unlikely to have crept in through my non-native English.

Thank you, again, for the wonderful testimonial and enjoy using organice :pray:

PR #970, opened 2023-11-29
Add visual cue for overdue headers when displaying headers

dmorlitz:
When you enabled displaying header due dates on the right side of the screen a new visual cue (i.e. **11-29-2023**) has been added to clearly identify and separate overdue items from future items.


src/components/OrgFile/components/Header/index.js
```diff
@@ -361,6 +361,21 @@ ${header.get('rawDescription')}`;
...
+    const headerDate =
+      headerDeadlineMap !== undefined
+        ? new Date(
+            headerDeadlineMap.get('month') +
+              '-' +
+              headerDeadlineMap.get('day') +
+              '-' +
+              headerDeadlineMap.get('year')
+          )
+        : today;
```
schoettl, line 374: I don't understand how you construct this `new Date()`.
When I enter this weird order of arguments in node:
```js
> Date('01-05-2022')
'Wed Nov 29 2023 21:54:27 GMT+0100 (Central European Standard Time)'
```
I'd suggest to pass a date in ISO format.

dmorlitz:
I am open to any suggestions and changes that you might recommend.

The basic premise is that I take the date a header is due and subtract
today (the time is stripped off - so everything is assumed to be midnight
00:00.00)

If the result is negative, then the header is due in the past and the date
is surrounded by ** **

If the result is 0 or a positive number, then the header is due in the
future and the date is displayed normally.

I am a very amateur coder - so I ended up structuring this code after some
examples I found online.  That code took the date in the format 11-29-2023
(for today)

I am not bound to any of the code - and I will gladly take any
recommendations you want to make.

schoettl:
Maybe you can add a screenshot to your PR description above for people to see how it will look.

You can try your JavaScript in the JS console of your browser (just press F12). Try your `new Date(...)` and see [MDN for Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#syntax).

Also you would need to use git rebase to remove that invalid file, the nano screen dump.

dmorlitz:
Here is an example of some overdue and some future tasks - when the option to show deadlines on the header line are enabled.  The ones surrounded by ** are overdue.
Screenshot_20231129_165538

PR #987, opened 2024-09-16
Add basic flake.nix to get a dev shell on NixOS

schoettl:
It's still not trivial to get it run on NixOS...

### DIFF
--- a/README.org
+++ b/README.org
@@ -380,6 +380,10 @@ nvm install
 nvm use
 #+END_SRC
 
+If you have the [[https://nixos.org/][Nix package manager]] installed, just type:
+
+: nix develop
+
 ** Setup
 
 *** Installation of packages
--- a/changelog.org
+++ b/changelog.org
@@ -7,6 +7,14 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* [2024-09-19 Thu]
+
+** Added
+
+- Add a =flake.nix= file to provide a development shell for
+  [[https://nixos.org/][Nix]] users.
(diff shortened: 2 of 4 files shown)


schoettl:
@munen with this PR I can at least work on organice on NixOS ^^
… still not with nvm and the correct node version. nvm is not a thing on NixOS.

munen:
Hi @schoettl :wave: 

Congrats on getting a dev environment to work with NixOS :muscle: 

I understand that NVM uses a diametrically different approach than using any functional package manager where the latter of course is superior(;

I tried getting organice to run in Guix [some time back](https://github.com/200ok-ch/organice/pulls), but have not succeeded myself. Some commands worked, but `yarn` is not really a thing that works in Guix.

Anyway, what I'm saying is that I'm in favor of using functional package managers wherever possible. So, congrats on getting this done! Feel free to merge this PR whenever you see fit. If you want others to get on it, maybe add it to the changelog or Readme. Or not if you want to keep it low key for time being. All good for me :+1:

schoettl:
Hi @munen, yeah thanks :)
This PR didn't fully nixify organice but it's a good start. I still use yarn for almost all node dependencies. There is `yarn2nix` but to be honest, yarn is simpler to use than nix – and it works ;-)

munen:
@schoettl Congrats :100: :confetti_ball: 

> I still use yarn for almost all node dependencies. There is yarn2nix but to be honest, yarn is simpler to use than nix – and it works ;-)

On the plus side: You will be using the same tooling for library dependencies as other devs. Since yarn packages can do so much stuff (like compiling external libraries), I could see a future where we'd have to test multiple branches for library dependencies - or all devs switch to Nix long term(;

Issue #992, opened 2024-09-19
Upgrade node and dependencies?


schoettl:
On my system, the node version is 20.15.1 whereas organice uses 12.13.1.
There are many open dependabot PRs and `yarn audit` tells about 637 vulnerabilities.

How about trying an upgrade @munen?

Apart from `yarn test`, CI and basic manual integration test – what do we need to safely upgrade? I've never done this before.

How would staging or production cope with such changes? Should we give it a try, starting with node itself?

munen:
Hi @schoettl 👋

> On my system, the node version is 20.15.1 whereas organice uses 12.13.1. There are many open dependabot PRs and `yarn audit` tells about 637 vulnerabilities.

I'm definitively in favor of upgrading node👍


> Apart from `yarn test`, CI and basic manual integration test – what do we need to safely upgrade? I've never done this before.

> How would staging or production cope with such changes? 

Since organice is a SPA, hosting is not affected. As long as the regular dev, testing and build tools work. These tasks should run:

- `rm -rf node_modules; yarn install` 
- `yarn test`
- `yarn install`
- `yarn eslint`
- `yarn prettier-eslint --write`
- `yarn run build`

CI should run. We will have to update the Docker image which includes the node version.

Lastly, a manual integration test should be done. As long as the sample and logins still work, I think we should be in the clear.

> Should we give it a try, starting with node itself?

Sure, let's do it! I endeavoured into a first start: https://github.com/200ok-ch/organice/pull/993

munen:
@schoettl I made a spike (https://github.com/200ok-ch/organice/pull/993) to get us started on this task.  Are you interested in picking it up from here?

schoettl:
Nice! I'll pick it up.

PR #993, opened 2024-09-20
chore: Upgrade to newest Node LTS version

munen:
Closes: https://github.com/200ok-ch/organice/issues/992

As for the specific NodeJS version used, I'm completely open. I went with v20.17.0 in this first spike, because that's the latest LTS version as far as I can see. If, for any reason, that's not a good choice (for example on Nix), let's use a different new version.
 
TODOs until this PR can be merged:

## CLI tooling needs to work

These tasks should run:

- [x] `rm -rf node_modules; yarn install`
- [ ] `yarn test`
- [x] `yarn run start` 
  - Note: Atm, it works only with the  `--openssl-legacy-provider` workaround. There's inline comments for more info.
- [ ] `yarn run build`
- [ ] `yarn prettier-eslint --write`
- [ ] `yarn eslint`

## CI needs to work

- [ ] Upgrade Docker image for CI
- [ ] CI should be green

## Manual integration test

- [ ] /sample still works
- [ ] Logging into dropbox still works
- [ ] Logging into gitlab still works
- [ ] Logging into webdav still works


Makefile
```diff
@@ -11,7 +11,7 @@ setup:
 
 .PHONY: run
 run: setup
-	yarn start
+	NODE_OPTIONS=--openssl-legacy-provider yarn start
```
munen, line 14: For the moment, this is a workaround. Without that flag, OpenSSL has errors. On the one hand, we don't really need harder OpenSSL security defaults (I think), because we have no backend. However, introducing workarounds for legacy options is probably also not the best idea.

munen: I think this has been introduced since node v17. So, if we cannot fix it and it is a hard problem, we could try upgrading to v16 first. However, v16 is not maintained anymore, as well. So that's not the greatest of all options..

schoettl:
At this point, for `yarn build`, I should probably upgrade eslint to allow date-fns use of newer language features (classProperties). But react-scripts seem to depend on the old eslint...

Next is to upgrade eslint* and react* and fix all errors at once...

munen:
"In between" kudos for the progress on this PR :100:

schoettl:
I now get this success message for `yarn start`
image

But the browser page stays blank with
```
Uncaught ReferenceError: process is not defined
```

It's because of the upgrade to react-scripts v5 [1]. For some NodeJS globals like `crypto`, `process`, etc., react-scripts used to generate polyfills. In previous versions, `process.env.MY_ENV_VAR` was apparently poly-filled and inlined at compile-time. We use that in some places, e.g. src/components/SyncServiceSignIn/index.js.

For most globals I've found replacements (commit a0e7a68) but not yet for process.

[1] https://medium.com/@runawaytrike/upgrading-react-scripts-to-v5-for-idiots-16ba48d01c20

schoettl:
Oh, it seems to be related with `dotenv` which I removed because I didn't understand how it's used....

schoettl:
No, it's still the upgrade to react-scripts v5.
https://github.com/facebook/create-react-app/issues/12354

When I downgrade back to v4 I'd also need to downgrade eslint (from 9 to 7) :-(

munen:
Superseded by https://github.com/200ok-ch/organice/pull/1000

PR #996, opened 2024-10-31
fix: Prevent app crash when deleting a list item immediately after exiting edit mode

### LINKED ISSUE #988: press + then x, the sample file crashed
Screenshot_2024-09-18-08-14-00-98_df198e732186825c8df26e3c5a10d7cd

i was using edge on my phone.
trying sample page

 
Screenshot_2024-09-18-08-16-16-85_df198e732186825c8df26e3c5a10d7cd

kjmatsuda:
Fix #988 where the application would crash if a list was deleted immediately after exiting edit mode for the list title (or content).


munen:
@kjmatsuda Thank you very much :pray: Your fix works perfectly!

I'll merge, and add it to the changelog with attribution :bow:

PR #997, opened 2024-11-09
chore: Some tests while upgrading

munen:
@schoettl I tried to work off your PR and maybe get to the next step. There's a lot of upgrading at the same time, though^^

Some things I did:

- Follow the advise in: https://github.com/facebook/create-react-app/issues/12354#issuecomment-1603145990 by working off [this tutorial](https://dev.to/henriquejensen/migrating-from-create-react-app-to-vite-a-quick-and-easy-guide-5e72).
  - Result: I reverted back. After some work, I got organice to 'run' with Vite. However, it was only a white screen. So I reverted.
- I kind of followed your lead by going back to the latest react-scripts@4 and eslint@7. But then I still had to do some things to get `yarn start` to run:
  - Pin `babel/core`. Otherwise a transient dependency of react-scripts will complain.
  - Install `sass`. I see you removed `node-sass`, but we still need some sass compilation.
- Revert the upgrade of date-fns. I'm generally in favor of the upgrade, but there was more to change and I wasn't able to. There's some 'chaining' happening which would require new babel plugins. I tried, but failed. So I opted for the revert.

Now, `yarn start` kind of works. I see no error in the console and I can use organice. But there's still a lot of work left. For example there's an error in `yarn test` and `yarn build` won't work at all.

I'll leave my changes in this PR in case you want to use them.


munen:
Made obsolete by https://github.com/200ok-ch/organice/pull/1000 and https://github.com/200ok-ch/organice/pull/1014

PR #998, opened 2024-11-09
feat: Upgrade node to 16.20.2

munen:
Since https://github.com/200ok-ch/organice/pull/993 is proving to be a pretty big PR, I thought I'll try how far we can go without having to 'change all the things'. We can at least go up four major versions.

Node v16.20.2 has been released on 2023-08-08. It's not an LTS release, but it's also not super old. I'd like to see it as a pragmatic interim upgrade.

I'm not saying we shouldn't do more upgrades. I'm just trying to help out in the original task. My hope is that it'll get easier if we do it step by step.

@schoettl  What do you think?


schoettl (approved):
Seems like a good step forward. On my system, `yarn install` hangs at 0/2132 when starting with clean node environment (node version 20.15.1 though). But that's probably a NixOS problem and shouldn't keep you from merging.

munen:
Let's hope it's a nix, yarn version or node version thing. If it persists, I'm happy to do a pairing session on it.

PR #1013, opened 2025-04-04
fix(1006): Creating a static /manifest.json

### LINKED ISSUE #1006: When self-hosted, manifest is missing when installing PWA on iOS
**Describe the bug**

When on iOS, PWA doesn't work properly on self-hosted Organice - that is, built and served from my own nginx, or `bin/entrypoint.sh`. (Not tested on Android yet.)

This problem goes away if:

- I use https://organice.200ok.ch instead of my self-hosted installation.
- I manually copy https://organice.200ok.ch/manifest.json to my installation's web root.

In digging a little: `index.html` contains `/manifest.json` [here](https://github.com/200ok-ch/organice/blob/e3cf1c59cb467ae0168b78eb2c49bc8aed8f0136/public/index.html#L24) as the manifest the PWA needs - but this `manifest.json` isn't present anywhere in the repository or build.

However, the script in `index.html` then rewrites this manifest's href to that of a JSON blob created in the script. If I use a desktop browser, the manifest then seems to be visible, e.g. if I go to Developer Tools, Application, & Manifest, the proper JSON is visible.

Does iOS refuse to run the script when installing as a PWA, or something like that? I suspect someone was aware of this, which is why https://organice.200ok.ch/manifest.json exists.

**To Reproduce**

1. Set up Organice to be self-hosted, e.g. with nginx.
2. Follow the iOS instructions at https://organice.200ok.ch/documentation.html#installation - using the URL of the self-hosted Organice, not https://organice.200ok.ch.
3. Press the icon just created on the home screen.

Organice will just open as a tab in the default browser - not as a PWA. iOS appears to be treating it just as a bookmark.

**Expected behavior**
It's expected that Organice opens as a PWA - not just a Safari tab/bookmark.

**Screenshots**
N/A

N/A

 - Device: iPhone 11
 - OS: iOS 18.1.1
 - Browser: Safari

munen:
Closes #1006 

Originally, I think we created it dynamically, because we thought the `start_url` has to be dynamic for the self hosting. However, according to current [MDN](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/start_url#start_url) this is not the case (anymore). So I think, we can go the easy route and just use a static file for every host.

For reference, here's the original implementation: https://github.com/200ok-ch/organice/pull/781

I tested this on Staging and it also looks OK to me in the dev tools:

image

@tbruckmaier As the original author, is it OK for you if we change the implementation as proposed in this PR?

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -37,42 +37,5 @@
       </div>
     </noscript>
     <div id="root"></div>
-    <script>
-     var {pathname} = window.location;
-
-     const absoluteUrl = function(path) {
-        return new URL(path, document.baseURI).href;
-     };
-
-     var manifest = {
-         "short_name": "organice",
-         "name": "organice",
-         "icons": [
-             {
-                 "src": absoluteUrl("organice-512x512.png"),
-                 "sizes": "512x512",
-                 "type": "image/png"
-             },
-             {
-                 "src": absoluteUrl("organice-192x192.png"),
-                 "sizes": "192x192",
-                 "type": "image/png"
-             },
-             {
-                 "src": absoluteUrl("favicon.ico"),
-                 "sizes": "64x64 32x32 24x24 16x16",
-                 "type": "image/x-icon"
-             }
-         ],
-         "start_url": absoluteUrl(pathname || "/index.html"),
-         "display": "standalone",
-         "theme_color": "#000000",
-         "background_color": "#ffffff"
-     }
-
-     const manifestBlob = new Blob([JSON.stringify(manifest)], {type: 'application/json'});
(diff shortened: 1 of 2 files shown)


tbruckmaier:
Hi @munen , I have not used organice for quite some time, so I cannot really evaluate this CR. The original issue #779 complained about the dynamic manifest though, so a static one should probably be fine. Best of luck with your project!

munen:
@tbruckmaier Thank you for your fast response :pray: 

All the best in your productivity journey :muscle:

PR #1014, opened 2025-04-11
chore: Switch from CRA to parcel

munen:
We've used Create React App as the main build tool which abstracted stuff like Babel config away from us. In a recent release, we upgraded from CRA v3 to v4 so we were able to upgrade some other dependencies. However, CRA is EOL: https://react.dev/blog/2025/02/14/sunsetting-create-react-app

There are *many* options given in the EOL announcement on how to continue. In this PR, I'm going with one of the ways that seems both easiest and at the same time less tightly coupled than others. I'm going with [parcel](https://github.com/parcel-bundler/parcel) which has the claim "The zero configuration build tool for the web. 📦🚀"

Here's the upstream migration guide that I followed: https://parceljs.org/migration/cra/

On top of that, we need some further adjustments to keep all functionality.


munen:
@schoettl I did spend half a day moving off the EOL Create React App which was an integral part of building organice. Next to the issue of CRA being EOL, when we upgraded from CRA v3 to v4, this created the bug that *no* updates would reach the user. On my iPhone, I still have the version of November. It didn't help to aggressively try with restarts and deleting the cache, even.

If you want, you can walk through this PR. There's some docs in the WIKI and changelog. Also the commits are pretty much semantic. Walking through is optional. If you don't have time, we can merge as is. 

I did some testing locally and on Staging. The one bit of bad news is that I had to delete my bookmark on the iPhone and 're-install' to get the new version. CRA v3 had such incompatible caching which I could not get rid of, unfortunately. To have a second test, I have also already deployed this PR to prod.

The good news is that the new caching seems to be a lot smarter, updates work fast and I can even see icons in offline mode, now(; Also other good news (see wiki and changelog for more info).

schoettl:
Wow, this is a great step towards the future! I could yarn test start and build on my system. Looks good to me.

Thank you for your efforts @munen!

munen:
@schoettl Wow, that was a fast response and action taken, kudos :rocket: 

And thank you for the kind words :pray:

PR #1018, opened 2025-05-06
fix: Resolved issue where search input field could scroll off-screen

munen:
- The search input in popups (like Search, Task List) will no longer attempt to automatically focus on iOS. This prevents the field from being pushed out of view by the virtual keyboard. iOS users will now need to tap the input to activate it.
- On non-iOS platforms, the search input will now auto-focus with a slight delay to improve usability.


munen:
Note: Cannot manually regression test right now, because Dropbox has login issues (infinite spinner after entering the password).

munen:
Tested :+1:

PR #1025, opened 2025-05-24
feat: Add long-press to duplicate header and enhance visual feedback

### LINKED ISSUE #894: copy/duplicate header
I'm trying again to use organice in our company for production planning using orgmode exporter ox-taskjuggler.el to generate Gantt and resource diagrams. New orders are automatically added to the org file and we can use organice to manage the file.

grafik

For this, I need a method to "split" an order if it is produced in two batches.

I'm thinking of a "duplicate" or "copy" button in the header toolbar. This way, the task (=header) is duplicated with same order number and customer information and I can sort the copies as sub tasks into different production batches.

Do you think this duplicate function would be general useful? Or can it be combined with the "+" button somehow? Long-press (hidden functionality)?

I think adding a dedicated "copy" icon to the toolbar may be best.

munen:
Closes #894 

Implements a long-press gesture on the plus icon within the
HeaderActionDrawer to duplicate the current header and its entire
subtree. This provides a quick alternative to manually recreating
similar header structures.

There's visual feedback for long-press interactions, too.

@schoettl It's taken a *loong* time, but I finally got around to your feature request. What do you say, can we use it like this?

NB: I would have liked to go with vibration feedback on mobile, but it never worked on my phone, so I went with a bit of clumsy visual feedback(;

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -8,6 +8,19 @@ All user visible changes to organice will be documented in this file.
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
 * [2025-05-24 Sat]
+
+** Added
+*** Duplicate header via long-press on '+' icon
+
+- Long-pressing the '+' icon in the header action drawer now
+  duplicates the current header and its entire subtree.
+- Visual feedback for the long-press interaction has been
+  significantly enhanced with larger surrounding shadows for better
+  visibility, especially on touch devices.
+- Short-press on the '+' icon continues to add a new, empty header.
+
+PR: https://github.com/200ok-ch/organice/pull/1025
+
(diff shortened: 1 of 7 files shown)


schoettl:
Nice, I had this feature request in mind for a long time. Thanks for picking this up!

I ran into a bug when duplicating a headline which has a single sub headline.

grafik

grafik

Can you reproduce?

munen:
I'll merge it and if there's something to be changed, let's do in another PR :+1:

schoettl:
I was just testing in this moment :) Looks good to me!

munen:
Sweet, thank you very much and kind greetings :wave: :hugs:

Issue #1011, opened 2025-03-27
TODO -> DONE on repeating tasks is sometimes logged into the wrong heading


schoettl:
**Describe the bug**
For headings with a repeating SCHEDULED timestamp like `<2025-03-27 .+1w>`, swiping it to DONE is logged in the wrong `:LOGBOOK:` drawer when the swiped heading is not selected/focused.

**To Reproduce**

1. Create two headings
2. Set a repeating timestamp for the first
3. Select the second heading
4. Swipe the first heading to DONE
5. Check out the log notes `- State "DONE" from "TODO" [<timestamp>]`

**Expected behavior**
The TODO->DONE should be logged to the swiped heading, not to the selected heading.

**Additional context**

- Is it also the case for the setting "Tap TODO to advance state"?
- Is it also the case for the setting "Log into LOGBOOK drawer when item repeats"?

Loosely related to #984.

PR #1029, opened 2025-06-01
fix: repeater state changes will now be filed to the correct header

vHugoObject:
Fixes: #1011 and #750 

I found this error when I added a headline to the end of `various_todos` fixtures, right after the "Repeating task" header. I was adding new Todos as I am going to rewrite #984. Adding a new header caused the 'should advance repeating task' test to fail. It failed at the line where it asserted the `logNotes` of the entry with a repeater had changed. I was able to trace this error to addNoteGeneric where I noticed `headerId` was undefined as `state.get('selectedHeaderId')` returned undefined. However, this would never cause `addNoteGeneric` to fail since `headerId` is passed to `indexOfHeaderWithId`, which relies on `findIndex`. `findIndex` returns -1 whenever it can find an element that satisfies your testing function. That is the reason `updatePlanningItemsWithRepeaters` appeared to work correctly. That is also why the 'should advance repeating task' test would fail whenever a header was added to `various_todos`.  To fix this, I added two more headers to `various_todos`. Then, I revised 'should advance repeating task' , 'should advance repeating task again' and 'should advance active timestamp with repeater in header'. Finally, I made sure to select the header that was being edited in `updatePlanningItemsWithRepeaters`.

I should add that although I resolved this specific issue, I really think `indexOfHeaderWithId` should be looked at and rewritten. Any function that relies on `indexOfHeaderWithId` could end up modifying the wrong header since `findIndex` returns -1, which is still technically an index, when it fails. There is a good chance `indexOfHeaderWithId` is causing other problems.

### DIFF
--- a/changelog.org
+++ b/changelog.org
@@ -7,6 +7,16 @@ All user visible changes to organice will be documented in this file.
 
 When there are updates to the changelog, you will be notified and see a 'gift' icon appear on the top right corner.
 
+* <2025-06-06 Fri>
+
+** Fixed
+
+*** Repeater state changes will now always be filed to the correct header
+
+- =updatePlanningItemsWithRepeaters= now selects the header it intends to update
+
+PR: [[https://github.com/200ok-ch/organice/pull/1029]]  
+  
 * [2025-05-24 Sat]
 
 ** Added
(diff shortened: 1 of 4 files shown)


vHugoObject:
> Thanks @vHugoObject, looks like a very clean bugfix even with tests! I'll test it the next day before we merge it.

Thank you. I will restart the work on logdone once this is merged.

vHugoObject:
@schoettl Added it to the changelog.

Issue #586, opened 2020-11-25
improvements when deleting header items


markamaze:
Before creating a new issue, please read the "How we work with issues" section in the documentation: https://organice.200ok.ch/documentation.html#how_we_work_with_issues

**Is your feature request related to a problem? Please describe.**
on a laptop, it is too difficult to swipe to delete with a mouse
on a phone, it is too easy to delete and may lead to unintentionally removing header items

**Describe the solution you'd like**
add a headerActionItem to initialize the same action as a swipe
add a confirmation handler after the remove action to allow user to cancel or confirm the delete

**Describe alternatives you've considered**

**Additional context**

I'd like to put some work into this. I'm a newer developer looking for open source projects and this seems doable for me.  before finding organice, I was looking at it on org-web and believe that to add the confirmation I need to look into react-motion and how it's implemented in the header in order to add a confirmation following the delete action.  If this issue is worth working on, any input on a good approach would be appreciated.

tarnung:
Hi @markamaze 
You are not the first having issues with this.
We recently disabled user select to make swiping on the desktop a bit more usable. At least now you don't accidently select text.

I recently played around with alternative UI approaches. There I used swipe for visibility cycling (exactly because it's an action that doesn't change anything about your file) and an action button for deleting a header. In pr #559 you can find some images further down the thread. Feel free to join the discussion about where to take it!

I don't really like the Idea of having a button and a swipe action doing the same thing. We could use swipe on mobile and buttons on desktop, but that also complicates things since we would have to keep both UIs in mind for all further changes.

A confirmation would be too cumbersome in my opinion. Undo lets you recover an accidentally deleted header easily (if you realise you deleted it). Granted it could be done minimally intrusive by highlighting the header itself after swipe, tapping it again to confirm deletion.

I think changing delete on swipe needs to be part of a bigger UI rethink. If we decide to keep the swipe to delete action, I'd support having a confirm. But I'm probably not the only one with opinions on this.

munen:
Hi @markamaze 

Thanks for checking out organice and offering to contribute!🙏🏻

My thoughts on the issue:

Delete on swipe is a very common UX pattern on iOS (Apple and non Apple apps use it). For me, it’s also working equally well in organice as in the native apps. I have never seen an app require a confirmation. Also, we have infinite undo/redo, hence it’s even better in organice compared to an Apple app (where you’d have to do an undiscoverable “shake the phone” to maybe get an undo with no redo). 

[...]

PR #1032, opened 2025-06-14
feat: Added confirm button on a swipe to remove a header

ApollonDeParnasse:
Related: #586 
Accidental deletes have been a major annoyance. They seem to happen constantly whenever I use Organice on my phone. This very small change will simply create a popup that will ask a user to confirm they want to delete a header. This prompt will only appear on left swipes.

### DIFF
--- a/src/components/OrgFile/components/Header/index.js
+++ b/src/components/OrgFile/components/Header/index.js
@@ -150,10 +150,11 @@ class Header extends PureComponent {
       }
 
       if (-1 * swipeDistance >= this.SWIPE_ACTION_ACTIVATION_DISTANCE) {
-        this.setState({
-          isPlayingRemoveAnimation: true,
-          heightBeforeRemove: this.containerDiv.offsetHeight,
-        });
+        window.confirm('Please confirm your request to delete this header') &&
+          this.setState({
+            isPlayingRemoveAnimation: true,
+            heightBeforeRemove: this.containerDiv.offsetHeight,
+          });
       }
     }
 


munen:
Hi @vHugoObject

Thank you for your contribution! I understand that accidental swipes were a thing in the past and they were very annoying! But are they still since [this semi-recent change](https://github.com/200ok-ch/organice/pulls)?  

For me, I don't have accidental swipes since that change. If you still have them, please check your changelog to ensure you're running the latest version. Currently it looks like this:

image

If there's still an issue for other users, I propose to fix the underlying issue, instead. There's loads of prior art of mobile apps having swipe left/right to do stuff like delete on both iOS and Android. I don't know of any app that has a pop-up as in your proposal. Wherever possible, we'd like to stick to UX that people are used to. So, let's discuss in what way organice behaves differently then those apps and make it behave the same way^^

vHugoObject:
I am running the latest version. For me, accidental deletes happen every single day. An accidental delete yesterday afternoon is why I sat down and figured this out in the first place. On the other hand, I rarely delete headers in Organice. 


As for prior art, when I swipe left to delete a text message thread on my iphone, a confirmation popup does shows up. That is actually what gave me this idea. For the notes app, swiping left is not enough to delete a header. You also have to press delete. Of course gmail does allow you to swipe delete with no prompt. But gmail has trash where you can easily retrieve emails. Google docs and google drive don't allow you swipe delete at at all. Dropbox also has no swipe delete option. 

You could say I am cheery picking. But it is also clear that there isn't a consensus on swipe delete.

schoettl:
A setting with 3 options may be a good solution:

left-swipe to delete: enable, enable with confirmation, disable

Although with "disable" set there would currently be no other option to delete an item. Maybe not a good solution.

vHugoObject:
Currently, you have an option to delete a header when you press the pencil. So the left swipe is not your only option.

PR #1043, opened 2025-07-19
refactor: Converted TableEditorModal into a function 

vHugoObject:
I have added Table, TableCell and TableCellEditorContainer components to Organice. These components form the foundation of the new TableEditorModal. This is a significant code revision. But none of the actual functionality of the TableEditorModal has changed. The goal was simply to get the TableEditorModal to a place where it would be easier to implement new features.

My original goal was to add to touch scrolling to the TableEditorModal as the lack of touch scrolling makes it impossible to edit large tables on mobile. It quickly became clear that it would be impossible to add that feature as long as TableEditorModal was a class component.

Table, TableCell and TableCellEditorContainer were placed in their own folder that is separate from TableEditorModal. This way, Table-Part will also be able to use those components if this pull request is approved.


munen:
This is amazing🙌

I’ll try and review this as soon possible. It’s a big help, I’m also refactoring some class components into functions for maintainability, so this is really great timing. Thank you for your effort🙏🏻

PS: I’ll be in the mountains until afternoon, maybe I can review/merge it then, already.

vHugoObject:
I believe the eslint error is because the files in the dist folder are being formatted and linted even though they should be ignored. The correct prettier-eslint command should be something like this: "./node_modules/.bin/prettier-eslint "src/**/*.js" --write".

PR #1053, opened 2025-10-21
Fix(Components/OrgFile/TimestampEditor): Incorrect Timestamp Format

ApollonDeParnasse:
The org timestamp format is YYYY-MM-DD HH:mm. The line I removed stripped the leading zero from the hour part of the date. So [2018-09-17 Sun 06:00-11:30] was incorrectly saved as [2018-09-17 Sun 6:00-11:30]. All tests still pass. This changes doesn't effect the timestamp backend at all since it already expects the time part of timestamps to be in HH:mm format.


schoettl:
Thanks for your fix @vHugoObject! Actually, in my tests when setting a scheduled or deadline timestamp, the 06:00 in [2018-09-17 Sun 06:00-06:30] was already correct but the 06:30 (end time) was wrong.

With your fix the end time both start and end time are correct.

PR #1055, opened 2025-11-02
Add "#" and "%" as valid characters for tag searches.

cperl82:
The official org documentation doesn't currently mention them, but they are explicitly listed as valid in org-set-tags-command.

Based on version control, it seems like these have been supported for a very long time. I'm planning on submitting a request to the org-mode mailing list to update the documentation to match, but I figured I would do that in parallel with this pull request.

If you'd prefer to wait until the documentation is updated, that is fine too.

I noticed this because I use tags beginning with "#" for context, e.g. "#technology".


cperl82:
Thanks! Fwiw my patch to update the org docs was just added to the org-mode repo (commit 326ca3d13), so should be reflected on the website in the not-too-distant future.

PR #1061, opened 2025-11-11
Fix: move preventDefault to prevent race with oauth request

### LINKED ISSUE #1059: GitLab sign-in for private repository fails with "Unexpected sign in error, please try again"
**Describe the bug**

Signing in to GitLab with a private repository URL at https://organice.200ok.ch/sign_in fails with the message "Unexpected sign in error, please try again".

This persists after clearing site data.

I'm able to reproduce this in Chrome on Linux (NixOS) and on Android. I can't reproduce this with Firefox, so it may be Chrome-specific.

**To Reproduce**
1. In Chrome, go to https://organice.200ok.ch/sign_in
2. Attempt to sign in to GitLab, giving the URL of a private repository
3. Sign-in fails with a popup showing "Unexpected sign in error, please try again" (see screenshot)

**Expected behavior**
Clicking "Sign-in" should not produce an error and should redirect to a GitLab login page.

**Screenshots**

Image

 - OS: Linux (NixOS)
 - Browser: Chrome

 - Device: Pixel 8 Pro
 - OS: Android 16
 - Browser Stock browser (Chrome)

mcwitt:
Fixes #1059 

Addresses a race condition where the default form submission request may preempt the request to the GitLab OAuth endpoint, causing the latter to be cancelled. When this happens, the local storage `oauth2authcodepkce-state` ends up in a corrupt state, resulting in the error observed in #1059.

This moves the call to `preventDefault()` to the beginning of the form submission handler, so it is called also when the project URL is valid. This placement is also consistent with the WebDAV handler.

It's unclear to me what changed in recent versions of Chrome to make the issue manifest, but I think the change is correct regardless.

### DIFF
--- a/src/components/SyncServiceSignIn/index.js
+++ b/src/components/SyncServiceSignIn/index.js
@@ -113,13 +113,13 @@ function GitLab() {
   const defaultProject = 'https://gitlab.com/your/project';
   const [project, setProject] = useState(defaultProject);
   const handleSubmit = (evt) => {
+    evt.preventDefault();
     const projectId = gitLabProjectIdFromURL(project);
     if (projectId) {
       persistField('authenticatedSyncService', 'GitLab');
       persistField('gitLabProject', projectId);
       createGitlabOAuth().fetchAuthorizationCode();
     } else {
-      evt.preventDefault();
       alert('Project does not appear to be a valid gitlab.com URL');
     }
   };


munen:
@mcwitt @dtvillafana Thank you for your hard relentless effort in debugging and fixing this issue. I have confirmed that it fixes the Gitlab login for Chrome versions > 137, and continues to work for the other browsers that did not show this issue so far.

I have added your contribution to the changelog and will merge your changes!

Thank you, again 🙏 🙇‍♂️

PR #1088, opened 2026-01-09
CI: Send junit test reports from CircleCI via API back to GH

munen:
Builds on https://github.com/200ok-ch/organice/pull/1090


munen:
Sending junit reports back to GH is not a feature that GH natively supports. It's a lot of custom code and brittle. Hence, we're not doing it. In GH you can see the steps and checks. In CircleCI you can see the tests that failed.

PR #1118, opened 2026-05-17
fix: capture into target files

munen:
Fixes capture templates that save into another file by loading and syncing the target file.


munen:
Superceded by https://github.com/200ok-ch/organice/pull/1119

PR #1060, opened 2025-11-11
Add open-source git-based sync backend https://gitea.com/

mergezalot:
- Also added branch selection, on top of what the gitlab sync backend does.

PR #1130, opened 2026-09-20
Add Forgejo as a sync backend (access token version)

tarnung:
See #1129


tarnung:
I just found #1060 which probably does much of the same as this pull request as Forgejo is a Fork of Gitea. I don't know if they are fully compatible though.

Anyways, having two pull requests reinforces that there is interest in this feature.

